PR #113, opened 2015-08-05
Added ability to add json flag to command to get a response in json f…

Zandrr:


client/go/commands/cluster_create.go
```diff
@@ -90,9 +91,14 @@ func (a *CreateNewClusterCommand) Exec(args []string) error {
...
 		return err
 	}
-
 	//if all is well, print stuff
-	fmt.Fprintf(stdout, "Cluster id: %v", body.Id)
+	if a.options.Json == false {
+		fmt.Fprintf(stdout, "Cluster id: %v", body.Id)
+	}
+	if a.options.Json == true {
+		res, _ := json.Marshal(body)
```
lpabon: Since the body already contains the JSON string, all we need to do is extract it and print it without parsing it.  Here is an example of cluster list:

```
	// Check if we need to print the JSON body
	if a.options.Json {
		// Print JSON body
		s, err := utils.GetStringFromResponse(r)
		if err != nil {
			return err
		}
		fmt.Fprint(stdout, s)
	} else {

		//check json response
		var body glusterfs.ClusterListResponse
		err = utils.GetJsonFromResponse(r, &body)
		if err != nil {
			fmt.Println("Error: Bad json response from server")
			return err
		}

		// Print to user cluster lists
		str := "Clusters: \n"
		for _, cluster := range body.Clusters {
			str += cluster + "\n"
		}
		fmt.Fprintf(stdout, str)
	}
```

This should be done for all commands that get JSON responses

client/go/commands/cluster_destroy.go
```diff
@@ -96,8 +96,13 @@ func (a *DestroyClusterCommand) Exec(args []string) error {
 	}
 
 	//if all is well, print stuff
-	fmt.Fprintf(stdout, "Successfully destroyed cluster with id: %v ", clusterId)
+	if a.options.Json == false {
```
lpabon: Since this command does not return JSON, do not print anything into  the stdout, because it would mess up parsers.  Instead just return success or failure.

Therefore you should change the `if` to:

```
if !a.options.Json {
   ... Print strings ...
}
```

client/go/commands/cluster_info.go
```diff
@@ -96,17 +96,25 @@ func (a *GetClusterInfoCommand) Exec(args []string) error {
...
+			s += volume + "\n"
+		}
+
+		fmt.Fprintf(stdout, s)
 	}
 
-	fmt.Fprintf(stdout, str)
+	//if json is yes, pass back some json
+	if a.options.Json == true {
+		res, _ := json.Marshal(body)
```
lpabon: See earlier comment for cluster list

client/go/commands/flags_test.go
```diff
@@ -0,0 +1,88 @@
...
+	var b bytes.Buffer
+	defer tests.Patch(&stdout, &b).Restore()
+
+	//create mock cluster and mock destroy
+	mockCluster := NewCreateNewClusterCommand(options)
+
+	//create new cluster and assert json
+	err := mockCluster.Exec([]string{})
+	tests.Assert(t, err == nil)
+	tests.Assert(t, strings.Contains(b.String(), "{\"id\":\""), b.String())
```
lpabon: Just a suggestion.. instead of using `\"`, you can just change the entire string to just the tick mark and double quotes will not be interpreted.  See app_X_test.go in glusterfs for examples.

client/go/commands/flags_test.go
```diff
@@ -0,0 +1,88 @@
...
+
+	//get cluster id and reset b.String()
+	MockClusterIdArray := strings.SplitAfter(b.String(), "")
+	tests.Assert(t, len(MockClusterIdArray) > 0)
+	MockClusterId := MockClusterIdArray[7:39]
```
lpabon: This part is weird.  What is being grabbed here?  Should it instead be marshaled?

client/go/commands/flags_test.go
```diff
@@ -0,0 +1,88 @@
...
+	b.Reset()
+
+	//get cluster info and assert json
+	clusterInfo := NewGetClusterInfoCommand(options)
+	args := []string{strings.Join(MockClusterId, "")}
```
lpabon: Why do we need to join clusterid with `""` ?

client/go/commands/flags_test.go
```diff
@@ -0,0 +1,88 @@
...
+	err = clusterInfo.Exec(args)
+	tests.Assert(t, err == nil)
+	tests.Assert(t, strings.Contains(b.String(), "{\"id\":\""), b.String())
+	b.Reset()
+
+	//get cluster list and assert json
+	clusterList := NewGetClusterListCommand(options)
+	err = clusterList.Exec([]string{})
+	tests.Assert(t, err == nil)
+	tests.Assert(t, strings.Contains(b.String(), "{\"clusters\":[\""), b.String())
```
lpabon: If you want, you can unmarshal the JSON into a struct from apps/glusterfs/models.go and check if the information is correct.  This is probably what you want to do to verify the calls where successful.

client/go/main.go
```diff
@@ -32,6 +32,7 @@ var (
 func init() {
 
 	flag.StringVar(&options.Url, "server", "", "server url goes here.")
+	flag.BoolVar(&options.Json, "json", false, "get response as json")
```
lpabon: Nice

client/go/main.go
```diff
@@ -52,6 +53,10 @@ func main() {
 		fmt.Fprintf(stdout, "You need a server!\n")
 		os.Exit(1)
 	}
+	if options.Json != false && options.Json != true {
```
lpabon: AFAIK, there is no need to check.  The Flags() API will take care of that.

client/go/commands/flags_test.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/gorilla/mux"
+	"github.com/heketi/heketi/apps/glusterfs"
+	"github.com/heketi/heketi/tests"
+	"net/http/httptest"
+	"os"
+	"strings"
+	"testing"
+)
+
+func TestJsonFlagsCreate(t *testing.T) {
```
lpabon: Should add a test that shows that not having -json should print NO JSON output.  In other words.. unmarshalling the output would return an error.

client/go/commands/flags_test.go
Zandrr: I'm not sure if I understand this one, can you give me a specific file/line number to look at?

client/go/commands/flags_test.go
Zandrr: Yes I meant to ask about this ugly hack.  It's grabbing the id from a string.  The issue is I cannot marshal a []string.

PR #115, opened 2015-08-06
Added CLI functionality for node commands

Zandrr:


client/go/commands/node_add.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+	//do Post and check if sent to server
+	r, err := http.Post(url+"/nodes", "application/json", bytes.NewBuffer(request))
+	if err != nil {
+		fmt.Fprintf(stdout, "Error: Unable to send command to server: %v", err)
+		return err
+	}
+
+	//check status code
+	if r.StatusCode != http.StatusAccepted {
```
lpabon: Need to wait until the work has been completed.  Look at [app_volume_test.go](https://github.com/heketi/heketi/blob/master/apps/glusterfs/app_volume_test.go#L715) to line 733.  Also checkout the [node api](https://github.com/heketi/heketi/wiki/API#node_add) asynchronous operation.  The thing is that the request is accepted, but then we need to keep checking the result of the request on a different endpoint.  Once the API is completed, it will return some status, depending on the API.  For Node Add, it will return JSON.

client/go/commands/node_destroy.go
```diff
@@ -0,0 +1,117 @@
...
+
+	//destroy node
+	r, err := http.DefaultClient.Do(req)
+	if err != nil {
+		fmt.Fprintf(stdout, "Error: Unable to send command to server: %v", err)
+		return err
+	}
+
+	//check status code
+	if r.StatusCode != http.StatusOK {
```
lpabon: Node destroy is also asynchronous

client/go/commands/node_info.go
```diff
@@ -0,0 +1,132 @@
...
+
+		//check json response
+		var body glusterfs.NodeInfoResponse
+		err = utils.GetJsonFromResponse(r, &body)
+		if err != nil {
+			fmt.Println("Error: Bad json response from server")
+			return err
+		}
+		//print revelent results
+		s := "Node: " + a.nodeId + "\n\nZone: " + strconv.Itoa(body.Zone) + "\n\nCluster: " + body.ClusterId + "\n\nManage hostnames:\n"
```
lpabon: use multiple lines, something like this:

```
s := "Node: " + a.nodeId + "\n\n" +
       "Zone: " + ...
```

or

```
s := "Node: " + a.nodeId + "\n\n"
s += "Zone: " ...
```

client/go/commands/usage_templates.go
```diff
@@ -92,3 +92,59 @@ Example:
...
+Use "heketi node [subcommand] -help" for more information about a subcommand
+
+`
+var usageTemplateNodeAdd = `node add is a command used for adding a node to the specified cluster.
+
+Usage:
+    heketi -server [server] [options] node add [flags]
+
+The flags are:
+    -zone                  The zone in which the node should reside.
```
lpabon: I'm not sure if we need to do the flags printout... I think we need to look into the flags API before we continue this model.  I have used the flags API and it prints out the string given when defining the flag.  Maybe we are not setting it up correctly, like in `cmd.flags.Usage = ....`

lpabon: Ok, now I see how it works.  There is an API called PrintDefaults() which prints the flags out to stderr if flag.SetOutput() hasn't been specified.  I have created a [gist](https://gist.github.com/lpabon/ab51a9a95a2b506a2b6d) which prints the following out:

```
$ go run main.go node add -h

Node add is a command used for adding a node to the specified cluster.

USAGE
    heketi node add [options]

OPTIONS
  -cluster="": The cluster in which the node should reside
  -managment-host-name="": Managment host name
  -storage-host-name="": Storage network host name
  -zone=0: The zone in which the node should reside

EXAMPLES
	Add a node to heketi
	    $ heketi -server http://localhost:8080 node add \
	    	-zone 3 \
	    	-cluster 3e098cb4407d7109806bb196d9e8f095 \
	    	-managment-host-name node1-manage.gluster.lab.com \
	    	-storage-host-name node1-storage.gluster.lab.com
exit status 2
```

client/go/commands/node_add.go
lpabon: Also I fixed node delete to be asynchronous in #116 .  Please update your code to merge the change.

lpabon:
@Zandrr  
I have merged #134 .  You will need to update your unit tests to use the new glusterfs.NewApp() API, which is now glusterfs.NewTestApp(...).  See [cluster_test.go](https://github.com/heketi/heketi/blob/master/client/go/commands/cluster_test.go#L49) for an example.  I had to update the API because the heketi server now supports command line arguments.

To run the new server, form the top of the tree, type:

```
$ go run main.go -config=etc/heketi.json
```

lpabon:
Great, I'll review tonight.

client/go/commands/node_add.go
```diff
@@ -0,0 +1,163 @@
...
+		utils.GetStringFromResponseCheck(r)
```
lpabon: I think this part is missing error handling.  Also if it fails we the program continues instead of returning immediately.

client/go/commands/node.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	// Check which of the subcommands we need to call the .Parse function
+	for _, cmd := range a.cmds {
+		if a.flags.Arg(0) == cmd.Name() {
+			err := cmd.Exec(a.flags.Args()[1:])
+			if err != nil {
+				return err
+			}
+			a.cmd = cmd
```
lpabon: Do we need this line anymore?

client/go/commands/node_add.go
```diff
@@ -0,0 +1,163 @@
...
+
+func (a *NodeAddCommand) Exec(args []string) error {
+
+	//parse args
+	a.flags.Parse(args)
+
+	//ensure we have Url
+	if a.options.Url == "" {
+		fmt.Fprintf(stdout, "You need a server!\n")
+		os.Exit(1)
```
lpabon: We probably should not have any of the command functions call os.Exit().  Instead it should return an error

client/go/commands/node_add.go
```diff
@@ -0,0 +1,163 @@
...
+		r, err := http.Get(location.String())
+		if err != nil {
+			return err
+		}
+		if r.Header.Get("X-Pending") == "true" {
+			if r.StatusCode == http.StatusOK {
+				time.Sleep(time.Millisecond * 3)
+				continue
+			} else {
+				utils.GetStringFromResponseCheck(r)
```
lpabon: This function only returns the string inside the Body.  We should then create an error from the string and return it.  This may need to be a utility function since the same sequence is used in many places.

client/go/commands/node_add.go
```diff
@@ -0,0 +1,163 @@
...
+				if a.options.Json {
+					s, err := utils.GetStringFromResponse(r)
+					if err != nil {
+						return err
+					}
+					fmt.Fprint(stdout, s)
+				} else {
+					var body glusterfs.NodeInfoResponse
+					err = utils.GetJsonFromResponse(r, &body)
+					fmt.Fprintf(stdout, "Successfully created node with id: %v", body.Id)
```
lpabon: Here we need a nicely formatted output

client/go/commands/node_destroy.go
```diff
@@ -0,0 +1,136 @@
...
+	//destroy node
+	r, err := http.DefaultClient.Do(req)
+	if err != nil {
+		fmt.Fprintf(stdout, "Error: Unable to send command to server: %v", err)
+		return err
+	}
+
+	//check status code
+	if r.StatusCode != http.StatusAccepted {
+		utils.GetStringFromResponseCheck(r)
```
lpabon: Return on exit

client/go/commands/node_destroy.go
```diff
@@ -0,0 +1,136 @@
...
+		} else {
+			if r.StatusCode == http.StatusNoContent {
+				if !a.options.Json {
+					fmt.Fprintf(stdout, "Successfully destroyed node with id: %v ", nodeId)
+				} else {
+					return nil
+				}
+				break
+			} else {
+				utils.GetStringFromResponseCheck(r)
```
lpabon: Return error string

client/go/commands/node_info.go
```diff
@@ -0,0 +1,129 @@
...
+	//do http GET and check if sent to server
+	r, err := http.Get(url + "/nodes/" + a.nodeId)
+	if err != nil {
+		fmt.Fprintf(stdout, "Unable to send command to server: %v", err)
+		return err
+	}
+
+	//check status code
+	if r.StatusCode != http.StatusOK {
+		utils.GetStringFromResponseCheck(r)
```
lpabon: Return  error

lpabon:
Also, you may want to check your test coverage.  Check out https://blog.golang.org/cover .  Go into the directory with the code and run the go test command with the switch to capture the coverage.  Then use go tool to see the coverage in HTML format.

client/go/commands/node_add.go
Zandrr: Is it?  I moved the functionality into [here](https://github.com/heketi/heketi/pull/115/files#diff-8e7d97cbd93fca20b1547cd97d3e764cR35), assuming you're talking about line 126.

client/go/commands/node.go
Zandrr: which line?

Zandrr:
I'm at 78% coverage right now.  My thoughts are anything >90% is solid.  Thoughts?

client/go/commands/cluster_create.go
```diff
@@ -87,11 +85,7 @@ func (a *CreateNewClusterCommand) Exec(args []string) error {
 
 	//check status code
 	if r.StatusCode != http.StatusCreated {
-		s, err := utils.GetStringFromResponse(r)
-		if err != nil {
-			return err
-		}
-		return errors.New(s)
+		utils.GetStringFromResponseCheck(r)
```
lpabon: I'm not sure why this code changed.  The code that was removed is the correct method.  Please check all calls that pull the error string from the body.  They need to return it to the caller.

lpabon:
78% is really good, and I'm not really looking to get in the 90s, but I just wanted to see if the code like cluster_create.go line 88 was being tested.  I think that line can be forced in the unit test, and the expected result checked.

lpabon:
Awesome!

Issue #18, opened 2015-06-19
Heketi Systemd service


lpabon:
Setup heketi as a systemd service

sidcarter:
Never worked with Heketi or Fedora, but work with CentOS a lot. Happy to provide my cycles to help with this.

lpabon:
Excellent!  I'm currently working on #17 .  This would be greatly appreciated.  We need a simple systemd file which would start up the service `heketi`.  The default configuration file will be placed in `/etc/heketi/heketi.json`.  The program should be started using the following command:  `/usr/bin/heketi -config=/etc/heketi.json` . 

Note that Heketi RPM spec file is still in progress, so you will need to place the `heketi.json` configuration file in `/etc/heketi` and place the `heketi` binary in `/usr/bin` by hand to work on this.

You should base the systemd service on `etcd` which is another service written in Go.

sidcarter:
Got it going, I think. But getting errors, probably cause I don't have glusterfs installed. Will test a bit more. The errors:

```
Oct 12 21:29:04 ip-10-158-134-179 heketi: Heketi (dev)
Oct 12 21:29:04 ip-10-158-134-179 heketi: [heketi] INFO 2015/10/12 21:29:04 Loaded mock executor
Oct 12 21:29:04 ip-10-158-134-179 heketi: [heketi] ERROR 2015/10/12 21:29:04 /home/sidcarter/golang/src/github.com/heketi/heketi/apps/glusterfs/app.go:97: Unable to open database
Oct 12 21:29:04 ip-10-158-134-179 heketi: panic: runtime error: invalid memory address or nil pointer dereference
Oct 12 21:29:04 ip-10-158-134-179 heketi: [signal 0xb code=0x1 addr=0x0 pc=0x486b3b]
Oct 12 21:29:04 ip-10-158-134-179 heketi: goroutine 1 [running]:
Oct 12 21:29:04 ip-10-158-134-179 heketi: github.com/heketi/heketi/apps/glusterfs.(*App).SetRoutes(0x0, 0xc20800d090, 0x0, 0x0)
Oct 12 21:29:04 ip-10-158-134-179 heketi: /home/sidcarter/golang/src/github.com/heketi/heketi/apps/glusterfs/app.go:180 +0x12b
Oct 12 21:29:04 ip-10-158-134-179 heketi: main.main()
[...]

Issue #17, opened 2015-06-19
Create RPM for deployment


lpabon:
Currently Heketi uses a simple Makefile to create a tar.gz distribution since it is written in Go.  We may need to support an RPM so that it is installed with systemd support.

lpabon:
According to [Fedora Golang Packaging](https://fedoraproject.org/wiki/PackagingDrafts/Go) all the dependencies must have their own repo for the build process.

I will first determine which dependencies need repos and which do not for Release 1

lpabon:
The following repo dependencies must be create:
### Release 1
* golang-golang-x-crypto
* golang-github-lpabon-godbc
* golang-github-auth0-go-jwt-middleware
* golang-github-codegangsta-negroni
* golang-github-dgrijalva-jwt-go

### Release 2
Release 2 also needs the following:

* golang-github-heketi-tests
* golang-github-heketi-rest
* golang-github-heketi-utils

lpabon:
There is this really cool tool called [gofed](https://github.com/gofed/gofed) highlighted in the Fedora document.  I used it to start the spec file.  It also points out which dependencies are not part of Fedora:

```shell
$ gofed github2spec -p heketi -r heketi \
   --commit 7d2a7830f0c793ae8151c119ea41b3e8f0148c60 -f

Repo URL: github.com/heketi/heketi
Commit: 7d2a7830f0c793ae8151c119ea41b3e8f0148c60
Name: golang-github-heketi-heketi

(1/4) Checking if the package already exists in PkgDB
(2/4) Downloading tarball
(3/4) Generating spec file
(4/4) Discovering golang dependencies
Class: github.com/auth0/go-jwt-middleware (golang-github-auth0-go-jwt-middleware) PkgDB=False
Class: github.com/boltdb/bolt (golang-github-boltdb-bolt) PkgDB=True
Class: github.com/codegangsta/negroni (golang-github-codegangsta-negroni) PkgDB=False
Class: github.com/dgrijalva/jwt-go (golang-github-dgrijalva-jwt-go) PkgDB=False
Class: github.com/gorilla/context (golang-github-gorilla-context) PkgDB=True
Class: github.com/gorilla/mux (golang-github-gorilla-mux) PkgDB=True
Class: github.com/lpabon/godbc (golang-github-lpabon-godbc) PkgDB=False
Class: golang.org/x/crypto (golang-github-golang-crypto) PkgDB=False

[...]

PR #219, opened 2015-10-13
Heketi systemd

sidcarter:
Created a systemd service for heketi. Issue #18 

Once things are in place for #17, the service file may need to change location.


lpabon:
This looks great thanks!

PR #251, opened 2015-11-23
Python API for Heketi

anushshetty:
With this, an application should be able to run heketi operations in python. These are wrappers for Heketi API calls. All calls provide return values as tuples, the first part containing the actual output, and the second part containing the error status code (if any). 

TODO: The volume creation part doesn't support all the options yet.

### DIFF
--- a/client/api/python/heketi/heketi.py
+++ b/client/api/python/heketi/heketi.py
@@ -59,108 +59,92 @@ class HeketiClient(object):
         method = 'GET'
         uri = '/hello'
 
-        headers={}
-        self._set_token_in_header(method, uri)
+        headers = self._set_token_in_header(method, uri)
         r = requests.get(self.host + uri, headers=headers)
         return r.status_code == requests.codes.ok
 
 
-    def _make_request(self, method, uri, data={}):
-        self.method = method
-        self.uri = uri
-        self.data = data
-
-        headers = self._set_token_in_header(self.method, self.uri)
+    def _make_request(self, method, uri, data={}, headers={}):
+        headers = self._set_token_in_header(method, uri)
 
         ''' Ref: http://docs.python-requests.org/en/master/_modules/requests/api/#request '''
-        r = requests.request(self.method,
-                             self.host + self.uri,
+        r = requests.request(method,
+                             self.host + uri,
                              headers=headers,
                              data=json.dumps(data))
 
         r.raise_for_status()
 
-
-        if r.status_code == requests.codes.accepted and \
-           r.headers['location']:
+        if r.status_code == requests.codes.accepted:
             return self._get_queued_response(r.headers['location'])
-
-        elif r.status_code == requests.codes.SEE_OTHER and \
-             r.headers['location']:
-            return self._get_queued_response(r.headers['location'])
-
-        elif r.status_code == requests.codes.created or \
-             r.status_code == requests.codes.ok:
-            if self.method == 'DELETE':
-                return True
-            else:
-                return r.json()
-
-        elif r.status_code == requests.codes.NO_CONTENT:
-            return True
-
+        else:
+            return r
 
 
     def _get_queued_response(self, queue_uri):
-        self.queue_uri = queue_uri
-        headers = self._set_token_in_header('GET', self.queue_uri)
+        queue_uri = queue_uri
+        headers = self._set_token_in_header('GET', queue_uri)
         response_ready = False
 
         while response_ready is False:
             q = requests.get(self.host + queue_uri,
                              headers=headers,
-                             allow_redirects=False)
+                             allow_redirects=True)
 
             # Raise an exception when the request fails
             q.raise_for_status()
 
-
-            if q.status_code == requests.codes.see_other and q.headers['location']:
-                # Temporary request of 303 with header info
-                response_ready = True
-                vol_req = q.headers['location']
-                return self._make_request('GET', vol_req)
-
-            elif q.status_code == requests.codes.NO_CONTENT:
-                # Request successful
-                return True
-                response_ready = True
-
+            if 'X-Pending' in q.headers:
+                time.sleep(2)
+            else:
+                return q
 
 
     def cluster_create(self):
-        return self._make_request('POST', '/clusters')
+        req = self._make_request('POST', '/clusters')
+        if req.status_code == requests.codes.created:
+            return req.json()
 
 
     def cluster_info(self, cluster_id):
         uri = "/clusters/" + cluster_id
-        return self._make_request('GET', uri)
+        req = self._make_request('GET', uri)
+        if req.status_code == requests.codes.ok:
+            return req.json()
+
 
     def cluster_list(self):
         uri = "/clusters"
-        return self._make_request('GET', uri)
+        req = self._make_request('GET', uri)
+        if req.status_code == requests.codes.ok:
+            return req.json()
 
 
     def cluster_delete(self, cluster_id):
         uri = "/clusters/" + cluster_id
-        return self._make_request('DELETE', uri)
+        req = self._make_request('DELETE', uri)
+        return req.status_code == requests.codes.NO_CONTENT
 
 
     def node_add(self, node_options = {}):
         ''' node_options is a dict consisting of paramters for \
             adding a node: https://github.com/heketi/heketi/wiki/API#add-node '''
         uri = "/nodes"
-        return self._make_request('POST', uri, node_options)
+        req = self._make_request('POST', uri, node_options)
+        if req.status_code == requests.codes.ok:
+            return req.json()
 
 
     def node_info(self, node_id):
         uri = '/nodes/' + node_id
-        return self._make_request('GET', uri)
-
+        req = self._make_request('GET', uri)
+        if req.status_code == requests.codes.ok:
+            return req.json()
 
     def node_delete(self, node_id):
         uri = '/nodes/'+ node_id
-        return self._make_request('DELETE', uri)
+        req = self._make_request('DELETE', uri)
+        return req.status_code == requests.codes.NO_CONTENT
 
 
     def device_add(self, device_options = {}):
@@ -169,17 +153,22 @@ class HeketiClient(object):
             https://github.com/heketi/heketi/wiki/API#add-device
         '''
         uri = '/devices'
-        return self._make_request('POST', uri, device_options)
+        req = self._make_request('POST', uri, device_options)
+        if req.status_code == requests.codes.accepted:
+            return req.json()
 
 
     def device_info(self, device_id):
         uri = '/devices/' + device_id
-        return self._make_request('GET', uri)
+        req = self._make_request('GET', uri)
+        if req.status_code == requests.codes.ok:
+            return req.json()
 
 
     def device_delete(self, device_id):
         uri = '/devices/' + device_id
-        return self._make_request('DELETE', uri)
+        req = self._make_request('DELETE', uri)
+        return req.status_code == requests.codes.NO_CONTENT
 
 
     def volume_create(self, volume_options = {}):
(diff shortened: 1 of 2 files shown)


client/api/python/heketi/heketi.py
```diff
@@ -54,3 +56,535 @@ def hello(self):
         r = requests.get(self.host + uri, headers=headers)
         return r.status_code == requests.codes.ok
```
lpabon: One overall comment.  There seem to be objects for each endpoint type.  This would require the client to keep creating objects for each endpoint.  Instead, all those endpoints should be functions under the _Client_ class.  That way it would enable the client to create a single initialized object which has access to all api functions.  Also it would enable the member functions to access self._set_toket_in_header() and other utility functions (like pending for the request to complete).

client/api/python/heketi/heketi.py
```diff
@@ -54,3 +56,535 @@ def hello(self):
...
 
+
+class Cluster(object):
+    """ Class to run cluster operations """
+
+    def __init__(self, client):
+        self.client = client
+
+    def create(self):
```
lpabon: So when they become part of the same class.. change the names to <endpoint>_<api>, for example `cluster_create()`

client/api/python/heketi/heketi.py
```diff
@@ -54,3 +56,535 @@ def hello(self):
...
+        uri = "/clusters"
+
+        output = ''
+        err = ''
+
+        self.client._set_token_in_header(headers, method, uri)
+        r = requests.post(self.client.host + uri, headers=headers)
+
+        if r.status_code == requests.codes.created:
+            output = r.json()
```
lpabon: What happens if the json is unreadable, does it throw an exception?

client/api/python/heketi/heketi.py
```diff
@@ -54,3 +56,535 @@ def hello(self):
...
+        else:
+            err = r.status_code
+
+        return output, err
```
lpabon: It seems that the output of each of these API functions is a dict.  That's ok, as long as it is documented, which reminds me, this needs documentation :).  I do not see the need to return the status code to the caller.  They probably want to know if it worked or not, which means most likely raising an exception.  I remember you mentioned this in the gitter chat, but now that I see the code, it seems clearer that an exception would need to be raised.  Here is an example from [OpenStack Swift python client](https://github.com/openstack/python-swiftclient/blob/master/swiftclient/client.py#L787).  It does not have to be that detailed, but I just wanted to show an example of it raising an exception when an incorrect HTTP code has been received.

client/api/python/heketi/heketi.py
```diff
@@ -54,3 +56,535 @@ def hello(self):
...
+
+        self.client._set_token_in_header(headers, method, uri)
+        r = requests.post(self.client.host + uri,
+                          data = json.dumps(addnode_params),
+                          headers=headers)
+
+        if r.status_code == requests.codes.accepted:
+            queue_loc = r.headers['location']
+
+            if queue_loc:
```
lpabon: This section that pends should definitely be made a utility function of the _Client_ class so that other functions use it.

client/api/python/heketi/heketi.py
```diff
@@ -54,3 +56,535 @@ def hello(self):
...
+        return err
+
+
+
+class Node(object):
+
+    def __init__(self, client):
+        self.client = client
+
+    def add(self, **kwargs):
```
lpabon: Taking arguments as `**kwargs` is probably not the best method because the python compiler will not be able to check.  All these functions should have a class or something that they pass.  It could be as simple as `d` which is a dict, or it could be more, but `**kwargs` is too vague.

client/api/python/heketi/test_client.py
```diff
@@ -0,0 +1,287 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import unittest
+from heketi import *
+
```
lpabon: Sweet.  Does this test require heketi to be running?  If it does, it should probably document the setup needed.

client/api/python/heketi/test_client.py
```diff
@@ -0,0 +1,287 @@
+#
```
lpabon: I would like to see this as part of `tox`.  You can even use `tox` to setup the testing environment.  That can be the next phase of this code.  I'd like to see this in the tree first, then we can add the `tox` setup (and setup.py for pypy support).

client/api/python/heketi/test_client.py
anushshetty: Yes, it needs a Heketi server running. Will add a note there.

client/api/python/heketi/heketi.py
anushshetty: You are right. **kwargs is ugly too. Just using the dict should be okay too, as long as the application passes data in the form of a dict.

client/api/python/heketi/heketi.py
anushshetty: @lpabon, would a generic _make_request(method, uri, data) function work?

lpabon:
Any update?

lpabon:
@anushshetty Hi, do you think you have time to finish this by mid April?

anushshetty:
Yes, will speed up and get this done in two weeks.

nandajavarma:
Great effort @anushshetty ! Agree with @lpabon on rewriting each class as methods under a common class. That will help DRY up the code in lot many places. 
I am planning to make gdeploy use this API. Would you like to collaborate, so we can finish this up fast? Then I will have less work to do in gdeploy front ;)

anushshetty:
@nandajavarma, I have the code ready. Need to test it. Will sync up with you in a while.

nandajavarma:
@anushshetty Perfect!

anushshetty:
@lpabon and @nandajavarma , please do have a look. Do you think the exception handling can be done any better? 

@nandajavarma, all yours now :)

nandajavarma:
@anushshetty Thanks for working on this. Code wise, LGTM. I am pulling this file out and using for gdeploy. I will let you know if I hit any issues. Great work! :+1:

nandajavarma:
@anushshetty I think since you have called raise_for_status(), requests module itself will take care of most of the exceptions(i.e. Raise exceptions whenever bad status codes are received). I would like to use it and see if there are any unhandled tracebacks, etc.
/me is a heketi n00b and she thinks @lpabon should have a look. :-)

client/api/python/heketi/test_client.py
```diff
@@ -0,0 +1,278 @@
...
+        info  = c.cluster_info(cluster['id'])
+        self.assertEqual(True, info == cluster)
+
+        # Get a list of clusters
+        list = c.cluster_list()
+        self.assertEqual(True, len(list['clusters']) == 1)
+        self.assertEqual(True, list['clusters'][0] == cluster['id'])
+
+        # Delete non-existent cluster
+        with self.assertRaises(requests.exceptions.HTTPError):
```
nandajavarma: @anushshetty Nice use of assertRaises as context manager :+1:

client/api/python/heketi/heketi.py
```diff
@@ -44,13 +52,156 @@ def _set_token_in_header(self, headers, method, uri):
...
+        return headers
+
 
     def hello(self):
         method = 'GET'
         uri = '/hello'
 
         headers={}
-        self._set_token_in_header(headers, method, uri)
+        self._set_token_in_header(method, uri)
```
lpabon: This should be: 

```python
headers = self._set_token_in_header(method, uri)
```

client/api/python/heketi/heketi.py
```diff
@@ -44,13 +52,156 @@ def _set_token_in_header(self, headers, method, uri):
...
         r = requests.get(self.host + uri, headers=headers)
         return r.status_code == requests.codes.ok
 
+
+    def _make_request(self, method, uri, data={}):
+        self.method = method
```
lpabon: Never copy data to the object if it is local to the function.  Here the function is copying data to the object as if it was to use it at another time outside this function.

client/api/python/heketi/heketi.py
```diff
@@ -44,13 +52,156 @@ def _set_token_in_header(self, headers, method, uri):
...
         uri = '/hello'
 
         headers={}
-        self._set_token_in_header(headers, method, uri)
+        self._set_token_in_header(method, uri)
         r = requests.get(self.host + uri, headers=headers)
         return r.status_code == requests.codes.ok
 
+
+    def _make_request(self, method, uri, data={}):
```
lpabon: What if the caller wanted to pass their own headers in some future release?  The function could have ```headers = {}```

lpabon: Documentation for this function should specify that the data should be in Python Dictionary format and not in JSON format already.

client/api/python/heketi/heketi.py
```diff
@@ -44,13 +52,156 @@ def _set_token_in_header(self, headers, method, uri):
...
+                # Temporary request of 303 with header info
+                response_ready = True
+                vol_req = q.headers['location']
+                return self._make_request('GET', vol_req)
+
+            elif q.status_code == requests.codes.NO_CONTENT:
+                # Request successful
+                return True
+                response_ready = True
+
```
lpabon: There should be an ```else``` sleep 1 to 5 seconds.  If not, the server will have to keep answering as fast as the client is requesting.

client/api/python/heketi/heketi.py
```diff
@@ -44,13 +52,156 @@ def _set_token_in_header(self, headers, method, uri):
...
+        while response_ready is False:
+            q = requests.get(self.host + queue_uri,
+                             headers=headers,
+                             allow_redirects=False)
+
+            # Raise an exception when the request fails
+            q.raise_for_status()
+
+
+            if q.status_code == requests.codes.see_other and q.headers['location']:
```
lpabon: I think this logic can be simplified by setting ```allow_redirects=True```.

For example:

```python

q.raise_for_status()

if 'X-Pending' in q.headers:
    sleep 1
else
   return q
```

See https://github.com/heketi/heketi/blob/master/client/api/go-client/client.go#L120 
and
See https://github.com/heketi/heketi/wiki/API#async

client/api/python/heketi/heketi.py
```diff
@@ -44,13 +52,156 @@ def _set_token_in_header(self, headers, method, uri):
...
+                return True
+            else:
+                return r.json()
+
+        elif r.status_code == requests.codes.NO_CONTENT:
+            return True
+
+
+
+    def _get_queued_response(self, queue_uri):
```
lpabon: This function returns multiple formats.  Sometimes a request object, sometimes a boolean.  Functions should be setup to return a single expected type.

client/api/python/heketi/heketi.py
```diff
@@ -44,13 +52,156 @@ def _set_token_in_header(self, headers, method, uri):
...
+            if q.status_code == requests.codes.see_other and q.headers['location']:
+                # Temporary request of 303 with header info
+                response_ready = True
+                vol_req = q.headers['location']
+                return self._make_request('GET', vol_req)
+
+            elif q.status_code == requests.codes.NO_CONTENT:
+                # Request successful
+                return True
+                response_ready = True
```
nandajavarma: Isn't this dead code? Why assign anything after return?

anushshetty: oops, yes! Will fix.

client/api/python/heketi/heketi.py
anushshetty: Sure. Making this function return the entire response.

client/api/python/heketi/heketi.py
anushshetty: Sure, I thought we should be diligently checking for temp responses.  If that's not necessary, this makes it simple.

client/api/python/heketi/heketi.py
```diff
@@ -44,13 +52,156 @@ def _set_token_in_header(self, headers, method, uri):
...
 
-        headers={}
-        self._set_token_in_header(headers, method, uri)
+        headers = self._set_token_in_header(method, uri)
         r = requests.get(self.host + uri, headers=headers)
         return r.status_code == requests.codes.ok
 
+
+    def _make_request(self, method, uri, data={}, headers={}):
+        headers = self._set_token_in_header(method, uri)
```
lpabon: This overwrites headers if it was passed in.  I think it should append to the dictionary instead:

```python
headers.update(self._set.....())
```

lpabon:
I think it looks great.  I'll make that simple change.  Thanks a lot @anushshetty  and @nandajavarma !

Issue #133, opened 2015-08-09
Adding an object should check for conflict


lpabon:
For example if adding a node, it should check if the node has already been added, and return statusConflict or something like that.

PR #259, opened 2016-02-18
Conflict management support

lpabon:
Heketi now checks to make sure that nodes or devices in nodes
are not added again to the database.  If a node is added with the
same host name, it will return HTTP Status Conflict.  For devices,
device names on the same node will return HTTP Status Conflict.


lpabon:
Closes #133

PR #282, opened 2016-04-13
cli with cobra for posix support

MohamedAshiqrh:


lpabon:
Looks like a few errors.  First is an error  shown on line 511 of the log https://travis-ci.org/heketi/heketi/jobs/122923461 .  Second is that TestClientNode is running forever which causes the test to time out

GLOCKFILE
```diff
@@ -8,4 +8,6 @@ github.com/heketi/rest 738570ea73f1cc2a0888c7c49fe80991c6050f34
 github.com/heketi/tests f3775cbcefd6822086c729e3ce4b70ca85a5bd21
 github.com/heketi/utils 3d373e21f062fe8e80e4d75e2b1282a96e5d6e23
 github.com/lpabon/godbc 9577782540c1398b710ddae1b86268ba03a19b0c
+github.com/spf13/cobra 4c05eb1145f16d0e6bb4a3e1b6d769f4713cb41f
```
lpabon: Still need pflag?

GLOCKFILE
```diff
@@ -8,4 +8,6 @@ github.com/heketi/rest 738570ea73f1cc2a0888c7c49fe80991c6050f34
...
+github.com/spf13/pflag 7f60f83a2c81bc3c3c0d5297f61ddfa68da9d3b7
```
lpabon: Still need pflag?

MohamedAshiqrh: cobra makes use of pflag ( https://github.com/spf13/cobra#flags )

lpabon:
Few issues found while trying it out:

* `-v` or `--version` does not work
* Global flags are printed on command error, when they shouldn't.  For example, trying to communicate with a server which does not exist, or adding a device which has already been added:

```
$ ./heketi-cli -s http://localhost:8080 device add --name=/dev/vvv --node=$NODE1
Error: Device /dev/vvv is already used on node 0b11abea078fc89dde97622b4e51888b by device 6cecbafd97561f93fddfd65d4d43bc4b

Usage:
  heketi-cli device add [flags]

Flags:
      --name string   Name of device to add
      --node string   Id of the node which has this device

Global Flags:
      --json
        Print response as JSON
      --secret string
        Secret key for specified user.  Can also be
        set using the environment variable HEKETI_CLI_KEY
  -s, --server string
        Heketi server. Can also be set using the
        environment variable HEKETI_CLI_SERVER
      --user string
        Heketi user.  Can also be set using the
        environment variable HEKETI_CLI_USER
  
Device /dev/vvv is already used on node 0b11abea078fc89dde97622b4e51888b by device 6cecbafd97561f93fddfd65d4d43bc4b
```

* Server URL is printed when creating a cluster:

```
$ ./heketi-cli -s http://localhost:8080 cluster create
http://localhost:8080
Cluster id: 7366f7329df8a26ee01dc11cc2716b78
```

* No Examples

client/cli/go/commands/tests/sample.json
```diff
@@ -1,100 +0,0 @@
-{
```
lpabon: This file should probably not be deleted.

MohamedAshiqrh: I have this file in client/cli/go/cmds/tests/sample.json, Will add that with the PR for cli tests. I will add this file with this PR if it is block it. Can I know why is sample.json is required here?

humblec:
@lpabon @MohamedAshiqrh do u think this has to wait till the topology command is merged?

client/cli/go/commands/tests/sample.json
lpabon: It was supposed to be used for unit tests and the load command

lpabon:
@humblec I think that we need this before topology, if not, we would have to change that also. You may want to base your work on Ashiq's branch, which will make it easier to merge later.

lpabon:
@MohamedAshiqrh Also, do you mind providing a summary git commit comment here on the github comments?  I will use it when the code is read and I merge the code.

MohamedAshiqrh:
Few issues found while trying it out:

    -v or --version does not work
    Global flags are printed on command error, when they shouldn't. For example, trying to communicate with a server which does not exist, or adding a device which has already been added:

$ ./heketi-cli -s http://localhost:8080 device add --name=/dev/vvv --node=$NODE1
Error: Device /dev/vvv is already used on node 0b11abea078fc89dde97622b4e51888b by device 6cecbafd97561f93fddfd65d4d43bc4b

Usage:
  heketi-cli device add [flags]

Flags:
      --name string   Name of device to add
      --node string   Id of the node which has this device

Global Flags:
      --json
        Print response as JSON
      --secret string
        Secret key for specified user.  Can also be
        set using the environment variable HEKETI_CLI_KEY
  -s, --server string
        Heketi server. Can also be set using the
        environment variable HEKETI_CLI_SERVER
      --user string
        Heketi user.  Can also be set using the
        environment variable HEKETI_CLI_USER

Device /dev/vvv is already used on node 0b11abea078fc89dde97622b4e51888b by device 6cecbafd97561f93fddfd65d4d43bc4b

    Server URL is printed when creating a cluster:

$ ./heketi-cli -s http://localhost:8080 cluster create
http://localhost:8080
Cluster id: 7366f7329df8a26ee01dc11cc2716b78

    No Examples

MohamedAshiqrh:
FYI, cobra supports suggestions.

```
./heketi-cli cluser              
Error: unknown command "cluser" for "heketi-cli"

Did you mean this?
	cluster

Run 'heketi-cli --help' for usage.

```

lpabon:
@MohamedAshiqrh Ignore AppVeyor.. I was trying it out to determine if it could save builds....but it is only for windows.

lpabon:
Global flags are printed on command error, when they shouldn't. For example, trying to communicate with a server which does not exist, or adding a device which has already been added.

    we can not avoid global flags as they are default in help command thats by cobra's design.

@MohamedAshiqrh There must be a way because kubectl and oc do not do this

And there must be a way to do examples also.

lpabon:
@MohamedAshiqrh When you make the new NewHeketiCliCmd() function, also pass in stdout, and stderr, that way we can add unit tests and inspect the output.

client/cli/go/cmds/load.go
```diff
@@ -0,0 +1,110 @@
...
+		// Load config file
+		fp, err := os.Open(jsonConfigFile)
+		if err != nil {
+			return errors.New("Unable to open config file")
+		}
+		defer fp.Close()
+
+		configParser := json.NewDecoder(fp)
+		var topology ConfigFile
+		if err = configParser.Decode(&topology); err != nil {
```
nandajavarma: Initializing and then the conditional! That's one pretty feature to have in a language. :+1: 
I would recommend you use this syntax in all the similar contexts! :-)

MohamedAshiqrh:
Have added a separate Usage to be displayed for error messages, This custom usage does not print Global Flags. Kubectl has lot of implementation which was hard to trace out the solution for this particular problem. So created a custom usage which will be used specific to  error message. 

Added Examples for the commands too.

lpabon:
Great, it looks good.  The help screen still prints during error conditions, which we need to fix, but we can do that after this fix is merged.  That way @humblec can continue with his work.

Since there are a lot of commits here, please squash them and create a nice git commit message and send as a new pull request. Follow the information here:  https://github.com/heketi/heketi/wiki/Development-Guide#squashing-multiple-commits-into-one

MohamedAshiqrh:
@lpabon I was wondering you were suggesting for help screen without global flags during error conditions. I have now completely removed the help screen for error conditions. This is last commit and will create a new PR.

MohamedAshiqrh:
sent a new PR https://github.com/heketi/heketi/pull/283
closing this one.

PR #291, opened 2016-04-26
Pvcreate should force creation

lpabon:
This will ensure that if a raw disk had a partition it will destroy it

### DIFF
--- a/executors/sshexec/device.go
+++ b/executors/sshexec/device.go
@@ -40,7 +40,7 @@ func (s *SshExecutor) DeviceSetup(host, device, vgid string) (d *executors.Devic
 
 	// Setup commands
 	commands := []string{
-		fmt.Sprintf("sudo pvcreate --metadatasize=128M --dataalignment=256K %v", device),
+		fmt.Sprintf("sudo pvcreate -f --metadatasize=128M --dataalignment=256K %v", device),
 		fmt.Sprintf("sudo vgcreate %v %v", s.vgName(vgid), device),
 	}
 


lpabon:
Too dangerous.  It could destroy the user's data.

dewet22:
This just bit me as well, with the topology creation just sitting around forever with no indication what's going wrong. When I `docker exec`ed into the glusterd container and `kill -9`ed the `pvcreate` command, the topology command gave something more useful:

```
heketi-cli topology load --json=topology.json
	Found node xxx01 on cluster 578ff09c16ebb66f0a82e2415e66b796
		Adding device /dev/sdb ... Unable to add device: Unable to execute command on glusterfs-wwsld: WARNING: xfs signature detected on /dev/sdb at offset 0. Wipe it? [y/n]:
	Creating node xxx02 ... ID: 227fa1ee4c2fd4b13df6f508f97fdfd9
		Adding device /dev/sdb ... OK
...
```

*At the very least* the documentation should be updated to reflect that the disk signature should be wiped (if `-f` is considered too dangerous). I used `dd if=/dev/zero of=/dev/sdb bs=512 count=1` myself.

lpabon:
@dewet22 Yeah, I agree it is a pain, but a bigger pain would be loosing data by mistakenly adding a drive with data.

dewet22:
I guess it depends on your use case. In ours we have a throwaway cluster that gets rebuilt automatically and stores no long-term data, so that prompt is just getting in the way.

lpabon:
@dewet22 Ah, if you delete all the clusters (by deleting all volumes, them removing all devices, then removing all nodes) using the Heketi API, you will be able to cleanly remove them and then add them. This is what we do in our tests.

See https://github.com/heketi/heketi/blob/master/tests/functional/TestSmokeTest/tests/heketi_test.go#L166

Issue #295, opened 2016-04-28
heketi-cli version is not showing up when the host variable is not set


lpabon:
```
$ ./heketi-cli --version
Server must be provided
```

MohamedAshiqrh:
https://github.com/heketi/heketi/pull/303

PR #303, opened 2016-05-02
heketi-cli: version error fixed

MohamedAshiqrh:
Version was throwing error "Server must be provided"
Version does not require heketi server as it specific to only cli.


MohamedAshiqrh:
https://github.com/heketi/heketi/issues/295

Issue #265, opened 2016-03-23
Node API needs Kubernetes support


lpabon:
This is required for #264 .

Issue #264, opened 2016-03-17
New executor to use Kubernetes Exec API


lpabon:

lpabon:
See https://github.com/gluster/aplo/blob/9665afce89707e094a32b7b71e6a6029ec246c29/src/ams/main.go

PR #312, opened 2016-05-10
Kubernetes Exec API Executor

lpabon:
* Kubeexec is a "subclass" of Sshexec
* Environment variables can be used to set configuration
* SshExec configuration must specify private key
* Gluster app now checks executor started correctly


lpabon:
Closes #265 
Closes #264

PR #330, opened 2016-05-17
fix:correct usage of backup-volfile-servers mount option

pcuzner:
Just watched a heketi demo and noticed this anomaly with the backup-volfile-server mount option

### DIFF
--- a/apps/glusterfs/models.go
+++ b/apps/glusterfs/models.go
@@ -181,14 +181,14 @@ func (v *VolumeInfoResponse) String() string {
 		"Id: %v\n"+
 		"Cluster Id: %v\n"+
 		"Mount: %v\n"+
-		"Mount Options: backupvolfile-servers=%v\n"+
+		"Mount Options: backup-volfile-servers=%v\n"+
 		"Durability Type: %v\n",
 		v.Name,
 		v.Size,
 		v.Id,
 		v.Cluster,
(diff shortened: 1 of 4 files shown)


lpabon:
Thanks for the pull request Paul, but from the manpage it looks like it supports both ways:

```
       backup-volfile-servers=SERVERLIST
              Provide list of backup volfile servers in the following format [default: None]

              $ mount -t glusterfs -obackup-volfile-servers=<server2>:
                     <server3>:...:<serverN> <server1>:/<volname> <mount_point>

       backupvolfile-server=SERVER
              Provide list of backup volfile servers in the following format [default: None]

               $ mount -t glusterfs -obackupvolfile-server=<server2>
                       <server1>:/<volname> <mount_point>
```

I'm not sure this change is needed

pcuzner:
The option you have now is incorrect - it specifies backupvolfile-servers...not backupvolfile-server - so syntactically it is currently invalid.

The two options allow either
a single server to be contacted as an alternate (backupvolfile-server=<server>)
OR
multiple servers to be tried using backup-volfile-servers=<server2>:<server3>:<server4> 

They are subtly different parameters - not interchangeable. For heketi, I think the 2nd form is the better option to improve mount reliability.

lpabon:
Ah! I see now :+1:

Issue #293, opened 2016-04-27
Brick cleanup deletes the volume even if it fails with "Volume <volname> already exists"


NehaRawat:
Try to create a volume with existing volume name. It fails with error "Stderr [volume create: vol23: failed: Volume vol23 already exists" and run the brick cleanup.

I think related commit https://github.com/heketi/heketi/commit/dd2a70dbfd7d3ee0a58d39799ed13bce260d4bd1

Should give a error and exit if the volume is already up and running, instead of deleting it.

lpabon:
Nice bug! Normally we do not set the volume name since that is not normal usage, so I would have not noticed this.  I'll investigate further.

lpabon:
The issue is that when the volume creation fails, it calls volume.Destroy() for that volume name.  What we should do is instead check the return value of https://github.com/heketi/heketi/blob/master/apps/glusterfs/volume_entry.go#L300 . If it is something like "ErrVolumeNameAlreadyExists" then it should return that error.  Then the error should be checked in https://github.com/heketi/heketi/blob/master/apps/glusterfs/volume_entry.go#L307 so that it doesn't call v.Destroy()

PR #336, opened 2016-05-19
Avoid deleting existing volume if there is an attempt to create the s…

humblec:
…ame volume again.

Signed-off-by: hchiramm <hchiramm@redhat.com>


executors/sshexec/volume.go
```diff
@@ -77,7 +78,12 @@ func (s *SshExecutor) VolumeCreate(host string,
 	// Execute command
 	_, err := s.RemoteExecutor.RemoteCommandExecute(host, commands, 10)
 	if err != nil {
-		s.VolumeDestroy(host, volume.Name)
+		volexist := "Volume " + volume.Name + " already exists"
+		if strings.Contains(err.Error(), volexist) {
+			fmt.Printf("\n Volume exist, Please recheck the volume create command")
```
lpabon: Cannot use fmt.Printf() This goes to stdout.

lpabon:
This method is really an educated guess and not the correct way to determine if there is another volume with that name.

humblec:
@lpabon, as long as the error message remains same from Gluster , this should work.

lpabon:
That is still not accurate enough.  Instead I commented how to do it here: https://github.com/heketi/heketi/issues/293#issuecomment-220877895 .

humblec:
@lpabon can you please review/merge this patch?

lpabon:
Let's discuss this tomorrow.

humblec:
@lpabon let me know whenever you are ready to discuss this.

humblec:
@lpabon can you review/merge this patch ?

lpabon:
@humblec Let me take a look if the issue is still there.

lpabon:
This fix is not longer necessary for Issue #293, because it was fixed by PR #388.

humblec:
@lpabon looks like #388 came after this fix was proposed. Didnt know about #388 is fixing this issue as well. Is there a way we can track the patches. Otherwise, its really waste of effort and dont want to come to this same situation again. :(

PR #353, opened 2016-05-26
set sudovar if its a non root user and execute gluster commands.

humblec:
If its root user dont set the sudovar.

### DIFF
--- a/GLOCKFILE
+++ b/GLOCKFILE
@@ -25,7 +25,7 @@ github.com/gorilla/context a8d44e7d8e4d532b6a27a02dd82abb31cc1b01bd
 github.com/gorilla/mux 0eeaf8392f5b04950925b8a69fe70f110fa7cbfc
 github.com/heketi/rest 738570ea73f1cc2a0888c7c49fe80991c6050f34
 github.com/heketi/tests f3775cbcefd6822086c729e3ce4b70ca85a5bd21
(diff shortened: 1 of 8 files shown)


humblec:
@lpabon can you please review/merge this asap?

lpabon:
@humblec Hi humble.  I get notified by email and Github when new pull requests are given.  If I do not provide a comment for more than a 5 days, then please ping me.

I have reviewed this fix on Friday, but I did not commit because we need to be really careful when editing the code which handles communication with the servers.  One of the main things I noticed is that sudo is still in one of the commands because it is past a `|`.  This would cause issues on systems without sudo.

Because we are so close to release and the code is not waiting for `sudo` fix, I would like instead to postpone in to Release 3.

This fix does provide two fixes.  One is the `sudo` fix, the other was better SSH failure strings.  Please make a new patch just to get the SSH failure strings, and we will fix `sudo` after Release 2.

humblec:
@lpabon  
>This fix does provide two fixes. One is the sudo fix, the other was better SSH failure strings. Please make a new patch just to get the SSH failure strings, and we will fix sudo after Release 2.
>
iic, you are referring the GLOCKFILE change with 'better SSH failure strings'. Please confirm, so that I can create a PR with only that change.

lpabon:
We will revisit this in Release 3.  The package utils is now part of Heketi as of #366 .

lpabon:
No longer needed

Issue #355, opened 2016-05-27
heketi-cli: proper error message for cluster/node/device delete


NehaRawat:
It is not possible to delete cluster without deleting nodes, devices and volumes.  

=> docker exec heketi heketi-cli --server http://localhost:8080 cluster delete 46d613cf778d1a11ab2f36a2dd3dc803
Error: Conflict

Logs:
[negroni] Started DELETE /clusters/46d613cf778d1a11ab2f36a2dd3dc803
[heketi] WARNING 2016/05/27 12:01:01 Unable to delete cluster [46d613cf778d1a11ab2f36a2dd3dc803] because it contains volumes and/or nodes
[negroni] Completed 409 Conflict in 466.033µs

=>docker exec heketi heketi-cli --server http://localhost:8080 node delete 53bb0c58ceaa2042c5088377de9c872f
Error: Conflict

Logs:
[negroni] Started DELETE /nodes/53bb0c58ceaa2042c5088377de9c872f
[negroni] Completed 409 Conflict in 323.013µs

=> docker exec heketi heketi-cli --server http://localhost:8080 device delete 77fddc8c8ed815ab09965095941cf899
Error: Conflict

Logs:
[negroni] Started DELETE /devices/77fddc8c8ed815ab09965095941cf899
[negroni] Completed 409 Conflict in 368.05µs

Should display proper error message to STDOUT.
For nodes and devices it doesn't log proper message in logs too. Creating a new issue to track it because if not wrong https://github.com/heketi/utils/commit/34e5a97154d1d016635ad6ce436a68fdc8a75876 will not fix this one.

lpabon:
Good point. This should be fixed.

PR #356, opened 2016-05-27
Fixing Logging issue for Node/Device delete

NehaRawat:
Signed-off-by: Neha rawat <nerawat@redhat.com>

### DIFF
--- a/apps/glusterfs/app_device.go
+++ b/apps/glusterfs/app_device.go
@@ -225,9 +225,10 @@ func (a *App) DeviceDelete(w http.ResponseWriter, r *http.Request) {
 		}
 
 		// Check if we can delete the device
-		if !device.IsDeleteOk() {
-			http.Error(w, ErrConflict.Error(), http.StatusConflict)
-			return ErrConflict
+		err = device.Delete(tx)
+		if err == ErrConflict {
+			http.Error(w, err.Error(), http.StatusConflict)
+			return err
 		}
 
 		// Access node entry
--- a/apps/glusterfs/app_node.go
+++ b/apps/glusterfs/app_node.go
@@ -226,9 +226,10 @@ func (a *App) NodeDelete(w http.ResponseWriter, r *http.Request) {
 		}
 
 		// Check the node can be deleted
-		if !node.IsDeleteOk() {
-			http.Error(w, ErrConflict.Error(), http.StatusConflict)
-			return ErrConflict
+		err = node.Delete(tx)
+		if err == ErrConflict {
+			http.Error(w, err.Error(), http.StatusConflict)
+			return err
 		}
 
 		// Access cluster information and peer node


NehaRawat:
Fix for logging  https://github.com/heketi/heketi/issues/355

apps/glusterfs/app_device.go
```diff
@@ -225,9 +225,10 @@ func (a *App) DeviceDelete(w http.ResponseWriter, r *http.Request) {
 		}
 
 		// Check if we can delete the device
-		if !device.IsDeleteOk() {
-			http.Error(w, ErrConflict.Error(), http.StatusConflict)
-			return ErrConflict
+		err = device.Delete(tx)
+		if err == ErrConflict {
+			http.Error(w, err.Error(), http.StatusConflict)
+			return err
```
lpabon: I'm not sure what this is fixing. It returns the same error.

apps/glusterfs/app_node.go
```diff
@@ -226,9 +226,10 @@ func (a *App) NodeDelete(w http.ResponseWriter, r *http.Request) {
 		}
 
 		// Check the node can be deleted
-		if !node.IsDeleteOk() {
-			http.Error(w, ErrConflict.Error(), http.StatusConflict)
-			return ErrConflict
+		err = node.Delete(tx)
+		if err == ErrConflict {
+			http.Error(w, err.Error(), http.StatusConflict)
+			return err
```
lpabon: Same here.

lpabon:
I think to really fix this issue, you need to fix heket-cli which gets a HTTP Status Conflict from the server, then print the string in the HTTP body which should show why there was a conflict.

NehaRawat:
@lpabon the check is already there. 

func (d *DeviceEntry) Delete(tx *bolt.Tx) error {
        godbc.Require(tx != nil)

        // Check if the devices still has drives
        if !d.IsDeleteOk() {
                logger.Warning("Unable to delete device [%v] because it contains bricks", d.Info.Id)
                return ErrConflict
        }

        return EntryDelete(tx, d, d.Info.Id)
}

Same for node

lpabon:
Ah, let me re-review.

lpabon:
@NehaRawat Ok, I reviewed again, but I think this does not fix the issue.  The issue we are trying to fix is that we need better strings in the returned HTTP body.  This patch returns the exact same HTTP body string of ErrConflict, because ErrConlict.Error() == err.Error() when err == ErrConflict.

According to #355, you pointed out that `Should display proper error message to STDOUT`.  This is true, and to fix that, we should fix it in `heketi-cli`, not the server.

NehaRawat:
@lpabon there are two issues one is not redirecting error to stdout and other not logging errors for node/device delete.

for cluster delete logs
[negroni] Started DELETE /clusters/46d613cf778d1a11ab2f36a2dd3dc803
[heketi] WARNING 2016/05/27 12:01:01 Unable to delete cluster [46d613cf778d1a11ab2f36a2dd3dc803] because it contains volumes and/or nodes
[negroni] Completed 409 Conflict in 466.033µs

for node/device

it doesn't log it
[negroni] Started DELETE /nodes/53bb0c58ceaa2042c5088377de9c872f
[negroni] Completed 409 Conflict in 323.013µs

And I think it should be fixed in server. Might be to avoid confusion ll create a new issue for that ?

lpabon:
Yes, that would be better. Ok. So this fix is about correct logging.  I will review accordingly.

apps/glusterfs/app_device.go
```diff
@@ -225,9 +225,10 @@ func (a *App) DeviceDelete(w http.ResponseWriter, r *http.Request) {
 		}
 
 		// Check if we can delete the device
-		if !device.IsDeleteOk() {
-			http.Error(w, ErrConflict.Error(), http.StatusConflict)
```
lpabon: Instead of the lines below, just add a logger.Warning(...) here

apps/glusterfs/app_node.go
```diff
@@ -226,9 +226,10 @@ func (a *App) NodeDelete(w http.ResponseWriter, r *http.Request) {
 		}
 
 		// Check the node can be deleted
-		if !node.IsDeleteOk() {
-			http.Error(w, ErrConflict.Error(), http.StatusConflict)
```
lpabon: Same here.

lpabon: @NehaRawat I have been looking at this a little more, and have determined that the error returned by the server should be fixed also.

When I wrote this, I didn't know about the call `fmt.Errorf(...)` (even though it is not useful exactly here.  Here instead we will use fmt.Sprintf() since we need to return a string instead of an error). Now I am starting to use that instead of `ErrXXX` errors.

I suggest something like the following:

```go
if !node.IsDeleteOk() {
    http.Error(w, fmt.Sprintf("Unable to delete node [%v] because it contains volumes", id), http.StatusConflict)
}
```

NehaRawat:
Thanks @lpabon. Agree this is the correct way of doing it.

lpabon:
@NehaRawat I'm not sure if you have time to work on this issue, if not, I do not mind taking it over.

lpabon:
Closed by #498

PR #358, opened 2016-05-30
Fix stdout errors for cluster/node/device delete

NehaRawat:
Signed-off-by: Neha rawat <nerawat@redhat.com>


lpabon:
@MohamedAshiqrh In the previous CLI model, `err` was printed to the screen as a string.  Now with cobra.RunE(...), what happens to the err that is returned?  Does RunE print it?

MohamedAshiqrh:
Yeah, RunE does handles err. I thought the returned value can be used to log. for now the return value lands [here ]( heketi/client/cli/go/main.go ).

PR #395, opened 2016-06-24
functional-tests: skip sudo when already running as root

nixpanic:
When running tests in the CentOS CI with the SCL for Vagrant, 'sudo'
cleans the environment to much. This prevents the run.sh scripts from
executing vagrant (not found in $PATH).

The new _sudo() function in lib.sh replaces the normal 'sudo' calls. The
function checks if the tests are run as root already, in which case it
will skip calling 'sudo'.

A few more details on using Vagrant with CentOS can be found on this
wiki page:
  https://wiki.centos.org/SpecialInterestGroup/SCLo/Vagrant

Signed-off-by: Niels de Vos <ndevos@redhat.com>


lpabon:
Closes #396

PR #401, opened 2016-06-30
Use go 1.4.2 in CI because CentOS still uses it

lpabon:


lpabon:
Seems like 1.4.2 has many issues with the latest godep.

PR #408, opened 2016-07-04
Adding services and endpoints sample files.

humblec:
Signed-off-by: Humble Chirammal <hchiramm@redhat.com>


lpabon:
@humblec Do you mind renaming the files as glusterfs-endpoints-sample.json or something like that?  The important part is to show that these are samples, and not the actual files.  If you want, you can also add a README.md in the directory providing some doc for developers with a pointer to other documentation on the internet describing this.

lpabon:
ok to test

PR #410, opened 2016-07-05
heketi-cli: heketi-cli command to return help

MohamedAshiqrh:

### DIFF
--- a/client/cli/go/cmds/root.go
+++ b/client/cli/go/cmds/root.go
@@ -47,6 +47,8 @@ var RootCmd = &cobra.Command{
 	Run: func(cmd *cobra.Command, args []string) {
 		if version {
 			fmt.Printf("heketi-cli %v\n", HEKETI_CLI_VERSION)
+		} else {
+			cmd.Usage()
 		}
 	},
 }
@@ -73,7 +75,7 @@ func initConfig() {
 	// Check server
 	if options.Url == "" {
 		options.Url = os.Getenv("HEKETI_CLI_SERVER")
-		if options.Url == "" && !version {
+		if options.Url == "" && !version && !(len(RootCmd.Flags().Args()) <= 0) {
 			fmt.Fprintf(stderr, "Server must be provided\n")
 			os.Exit(3)
 		}


lpabon:
@MohamedAshiqrh could you describe this a little more, I am not sure what this is fixing.

lpabon:
test this please

MohamedAshiqrh:
# heketi-cli                                  
Usage:
  heketi-cli [flags]
  heketi-cli [command]

Examples:
  $ export HEKETI_CLI_SERVER=http://localhost:8080
  $ heketi-cli volume list

Available Commands:
  cluster                        Heketi cluster management
  device                         Heketi device management
  setup-openshift-heketi-storage Setup OpenShift/Kubernetes persistent storage for Heketi
  node                           Heketi Node Management
  topology                       Heketi Topology Management
  volume                         Heketi Volume Management

Flags:
  -h, --help[=false]: help for heketi-cli
      --json[=false]: 
	Print response as JSON
      --log-flush-frequency=5s: Maximum number of seconds between log flushes
      --secret="": 
	Secret key for specified user.  Can also be
	set using the environment variable HEKETI_CLI_KEY
  -s, --server="": 
	Heketi server. Can also be set using the
	environment variable HEKETI_CLI_SERVER
      --user="": 
	Heketi user.  Can also be set using the
	environment variable HEKETI_CLI_USER
  -v, --version[=false]: 
	Print version

Use "heketi-cli [command] --help" for more information about a command.

lpabon:
@MohamedAshiqrh awesome

lpabon:
retest this please

lpabon:
@MohamedAshiqrh FYI, the command "test this please" and "retest this please" are not for you, they are commands for the CI system :smile: 

https://ci.centos.org/view/Gluster/job/gluster_heketi-functional/

lpabon:
Maybe I should do:

@centos-ci retest this please

That's probably better ;-)

client/cli/go/cmds/crawl.xml
```diff
@@ -0,0 +1 @@
+<url>"tool"</url>
```
lpabon: Is this file needed?

lpabon:
This is not working for me:

```
(heketi) lpabon@localhost : ~/git/golang/heketi/src/github.com/heketi/heketi/client/cli/go
$ ./heketi-cli 
Server must be provided


(heketi) lpabon@localhost : ~/git/golang/heketi/src/github.com/heketi/heketi/client/cli/go
$ ./heketi-cli -v
heketi-cli v2.0.3-dev-2-ge414e7a-pr_410

$ git log --oneline | head
e414e7a heketi-cli: heketi-cli command to return help

```

client/cli/go/cmds/crawl.xml
MohamedAshiqrh: No this is not needed.

MohamedAshiqrh:
I had server already set, I think this patch needs little more changes to make help work without server set.

lpabon:
Without the server set, now I get this error:

```
$ ./heketi-cli cluster list
Error: Get /clusters: unsupported protocol scheme ""
```

lpabon:
When changing the behavior, please check that all other calls are unaffected.

obnoxxx:
I think that disabling the "Server must be provided" message is a different issue.
I agree that it does not make a lot of sense to have this message instead of the usage when just the naked heketi-cli cmd is called. But currently having the server set is a prerequisite for further processing, so imho a fix for this could be valid if it works with this assumption.

MohamedAshiqrh:
@obnoxxx yeah, I am not sure about the way to fix this. But in search of the fix have learned that there can be a lot of improvements in cli. IMO, The easy-fix can be changing the version from flag to command. only operation that heketi-cli command does is prints the version, If we remove that from heketi-cli command. Then it will print usage and version will be a command instead of flag.
`# heketi-cli --version` to `# heketi-cli version`

MohamedAshiqrh:
Hey,

So the reason why the help is not being printed on 'heketi-cli' invoke is because do version handling. Was looking into kubectl code, They have really cool cli with utils, templates, etc. which can implemented in future. For our current problem, how they handle is they write a bash function which will be invoked on setting a flag. I am trying this [one](https://godoc.org/github.com/spf13/cobra#Command.MarkFlagCustom) will let know how it goes.

https://github.com/heketi/heketi/pull/475

obnoxxx:
@MohamedAshiqrh Ok. I see now what you mean with the version. I think it is not necessary to change --version to a command in order to have the root cmd print usage. I have set up a PR #477 which containes a modification of your earlier patch which works for me. The point is that your check for RootCmd.Flags().Args() does not seem to have done the trick. I check os.Args[1:] instead, which may not be the best or correct solution, but works for me.

lpabon:
I will review #477

obnoxxx:
Closes #415

obnoxxx:
But superseded by #477

lpabon:
Closed by #477

PR #438, opened 2016-07-20
Set snapshot limit using environment variable

### LINKED ISSUE #419: Heketi should have a configurable value for setting up max number of snapshots in a volume
There needs to be a configurable value in the config file and in the environment variable which will be able to set the max number of snapshots in a volume

lpabon:
Closes #419


MohamedAshiqrh:
@lpabon Hey, can I know where this value been used or set.

I can not find the place it sets the value on gluster snapshot

Issue #445, opened 2016-08-02
golang panic when running `heketi-cli load <topology.json-file>`


jeffvance:
go panic:
=======
```
# heketi-cli topology load --json=erin-topo.json
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x0 pc=0x469d25]

goroutine 1 [running]:
panic(0xc6b280, 0xc82000e0e0)
	/usr/lib/golang/src/runtime/panic.go:481 +0x3e6
github.com/heketi/heketi/client/cli/go/cmds.getNodeIdFromHeketiTopology(0x0, 0xc8201c6820, 0xc, 0xc820126f60)
	/builddir/build/BUILD/heketi-dd29da6/src/github.com/heketi/heketi/client/cli/go/cmds/topology.go:70 +0x2c5
github.com/heketi/heketi/client/cli/go/cmds.glob.func17(0x13f5800, 0xc8201c66f0, 0x0, 0x1, 0x0, 0x0)
	/builddir/build/BUILD/heketi-dd29da6/src/github.com/heketi/heketi/client/cli/go/cmds/topology.go:137 +0x89b
github.com/spf13/cobra.(*Command).execute(0x13f5800, 0xc8201c6680, 0x1, 0x1, 0x0, 0x0)
	/builddir/build/BUILD/heketi-dd29da6/Godeps/_workspace/src/github.com/spf13/cobra/command.go:568 +0x62c
github.com/spf13/cobra.(*Command).ExecuteC(0x13f5400, 0x13f5800, 0x0, 0x0)
	/builddir/build/BUILD/heketi-dd29da6/Godeps/_workspace/src/github.com/spf13/cobra/command.go:662 +0x53f
github.com/spf13/cobra.(*Command).Execute(0x13f5400, 0x0, 0x0)
	/builddir/build/BUILD/heketi-dd29da6/Godeps/_workspace/src/github.com/spf13/cobra/command.go:618 +0x2d
main.main()
	/builddir/build/BUILD/heketi-dd29da6/client/cli/go/main.go:35 +0xe0
```

JSON file:
```
# cat erin-topo.json  #passes several json validators
{
    "clusters": [
        {
            "nodes": [
                {
                    "node": {
                        "hostnames": {
                            "manage": [
                                "aplo-rhel7.2"
                            ],
                            "storage": [
                                "192.168.122.10"
                            ]
                        },
                        "zone": 1
                    },
                    "devices": [
[...]

PR #454, opened 2016-08-12
Update topology.go

humblec:


client/cli/go/cmds/topology.go
```diff
@@ -126,6 +126,9 @@ var topologyLoadCommand = &cobra.Command{
 
 		// Load current topolgy
 		heketiTopology, err := heketi.TopologyInfo()
+		if err != nil {
+			return err
```
lpabon: Please reword return a nice error and append err to it.  For example:

```go
return fmt.Errorf("Unable to get topology information: %v\n", err)
```

Also, there seems to be a go fmt error.  Please fix

lpabon:
@humblec Is this to fix #445?

obnoxxx:
I was providing a patch for issue #445 and did not realize this PR was already a fix for that.

@humblec in addition to @lpabon 's comments, please use a more descriptive commit message, e.g.:

```
heketi-cli: fix panic in 'topology load' if TolologyInfo() fails

React to error from TopologyInfo function.

Closes #445
```

Something like this is helpful than having the commit message say that it changes 'topology.go'. It will be much more easy for someone to tell what this patch is about from looking at the git log. It will also link the PR with the issue!

Thanks - Michael

humblec:
@obnoxxx Totally agree, Actually I failed to put a commit message in this PR and iic, 'update topology.go' is deafult commit message in github :)

client/cli/go/cmds/topology.go
```diff
@@ -127,7 +127,8 @@ var topologyLoadCommand = &cobra.Command{
 		// Load current topolgy
 		heketiTopology, err := heketi.TopologyInfo()
 		if err != nil {
-			return err
+			fmt.Fprintf(stdout, "Unable to get topology information\n")
+			return fmt.Errorf("Unable to get topology information: %v", err)
                 }
```
obnoxxx: There is still an indentation error here.
Otherwise LGTM. Thanks, @humblec .

needs to be squashed with previous commit but keeping this commit msg.

obnoxxx:
@humblec Ok, that makes more sense. I still wonder how you can commit to git with an empty commit message . At least my git aborts the commit when I try that.  And for all I know, I need to push a local commit to my github fork before I can create a PR. ... :-)

obnoxxx:
@humblec Can we please accelerate the trivial formatting fix for this PR. This issue lets the tests fail sometimes, and it would be good to fix it... :-)

humblec:
@obnoxxx hmmm. making me to reveal the truth abt empty commit message. :).  I directly made that commit from github UI :+1:  .  Now I have squashed and sent the PR and its ready to merge :+1:

obnoxxx:
@humblec Thanks, this looks good to me now! :+1: 

But -- oh gosh -- how can you make commits directly in github UI, that sounds creepy! :-o

client/cli/go/cmds/topology.go
```diff
@@ -126,6 +126,10 @@ var topologyLoadCommand = &cobra.Command{
 
 		// Load current topolgy
 		heketiTopology, err := heketi.TopologyInfo()
+		if err != nil {
+			fmt.Fprintf(stdout, "Unable to get topology information\n")
```
lpabon: No need to print out.  This will happen automatically when the error is returned.

lpabon:
@humblec Also please describe the reason for the change.

obnoxxx:
@lpabon isn't the commit message (fix panic in toplology load if TopologyInfo fails) enough of reason for the change?

lpabon:
@obnoxxx The issue is that the git commit log has no information on what it is fixing.

obnoxxx:
@lpabon yes, the update patch has a commit msg. precisely the one that I proposed.

https://github.com/heketi/heketi/pull/454/commits/683bb883ce59aff652bedf643e1f2c171faf80b0

lpabon:
@humblec I added the commit header from @obnoxxx .  Please make sure to update the header next time to make sure we link it to the issue which it closes.

Issue #462, opened 2016-09-01
When topology file is loaded, we get an error saying pvcreate: command not found


raghavendra-talur:
When `heketi-cli topology load --json=topology_libvirt.json` is executed, we get the error mentioned in the subject.

The reason seems to be how we are executing the command over ssh. The fix would probably be in 
`heketi/pkg/utils/ssh/ssh.go`.

raghavendra-talur:
This would happen only when the "user" as mentioned in the heketi.json file is not a root user and sudo: true is not set in sshexec section.

raghavendra-talur:
Closing this now, you can refer to https://github.com/heketi/heketi/issues/462#issuecomment-244045472 as possible reason for hitting this bug.

PR #463, opened 2016-09-01
make bash behave as login shell

raghavendra-talur:
When bash is executed over ssh, if -l is not used it does not behave
like a login shell causing differences in $PATH and ENV variables.


raghavendra-talur:
This is fix for #462

raghavendra-talur:
I am not sure if we require this. Bug https://github.com/heketi/vagrant-heketi/issues/11 does not need this fix.

We can probably close this for now. Unless we notice a problem with ENV vars or $PATH for root we don't need this patch.

lpabon:
Closed

Issue #466, opened 2016-09-02
Allow <10G volume creation in Heketi


humblec:
At present, if we try to create volumes < 10G , the user get 'NoSpace' error. We have to remove this limitation as there can be volume creation requests of small size in Openshift/Kubernetes.

obnoxxx:
It seems 8G is the magic threshold in my 4-node vagrant-heketi test setup:

```
[vagrant@storage0 ~]$ heketi-cli volume create --size=2
Error: No space

[vagrant@storage0 ~]$ heketi-cli volume create --size=3
Error: No space

[vagrant@storage0 ~]$ heketi-cli volume create --size=4
Error: No space

[vagrant@storage0 ~]$ heketi-cli volume create --size=5
Error: No space

[vagrant@storage0 ~]$ heketi-cli volume create --size=6
Error: No space

[vagrant@storage0 ~]$ heketi-cli volume create --size=7
Error: No space

[vagrant@storage0 ~]$ heketi-cli volume create --size=8
Name: vol_c76ed273f064db70ba44256fd6d83e4f
Size: 8
Volume Id: c76ed273f064db70ba44256fd6d83e4f
Cluster Id: bac92e2f4adf87159d193576acf7667a
Mount: 192.168.10.100:vol_c76ed273f064db70ba44256fd6d83e4f
Mount Options: backup-volfile-servers=192.168.10.101,192.168.10.103
Durability Type: replicate
Distributed+Replica: 3
[vagrant@storage0 ~]$ 
```

humblec:
Yep, even I was able to create 8G volume at times, but not always. But I never got success in <8G

obnoxxx:
Better error propagation may already help here?

https://github.com/obnoxxx/heketi/commit/df19766fe8d3cec20d590c92ebfa155c2e7c177f

https://github.com/obnoxxx/heketi/tree/error-propagation

lpabon:
See #331

humblec:
@obnoxxx @lpabon regardless of error propagation, we have to document this . Because there is no way for a user to know what is the minimum volume size and how the "min_brick_size" is affectging the minimum volume size.

Issue #331, opened 2016-05-18
No space error if creating replica 3 volume on two node cluster


NehaRawat:
Try to create volume with default replica configuration on two node cluster, it will fail with "No space" error though disks have enough space.

heketi-cli volume create --name=vol3 --size=10 
Error: No space

```
May 18 11:12:28 hostname heketi: [heketi] ERROR 2016/05/18 11:12:28 /src/github.com/heketi/heketi/apps/glusterfs/volume_entry_allocate.go:41: Minimum brick size limit reached.  Out of space.
May 18 11:12:28 hostname heketi: [heketi] ERROR 2016/05/18 11:12:28 /src/github.com/heketi/heketi/apps/glusterfs/app_volume.go:128: Failed to create volume: No space

```

if not wrong its  allocBricks function which is returning this error if there is no device found.

 // No devices found
      return ErrNoSpace

There is no functional issue as heketi will not create two replica bricks on same node/device. But it should print proper error message.

lpabon:
Interesting.  Do you mind describing the topology you used?  Also, fyi, try not to use `--name=`, it could create issues.  I'm almost thinking about removing it.

NehaRawat:
* docker exec heketi heketi-cli --server http://localhost:8080 topology info

Cluster Id: 46d613cf778d1a11ab2f36a2dd3dc803

    Volumes:

    Nodes:

	Node Id: 53bb0c58ceaa2042c5088377de9c872f
	Cluster Id: 46d613cf778d1a11ab2f36a2dd3dc803
	Zone: 1
	Management Hostname: 10.70.43.225
	Storage Hostname: 10.70.43.225
	Devices:
		Id:77fddc8c8ed815ab09965095941cf899   Name:/dev/sde            Size (GiB):499     Used (GiB):0       Free (GiB):499     
			Bricks:

	Node Id: d3502d07f469ce0eef1e1a11b16d3756
	Cluster Id: 46d613cf778d1a11ab2f36a2dd3dc803
	Zone: 1
	Management Hostname: 10.70.43.245
	Storage Hostname: 10.70.43.245
	Devices:
		Id:c84bd8d40a04a6e08058bc4a97cccafe   Name:/dev/sdc            Size (GiB):499     Used (GiB):0       Free (GiB):499     
			Bricks:

* docker exec heketi heketi-cli --server http://localhost:8080 volume create --size=10
Error: No space

lpabon:
[...]

PR #470, opened 2016-09-02
app:glusterfs: better error propagation from brick allocation

obnoxxx:

### DIFF
--- a/apps/glusterfs/volume_entry.go
+++ b/apps/glusterfs/volume_entry.go
@@ -304,6 +304,11 @@ func (v *VolumeEntry) Create(db *bolt.DB,
 			break
 		}
 	}
+
+	if err != nil {
+		return err
+	}
+
 	if brick_entries == nil {
 		return ErrNoSpace
 	}


obnoxxx:
Not sure if it's correct.
But it may bring some clarity to issues #466  and #331 .

obnoxxx:
ok. patch is buggy... doesn't compile, var out of scope.  closing. will re-PR once I have a better patch ;-)

apps/glusterfs/volume_entry.go
```diff
@@ -304,6 +304,11 @@ func (v *VolumeEntry) Create(db *bolt.DB,
 			break
 		}
 	}
+
+	if err != nil {
+		return err
+	}
+
```
lpabon: I think the correct fix will probably be to add what you have, but then move the logic of _ErrNoSpace_ to v.allocBricksInCluster(), then change line 300 to say something like: if err == ErrNoSpace, then continue, else get out!

lpabon: You will also need to make sure unit tests are adjust to expect the correct paths to test the change

obnoxxx:
@lpabon : I have been looking over the code quite some time now. Here are my thoughts:

The v.allocBricksInCluster() function does currently return three different errors that indicate that the current cluster can provide space for the requested volume size:
- ErrNoSpace
- ErrMinimumBrickSize
- ErrMaxBricks
These errors justify continuing with the next cluster as a candidate for the volume.
All other errors are genuine, serious errors, like database errors,  that justify aborting the creation.

So, your proposal is to map the three 'valid' errors to ErrNoSpace in v.allocBricksInCluster(), and then just treat the ErrNoSpace case in the caller (Create, or Expand) ?

It may be worth keeping the different errors possibly returned from allocBricksInCluster(), for the better error propagation. (At least the unit tests for expand use /test that.) In that case, the loop in Create() just needs to check three 'valid' errors instead of one.

On the other hand, if we mapped the valid errors to ErrNoSpace in allocBricksInCluster(), then we could just get rid of ErrMaxBricks and ErrMinumumBrickSize. That would be a simplification, but a loss of granularity in error reporting and possibly debuggability?

Not 100% convinced yet what the better change would be, but I'll propose a patch next that will keep the granularity.. :-)

PR #472, opened 2016-09-02
app:glusterfs: better error propagation from brick allocation

obnoxxx:

### DIFF
--- a/apps/glusterfs/volume_entry.go
+++ b/apps/glusterfs/volume_entry.go
@@ -304,7 +304,7 @@ func (v *VolumeEntry) Create(db *bolt.DB,
 		} else if err == ErrNoSpace ||
 			err == ErrMaxBricks ||
 			err == ErrMinimumBrickSize {
-			logger.Debug("Cluster %v can not accomodate volume " +
+			logger.Debug("Cluster %v can not accomodate volume "+
 				"(%v), trying next cluster", cluster, err)
 			continue
 		} else {
--- a/apps/glusterfs/volume_entry.go
+++ b/apps/glusterfs/volume_entry.go
@@ -291,20 +291,33 @@ func (v *VolumeEntry) Create(db *bolt.DB,
 
 	// For each cluster look for storage space for this volume
 	var brick_entries []*BrickEntry
+	var err error
 	for _, cluster := range clusters {
-		var err error
 
 		// Check this cluster for space
 		brick_entries, err = v.allocBricksInCluster(db, allocator, cluster, v.Info.Size)
 
-		// Check if allocation was successful
 		if err == nil {
 			v.Info.Cluster = cluster
 			logger.Debug("Volume to be created on cluster %v", cluster)
 			break
+		} else if err == ErrNoSpace ||
+			err == ErrMaxBricks ||
+			err == ErrMinimumBrickSize {
+			logger.Debug("Cluster %v can not accomodate volume " +
+				"(%v), trying next cluster", cluster, err)
+			continue
+		} else {
+			// A genuine error occurred - bail out
+			return logger.LogError("Error calling v.allocBricksInCluster: %v", err)
 		}
 	}
-	if brick_entries == nil {
+
+	if err != nil || brick_entries == nil {
+		// Map all 'valid' errors to NoSpace here:
+		// Only the last such error could get propagated down,
+		// so it does not make sense to hand the granularity on.
+		// But for other callers (Expand), we keep it.
 		return ErrNoSpace
 	}
 
@@ -321,7 +334,7 @@ func (v *VolumeEntry) Create(db *bolt.DB,
 	}()
 
 	// Create the bricks on the nodes
-	err := CreateBricks(db, executor, brick_entries)
+	err = CreateBricks(db, executor, brick_entries)
 	if err != nil {
 		return err
 	}


lpabon:
@obnoxxx See code review comments in #470

obnoxxx:
Now this one actually works. Would hopefully shed some light on why creating a volume fails. No automated tests for heketi-cli yet. I need to figure out how to test this.

obnoxxx:
Oops sorry, missed the comments on #470. Sounds good -- will redo.

obnoxxx:
updated the patches as discussed (see #470).
This contains the rename patch from #473 because it's a dependency.

obnoxxx:
Gosh, seems like I did had not figured out yet how to run the unit tests...

obnoxxx:
figured it out, will push updated patchset soon

obnoxxx:
Updated the patchset. Figured out what the problem was: Tried to be clever and use switch instead of if .... Bad thouh that 'break' only exits the switch block and not the enclosing for loop ... ;-)

obnoxxx:
This patch is not perfect in the sense that it does not have unit test yet for the changed behavior. My problem is that in order to test the new aspects, I need to create a test that lets v.allocBricksInCluster() return a 'real' error, i.e. something other than ErrNoSpace, ErrMinimumBrickSize, and ErrMaxBricks. Not sure yet how to fake up such a behavior.

obnoxxx:
Hmm now this is strange: the tests are failing, but all I can see which looks like a failure is this:

```
The command "go fmt $(go list ./... | grep -v vendor) | wc -l | grep 0" exited with 1.
```

Any hints? ...

obnoxxx:
Also, after all this back and forth with this patch, which has been very instructive indeed, I am not fully convinced any more whether this is actually the right thing to do..

What does it change:

in the loop over the clusters to find bricks to accomodate a volume of given size:
- errors from probing a given cluster are logged / treated more explicitly (which is a good thing)
- 'valid' errors (NoSpace, MaxBricks, MinimumBrickSize) are (apart from logging) treated as before, i.e. try next cluster
- 'real' errors, like database errors (what else?) lead to aborting the loop and exiting the function and the error is preserved instead of mapping it to ErrNoSpace ==> This is the real change in behavior. I thought that this was appropriate. Not 100% certain about that. @lpabon: what's your thought?

If we do agree that this would be an improvement, how do I insert such failures that can come from the db operations or possibly brick creation? Hints appreciated.

obnoxxx:
@lpabon any thoughts?

apps/glusterfs/volume_entry.go
```diff
@@ -291,19 +291,38 @@ func (v *VolumeEntry) Create(db *bolt.DB,
...
+				"(%v), trying next cluster", cluster, err)
+			// Map all 'valid' errors to NoSpace here:
+			// Only the last such error could get propagated down,
+			// so it does not make sense to hand the granularity on.
+			// But for other callers (Expand), we keep it.
+			err = ErrNoSpace
+			continue
+		} else {
+			// A genuine error occurred - bail out
+			logger.LogError("Error calling v.allocBricksInCluster: %v", err)
```
lpabon: If we are bailing out, we can just return here, right?  With Jose's change, you can use `return logger.LogError(...)`

apps/glusterfs/volume_entry.go
```diff
@@ -291,19 +291,38 @@ func (v *VolumeEntry) Create(db *bolt.DB,
...
+			break
 		}
 	}
+
+	if err != nil {
+		return err
+	}
+
 	if brick_entries == nil {
```
lpabon: Could there be a situation where `brick_entries` is nil and `err==nil` ?  Should this be before `err != nil` ?

apps/glusterfs/volume_entry.go
obnoxxx: True. Gonna update. I will also make the whole patch a tad more simple.

apps/glusterfs/volume_entry.go
obnoxxx: This was deliberate. I wanted a 'real' error (from `allocBricksInCluster()`) to be treated with prio. But with bailing out directly as commented above, the distintion is not necessary any more... updated patch will follow.

obnoxxx:
Updated patch addressing above requests.

lpabon:
@obnoxxx how can the patch be updated if there is only one commit?

lpabon:
ok to test @centos-ci

obnoxxx:
I amended the patch an force-pushed over.

So you now have the complete, improved patch instead of an additional patch which you would have to squash (which drives me crazy if I am the reviewer). 

I also don't like github to do these things, I want to do them myself. If I push separate patches I am proposing them exactly as is, and don't want them squashed. That's my thinking... maybe I am special ;-)

lpabon:
@obnoxxx yeah, there is no need to do that. Plus, we lose the changes from in between pushes.  Don't worry, trust in Git and Github. In fact, Github follows the git workflow, unlike Gerrit.

Amend? What is that? :smile:

lpabon:
retest this please @centos-ci

lpabon:
@obnoxxx looks like a go fmt error.  Use [vim-go](https://github.com/fatih/vim-go) if you are using Vim.

obnoxxx:
@lpabon amend: to improve. see git commit --amend. same effect as committing and then squashing into.

I dont' fully trust github (yet?) (don't trust gerrit either, and like it a lot less). 

Also imho github follows one possible git workflow, which is the one it imposes (which is different from the flow I am used to).

All that being said, working on a github based project, I do of course have to abide.
But I still don't like my patches being squashed for me by github.
And also it's inconsistent: if you use squash in the ui, you don't get a merge commit. otherwise you do.

It also confuses me when I agree that it is good to be able to see the diff between two revisions of a patch, but I can do that locally. too. And what is as least as important to me is to see the overall diff if it is meant to be _one_ commit in the end. That can be done locally as well, of course.

It takes some getting used to ... *sigh*

obnoxxx:
Okay... I _added_ a patch to fix the formatting. This should be squashed with the actual patch... :-)

PR #492, opened 2016-09-13
Allow LogError to return an error from the supplied format string.

### LINKED ISSUE #491: Create a logger which acts like fmt.Errorf
What we would like to see is a logging function which would not just log the current error, but also return it.  It is similar to logger.Err(error), with the exception, that it takes string formatting.

jarrpa:
Closes #491

Signed-off-by: Jose A. Rivera <jarrpa@redhat.com>

### DIFF
--- a/pkg/utils/log.go
+++ b/pkg/utils/log.go
@@ -114,10 +114,12 @@ func (l *Logger) Critical(format string, v ...interface{}) {
 }
 
 // Log error string
-func (l *Logger) LogError(format string, v ...interface{}) {
+func (l *Logger) LogError(format string, v ...interface{}) error {
 	if l.level >= LEVEL_ERROR {
 		logWithLongFile(l.errorlog, format, v...)
 	}
+
+	return fmt.Errorf(format, v...)
 }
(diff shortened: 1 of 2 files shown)


obnoxxx:
@jarrpa I'm confused: why 2 pull requests (this and #493) with an identical patch? Just by accident?

jarrpa:
This one is wrong, it contains a typo.

obnoxxx:
@jarrpa Ah!. Hint: you could have pushed over this branche and that would have updated this pull request. No need to close and create a new one.

lpabon:
@jarrpa Yes, please create new branches for pull requests.

obnoxxx:
@lpabon Oh, just to make sure I don't mistunderstand: I was suggesting that @jarrpa pushes over the same branche when updating a pull request instead of creating a new pull request (possibly by creating a new branch for each revision of the patch).

lpabon:
@obnoxxx @jarrpa Ah, I see.  As long as it is not merged, that is definitely the process.

PR #495, opened 2016-09-14
functional/TestSmokeTest: use 'become' instead of deprecated 'sudo'

obnoxxx:


lpabon:
ok to test @centos-ci

PR #498, opened 2016-09-15
Improve error handling and reporting when deleting items.

### LINKED ISSUE #355: heketi-cli: proper error message for cluster/node/device delete

jarrpa:
Allow more useful error messages to be reported and logged when attempting to
delete clusters, nodes, and devices in the glusterfs app. Minor changes also
to simplify and bring consistency to error reporting within related functions.

Closes #355

Signed-off-by: Jose A. Rivera <jarrpa@redhat.com>

### DIFF
--- a/apps/glusterfs/app_cluster.go
+++ b/apps/glusterfs/app_cluster.go
@@ -18,10 +18,11 @@ package glusterfs
 
 import (
 	"encoding/json"
+	"net/http"
+
 	"github.com/boltdb/bolt"
 	"github.com/gorilla/mux"
 	"github.com/heketi/heketi/pkg/glusterfs/api"
-	"net/http"
 )
 
 func (a *App) ClusterCreate(w http.ResponseWriter, r *http.Request) {
@@ -143,11 +144,12 @@ func (a *App) ClusterDelete(w http.ResponseWriter, r *http.Request) {
 		}
 
 		err = entry.Delete(tx)
-		if err == ErrConflict {
-			http.Error(w, err.Error(), http.StatusConflict)
-			return err
-		} else if err != nil {
-			http.Error(w, err.Error(), http.StatusInternalServerError)
+		if err != nil {
+			if err == ErrConflict {
+				http.Error(w, entry.ConflictString(), http.StatusConflict)
+			} else {
+				http.Error(w, err.Error(), http.StatusInternalServerError)
+			}
 			return err
 		}
 
--- a/apps/glusterfs/app_cluster_test.go
+++ b/apps/glusterfs/app_cluster_test.go
@@ -334,6 +334,7 @@ func TestClusterDelete(t *testing.T) {
 	r, err := http.DefaultClient.Do(req)
 	tests.Assert(t, err == nil)
 	tests.Assert(t, r.StatusCode == http.StatusConflict)
+	tests.Assert(t, utils.GetErrorFromResponse(r).Error() == entries[0].ConflictString())
 
 	// Check that we cannot delete a cluster with volumes
 	req, err = http.NewRequest("DELETE", ts.URL+"/clusters/"+"a2", nil)
@@ -341,6 +342,7 @@ func TestClusterDelete(t *testing.T) {
 	r, err = http.DefaultClient.Do(req)
 	tests.Assert(t, err == nil)
 	tests.Assert(t, r.StatusCode == http.StatusConflict)
+	tests.Assert(t, utils.GetErrorFromResponse(r).Error() == entries[1].ConflictString())
 
 	// Check that we cannot delete a cluster with nodes
 	req, err = http.NewRequest("DELETE", ts.URL+"/clusters/"+"a3", nil)
@@ -348,6 +350,7 @@ func TestClusterDelete(t *testing.T) {
 	r, err = http.DefaultClient.Do(req)
 	tests.Assert(t, err == nil)
 	tests.Assert(t, r.StatusCode == http.StatusConflict)
+	tests.Assert(t, utils.GetErrorFromResponse(r).Error() == entries[2].ConflictString())
 
 	// Delete cluster with no elements
 	req, err = http.NewRequest("DELETE", ts.URL+"/clusters/"+"000", nil)
--- a/apps/glusterfs/app_device.go
+++ b/apps/glusterfs/app_device.go
@@ -225,7 +225,7 @@ func (a *App) DeviceDelete(w http.ResponseWriter, r *http.Request) {
 
 		// Check if we can delete the device
 		if !device.IsDeleteOk() {
-			http.Error(w, ErrConflict.Error(), http.StatusConflict)
+			http.Error(w, device.ConflictString(), http.StatusConflict)
 			return ErrConflict
 		}
 
--- a/apps/glusterfs/app_node.go
+++ b/apps/glusterfs/app_node.go
@@ -227,7 +227,7 @@ func (a *App) NodeDelete(w http.ResponseWriter, r *http.Request) {
 
 		// Check the node can be deleted
 		if !node.IsDeleteOk() {
-			http.Error(w, ErrConflict.Error(), http.StatusConflict)
+			http.Error(w, node.ConflictString(), http.StatusConflict)
 			return ErrConflict
 		}
 
--- a/apps/glusterfs/cluster_entry.go
+++ b/apps/glusterfs/cluster_entry.go
@@ -19,6 +19,7 @@ package glusterfs
 import (
 	"bytes"
 	"encoding/gob"
+	"fmt"
 	"sort"
 
 	"github.com/boltdb/bolt"
@@ -77,12 +78,16 @@ func (c *ClusterEntry) Save(tx *bolt.Tx) error {
 	return EntrySave(tx, c, c.Info.Id)
 }
 
+func (c *ClusterEntry) ConflictString() string {
+	return fmt.Sprintf("Unable to delete cluster [%v] because it contains volumes and/or nodes", c.Info.Id)
+}
+
 func (c *ClusterEntry) Delete(tx *bolt.Tx) error {
 	godbc.Require(tx != nil)
 
-	// Check if the nodes still has drives
+	// Check if the cluster still has nodes or volumes
 	if len(c.Info.Nodes) > 0 || len(c.Info.Volumes) > 0 {
-		logger.Warning("Unable to delete cluster [%v] because it contains volumes and/or nodes", c.Info.Id)
+		logger.Warning(c.ConflictString())
 		return ErrConflict
 	}
 
--- a/apps/glusterfs/device_entry.go
+++ b/apps/glusterfs/device_entry.go
@@ -148,12 +148,16 @@ func (d *DeviceEntry) IsDeleteOk() bool {
 	return true
 }
 
+func (d *DeviceEntry) ConflictString() string {
+	return fmt.Sprintf("Unable to delete device [%v] because it contains bricks", d.Info.Id)
+}
+
 func (d *DeviceEntry) Delete(tx *bolt.Tx) error {
 	godbc.Require(tx != nil)
 
 	// Check if the devices still has drives
 	if !d.IsDeleteOk() {
-		logger.Warning("Unable to delete device [%v] because it contains bricks", d.Info.Id)
+		logger.Warning(d.ConflictString())
 		return ErrConflict
 	}
 
--- a/apps/glusterfs/node_entry.go
+++ b/apps/glusterfs/node_entry.go
@@ -159,12 +159,16 @@ func (n *NodeEntry) IsDeleteOk() bool {
 	return true
 }
 
+func (n *NodeEntry) ConflictString() string {
+	return fmt.Sprintf("Unable to delete node [%v] because it contains devices", n.Info.Id)
+}
+
 func (n *NodeEntry) Delete(tx *bolt.Tx) error {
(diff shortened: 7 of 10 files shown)


apps/glusterfs/app_cluster.go
```diff
@@ -139,16 +141,17 @@ func (a *App) ClusterDelete(w http.ResponseWriter, r *http.Request) {
...
 
 		err = entry.Delete(tx)
-		if err == ErrConflict {
-			http.Error(w, err.Error(), http.StatusConflict)
-			return err
-		} else if err != nil {
-			http.Error(w, err.Error(), http.StatusInternalServerError)
-			return err
+		if err != nil {
+			if strings.Contains(err.Error(), "Unable to delete") {
```
lpabon: If an error is being checked, then it must be a variable saved in errors.go

apps/glusterfs/cluster_entry_test.go
```diff
@@ -248,7 +249,7 @@ func TestNewClusterEntrySaveDelete(t *testing.T) {
 		return nil
 
 	})
-	tests.Assert(t, err == ErrConflict)
+	tests.Assert(t, strings.Contains(err.Error(), "Unable to delete"))
```
lpabon: Same here

apps/glusterfs/cluster_entry_test.go
```diff
@@ -277,7 +278,7 @@ func TestNewClusterEntrySaveDelete(t *testing.T) {
...
```
lpabon: Same here

lpabon:
The change is based on the style of checking for strings in err.  That should be avoided.  If an error is going to be checked, then it must have a value.  Most of these are stored in errors.go.

There are parts of the fix which add logging to errors, those are cool, but changing ErrConflict does not seem to add benefit.

jarrpa:
But the entire point of the issue is that the error message is useless. It is also impossible to check the current useful error message against a constant because the error message is generated per ID. The main alternative would be to remove ErrConflict entirely and create unique conflict errors for each entity that needs them. We would still lose the ID of the conflicting item, but that might be an okay compromise?

lpabon:
Let's see if there is another way to fix this issue.  Let's talk on Monday.

apps/glusterfs/app_cluster.go
```diff
@@ -143,12 +143,13 @@ func (a *App) ClusterDelete(w http.ResponseWriter, r *http.Request) {
...
-		} else if err != nil {
-			http.Error(w, err.Error(), http.StatusInternalServerError)
-			return err
+		if err != nil {
+			if err == ErrConflict {
+				http.Error(w, entry.ErrConflictFmt(), http.StatusConflict)
+			} else {
+				http.Error(w, err.Error(), http.StatusInternalServerError)
+			}
+			return logger.Err(err)
```
lpabon: This is cool, but like the #503 we should not log ERROR on conflict.  ERROR is more of a "Hey this is messed up, fix this, i don't know what to do!".  Also, although I didn't point this out in #503, most likely there is no need to log errors in app_XXX .

apps/glusterfs/node_entry.go
```diff
@@ -159,13 +159,16 @@ func (n *NodeEntry) IsDeleteOk() bool {
 	return true
 }
 
+func (n *NodeEntry) ErrConflictFmt() error {
```
lpabon: Interesting style, but I would name it something different because ErrXXX is normally for errors and XXXFmt() is normally for something like the string in Printf( "fmt..").  How about:

```go
func (n *XXX) ErrorMsg( err error ) string {

   switch err {
   case ErrConflict:
       return fmt.Sprintf("...")
...
   }
}
```

lpabon: Or you can call it ErrorInfo, ErrorString... you get the idea.

apps/glusterfs/node_entry.go
```diff
@@ -159,13 +159,16 @@ func (n *NodeEntry) IsDeleteOk() bool {
...
+}
+
 func (n *NodeEntry) Delete(tx *bolt.Tx) error {
 	godbc.Require(tx != nil)
 
 	// Check if the nodes still has drives
 	if !n.IsDeleteOk() {
-		logger.Warning("Unable to delete node [%v] because it contains devices", n.Info.Id)
-		return ErrConflict
+		return logger.Err(ErrConflict)
```
lpabon: Here, this is not an ERROR.  Same as the others.  You may want to have logger.Warning (and others) return `error`.  You can do that in another PR, and then have this one use that.

pkg/utils/log.go
```diff
@@ -138,6 +138,15 @@ func (l *Logger) Warning(format string, v ...interface{}) {
 	}
 }
 
+// Log error variable as a warning
+func (l *Logger) Warn(err error) error {
```
lpabon: Awesome.  This needs unit tests.  You may want to name the ones that take error something similar.  For example, we already have `Err()` and `LogError()`, so this could be `WarnErr()` or `WarningError()` or something like that.

apps/glusterfs/node_entry.go
```diff
@@ -164,8 +164,7 @@ func (n *NodeEntry) Delete(tx *bolt.Tx) error {
 
 	// Check if the nodes still has drives
 	if !n.IsDeleteOk() {
-		logger.Warning("Unable to delete node [%v] because it contains devices", n.Info.Id)
```
lpabon: We are loosing logging information here.  These lines should not be removed.

apps/glusterfs/device_entry.go
```diff
@@ -153,8 +153,7 @@ func (d *DeviceEntry) Delete(tx *bolt.Tx) error {
 
 	// Check if the devices still has drives
 	if !d.IsDeleteOk() {
-		logger.Warning("Unable to delete device [%v] because it contains bricks", d.Info.Id)
```
lpabon: We are loosing logging information here.  These lines should not be removed.

apps/glusterfs/cluster_entry.go
```diff
@@ -80,10 +80,9 @@ func (c *ClusterEntry) Save(tx *bolt.Tx) error {
 func (c *ClusterEntry) Delete(tx *bolt.Tx) error {
 	godbc.Require(tx != nil)
 
-	// Check if the nodes still has drives
+	// Check if the cluster still has nodes or volumes
 	if len(c.Info.Nodes) > 0 || len(c.Info.Volumes) > 0 {
-		logger.Warning("Unable to delete cluster [%v] because it contains volumes and/or nodes", c.Info.Id)
```
lpabon: We are loosing logging information here.  These lines should not be removed.

apps/glusterfs/app_node.go
```diff
@@ -227,7 +228,7 @@ func (a *App) NodeDelete(w http.ResponseWriter, r *http.Request) {
 
 		// Check the node can be deleted
 		if !node.IsDeleteOk() {
-			http.Error(w, ErrConflict.Error(), http.StatusConflict)
+			http.Error(w, fmt.Sprintf("Unable to delete node [%v] because it contains devices", node.Info.Id), http.StatusConflict)
```
lpabon: Good, this needs unit test to check.

apps/glusterfs/app_device.go
```diff
@@ -226,7 +227,7 @@ func (a *App) DeviceDelete(w http.ResponseWriter, r *http.Request) {
 
 		// Check if we can delete the device
 		if !device.IsDeleteOk() {
-			http.Error(w, ErrConflict.Error(), http.StatusConflict)
+			http.Error(w, fmt.Sprintf("Unable to delete device [%v] because it contains bricks", device.Info.Id), http.StatusConflict)
```
lpabon: Good, this needs unit test to check.

apps/glusterfs/app_cluster.go
```diff
@@ -143,11 +145,12 @@ func (a *App) ClusterDelete(w http.ResponseWriter, r *http.Request) {
...
 
 		err = entry.Delete(tx)
-		if err == ErrConflict {
-			http.Error(w, err.Error(), http.StatusConflict)
-			return err
-		} else if err != nil {
-			http.Error(w, err.Error(), http.StatusInternalServerError)
+		if err != nil {
+			if err == ErrConflict {
+				http.Error(w, fmt.Sprintf("Unable to delete cluster [%v] because it contains volumes and/or nodes", entry.Info.Id), http.StatusConflict)
```
lpabon: Good, this needs unit test to check.

pkg/utils/log.go
lpabon: But you may want to do this in another PR.  I do not think you need this in this PR.

jarrpa:
Moved the first patch into #509.

jarrpa:
Regarding the loss of logging information: would a reasonable compromise be to change the error text of ErrConflict from "Conflict" to "The target either contains other items or is in use."? Or do you want to maintain the current level of verbosity in the logs for the XXX_entry.go files?

apps/glusterfs/app_cluster.go
```diff
@@ -18,10 +18,12 @@ package glusterfs
 
 import (
 	"encoding/json"
+	"fmt"
+	"net/http"
+
```
lpabon: Nice! thanks :smile:

lpabon:
I like both :smile: .  Make the change to change from just `Conflict` to `The target either contains other items or is in use`.  And keep the current logs because they log the ID which had the issue.

lpabon:
One more thing.  These changes need unit tests updates to verify them.

jarrpa:
Did both, added tests. I don't directly test the ConflictString() functions right now, since they are one-line functions and the test would just be to copy out the one line. Instead I test they work by calling them for the verification of the REST response body content.

apps/glusterfs/app_cluster_test.go
```diff
@@ -334,20 +334,23 @@ func TestClusterDelete(t *testing.T) {
...
 	tests.Assert(t, r.StatusCode == http.StatusConflict)
+	tests.Assert(t, utils.GetErrorFromResponse(r).Error() == entries[1].ConflictString())
 
 	// Check that we cannot delete a cluster with nodes
 	req, err = http.NewRequest("DELETE", ts.URL+"/clusters/"+"a3", nil)
 	tests.Assert(t, err == nil)
 	r, err = http.DefaultClient.Do(req)
 	tests.Assert(t, err == nil)
 	tests.Assert(t, r.StatusCode == http.StatusConflict)
+	tests.Assert(t, utils.GetErrorFromResponse(r).Error() == entries[2].ConflictString())
```
lpabon: nice

apps/glusterfs/cluster_entry.go
```diff
@@ -77,12 +78,16 @@ func (c *ClusterEntry) Save(tx *bolt.Tx) error {
...
+}
+
 func (c *ClusterEntry) Delete(tx *bolt.Tx) error {
 	godbc.Require(tx != nil)
 
-	// Check if the nodes still has drives
+	// Check if the cluster still has nodes or volumes
 	if len(c.Info.Nodes) > 0 || len(c.Info.Volumes) > 0 {
-		logger.Warning("Unable to delete cluster [%v] because it contains volumes and/or nodes", c.Info.Id)
+		logger.Warning(c.ConflictString())
```
lpabon: Ah nice!

lpabon:
Code looks good.  I just need to verify it.  I'll get to that tomorrow.

jarrpa:
Great! Once this gets merged we should probably also close #356.

lpabon:
Awesome!

```
$ ./heketi-cli -s http://localhost:8080 node delete c2304ee9667c36ad4dc6b0c11b954bb6
Error: Unable to delete node [c2304ee9667c36ad4dc6b0c11b954bb6] because it contains devices

$ ./heketi-cli -s http://localhost:8080 cluster delete 9d525bf98fa9c20816c3cdcfdfa70f64
Error: Unable to delete cluster [9d525bf98fa9c20816c3cdcfdfa70f64] because it contains volumes and/or nodes
```

lpabon:
Closes #356

PR #499, opened 2016-09-15
Two minor patches

jarrpa:
Incidental stuff I found while working on #355. Not required for it, but I figured I may as well see if they're worth taking anyway. :)

### DIFF
--- a/apps/glusterfs/app_node_test.go
+++ b/apps/glusterfs/app_node_test.go
@@ -352,7 +352,6 @@ func TestNodeAddDelete(t *testing.T) {
 	r, err = http.Post(ts.URL+"/nodes", "application/json", bytes.NewBuffer(request))
 	tests.Assert(t, err == nil)
 	tests.Assert(t, r.StatusCode == http.StatusConflict)
-	tests.Assert(t, err == nil)
 
 	// Check Cluster has node
 	r, err = http.Get(ts.URL + "/clusters/" + clusterinfo.Id)
--- a/pkg/utils/bodystring.go
+++ b/pkg/utils/bodystring.go
@@ -21,6 +21,7 @@ import (
 	"io"
 	"io/ioutil"
 	"net/http"
+	"strings"
 )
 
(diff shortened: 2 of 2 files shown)


pkg/utils/bodystring.go
```diff
@@ -18,14 +18,14 @@ package utils
...
-	"io"
 	"io/ioutil"
 	"net/http"
+	"strings"
 )
 
 // Return the body from a response as a string
 func GetStringFromResponse(r *http.Response) (string, error) {
-	body, err := ioutil.ReadAll(io.LimitReader(r.Body, r.ContentLength))
+	body, err := ioutil.ReadAll(r.Body)
```
lpabon: We need the limit to make sure we do not read forever.  So this line does not need to be fixed.

pkg/utils/bodystring.go
```diff
@@ -39,5 +39,5 @@ func GetErrorFromResponse(r *http.Response) error {
 	if err != nil {
 		return err
 	}
-	return errors.New(s)
+	return errors.New(strings.TrimSpace(s))
```
lpabon: This is nice!

apps/glusterfs/app_node_test.go
```diff
@@ -352,7 +352,6 @@ func TestNodeAddDelete(t *testing.T) {
 	r, err = http.Post(ts.URL+"/nodes", "application/json", bytes.NewBuffer(request))
 	tests.Assert(t, err == nil)
 	tests.Assert(t, r.StatusCode == http.StatusConflict)
-	tests.Assert(t, err == nil)
```
lpabon: Nice catch

jarrpa:
@lpabon correctly identified the need to still specify r.ContentLength, and I verified the lack of a guarantee from ReadAll that it wouldn't hoard memory. Further, io.LimitReader will return 0, EOF if it's limit <= 0. Removed that commit. :)

lpabon:
Looks good!

Issue #461, opened 2016-08-30
Need to check if we have unit tests to test cluster resizing


lpabon:
We need to check if we have unit tests which test increasing available Heketi storage by adding devices, nodes, and/or clusters.

lpabon:
Unit tests should test https://github.com/heketi/heketi/wiki/Managing-your-clusters

ramkrsna:
## Testing Specification for Increasing Available Nodes :: 
---------------------------------------------------------------------
### -  Adding a Devices
1. CLI - test argument, to check valid parameters 
2. API -> AddDevice -> RESTFUl API | POST
3. topology.json -> test rerun mock

### -  Adding Nodes or Increase Cluster Size 
-----------------------------------------------------
1. CLI - add node + register device to the added node 
2. API - /nodes POST
3. topology.json -> mock test to Add Node information after existing ones, test function in heketi to figure out the cluster on which node
### 
-  Adding Clusters i.e Adding an Entire New Cluster
-----------------------------------------------------------------  
1. CLI
2. API
3. topology.json file object

### - Reduce Storage Capacity
----------------------------------
1. CLI
2. API  

user roles 
-------------
role :: admin  :: Manage -> Single Gluster Cluster
role :: admin  :: Manage ->  Multiple Cluster

lpabon:
@ramkrsna For the unit tests, only concentrate on the API, do not use the cli or topology file.  @MohamedAshiqrh will be working on a ci system for the cli in the future.  This will make it easier.

For functional tests, you can (after this fix) look at `tests/functional/...` and create a new functional test which checks the heketi-cli and topology file.

ramkrsna:
https://github.com/ramkrsna/heketi/tree/issue461.  Its failing to trigger a test. 
 godep go test -v -run TestVolumeClusterResizeByAddingNodes
=== RUN   TestVolumeClusterResizeByAddingNodes
--- PASS: TestVolumeClusterResizeByAddingNodes (0.03s)
PASS
ok  	github.com/heketi/heketi/apps/glusterfs	0.047s

lpabon:
[...]

PR #513, opened 2016-09-21
volume: Test creation of volume by adding devices

lpabon:
This unit test checks to make sure that a volume can be created
on a full cluster by adding more devices.  The test first creates
a cluster, then creates a volume which consumes the entire cluster.
Then it tries to create another which it knows it cannot.  The test
will then add devices to the full cluster, and create the volume
again.

This test is one of three tests needed to satisfy #461.

### DIFF
--- a/apps/glusterfs/app_volume_test.go
+++ b/apps/glusterfs/app_volume_test.go
@@ -30,6 +30,7 @@ import (
 
 	"github.com/boltdb/bolt"
 	"github.com/gorilla/mux"
+	client "github.com/heketi/heketi/client/api/go-client"
 	"github.com/heketi/heketi/pkg/db"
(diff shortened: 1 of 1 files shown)


lpabon:
Ignore the centos-ci error.  openshift-ansible repo has removed the tag we were using.

```
TASK [client : get openshift-ansible] ******************************************
fatal: [client]: FAILED! => {"changed": false, "cmd": "/bin/git checkout --force enterprise-3.2", "failed": true, "msg": "Failed to checkout enterprise-3.2", "rc": 1, "stderr": "error: pathspec 'enterprise-3.2' did not match any file(s) known to git.\n", "stdout": "", "stdout_lines": []}
```

PR #517, opened 2016-09-22
app: Correctly handle stale registrations

### LINKED ISSUE #361: Conflict checking needs to make sure the element exists
Sometimes (maybe the server was killed or crashed), there could be a lingering Conflict key word in the DB referring to a non existing object, like in line: https://github.com/heketi/heketi/blob/master/apps/glusterfs/node_entry.go#L85 .

Instead it should check that the object reserving that name actually exists.

lpabon:
Heketi uses the concept of registrations to manage conflict
resolution.  For example, a device `/dev/sda` cannot be added
twice to the same node.  In the same manner a node with the
same hostnames cannot be added twice.

To resolve conflicts, Heketi _Registers_ the conflicting string
as a _key_ in the database.  If for any reason the device or
node was unable to be added, then Heketi has methods to remove
that registration.

The issue arises when Heketi is adding a device, then crashes
for any number of reasons.  At this point, the registered key
is still in the database for conflict resolution, but the entry
does not exist.

This patch checks to make sure that a registered name has an
entry in the database, and it is not a stale key describing
and unknown device or node.

Closes #361

### DIFF
--- a/apps/glusterfs/device_entry.go
+++ b/apps/glusterfs/device_entry.go
@@ -19,7 +19,6 @@ package glusterfs
 import (
 	"bytes"
 	"encoding/gob"
-	"errors"
 	"fmt"
 	"sort"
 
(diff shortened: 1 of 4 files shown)


lpabon:
@heketi/dev please review

MohamedAshiqrh:
@lpabon looks good to me. will it be good, if we have a generic id conflict function, which can be used by device, cluster and nodes? just a suggestion.

lpabon:
@MohamedAshiqrh Do you mean, could we create an Interface for Registration/Deregistration where the Node and devices (not clusters) could implement?  That could be done if the conflict was itself something accessed directly by callers.  But because you need to create the _entry_ then register it, it does not provide help today. 

Although, I do like the idea, and is something we should keep in mind as we go forward.

PR #526, opened 2016-09-27
use a different file as systemd environmentfile

### LINKED ISSUE #525: systemd unit file should not use json config as environmentfile
EnvironmentFile option in unit file is used to keep options/args to exec command modular.

Currently we have
`EnvironmentFile=-/etc/heketi/heketi.json`  
but the idea is to be able to have a file say heketi.env which has
`CONFIG_FILE=/etc/heketi/heketi-new.json`  
and then the ExecStart line in unit file should reference it like so
`ExecStart=/usr/bin/heketi -config=$CONFIG_FILE`  
Hence, it should not be a config file that heketi can use; it should be a config file that systemd can use.

raghavendra-talur:
Closes #525

### DIFF
--- a/extras/systemd/heketi.service
+++ b/extras/systemd/heketi.service
@@ -4,7 +4,7 @@ Description=Heketi Server
 [Service]
 Type=simple
 WorkingDirectory=/var/lib/heketi
-EnvironmentFile=-/etc/heketi/heketi.json
+EnvironmentFile=-/etc/heketi/heketi.env
 User=heketi
 ExecStart=/usr/bin/heketi -config=/etc/heketi/heketi.json
 Restart=on-failure


extras/systemd/heketi.service
```diff
@@ -4,7 +4,7 @@ Description=Heketi Server
 [Service]
 Type=simple
 WorkingDirectory=/var/lib/heketi
-EnvironmentFile=-/etc/heketi/heketi.json
+EnvironmentFile=-/etc/heketi/heketi.env
```
lpabon: Hi Talur.  There is no heketi.env installed.  Should we create one, just remove this line completely?

raghavendra-talur: It is optional, so user may choose to create it when they want to. Non-existence of the file is not a problem as we have '-' before the file path.

PR #530, opened 2016-09-29
api: New GID support in VolumeCreate API

### LINKED ISSUE #512: "permission denied" error for writes on app container
The issue is that volumes come with UID/GID of 0 (root).  For a container to be able to write to them, we need to have the container also have a root ID, but this, of course, is not a good solution.

Instead we have determined that Heketi can set a GID on a volume on creation and set the permissions to 775.  This would then allow the POD/container to have the GID set to that of the volume.  Allowing the container to write to the volume.

[BZ 1356437](https://bugzilla.redhat.com/show_bug.cgi?id=1356437)

lpabon:
When a volume is created, it is necessary for the provisioner to
request that the volume has a certain group id.  This group id
would allow the owner of the volume to be able to write to the
GlusterFS volume as long as the user belongs to the requested
group id.

The problem was seen during the access of GlusterFS volumes
from OpenShift containerized applications.  When the volume
was created, it would have a default GID of 0 (root). The
non-privileged containerized application would then be denied
permission to write to the volume due to the fact that it had
a GID other than 0 (root).

With this change, the provisioner can request that the volume
be initialized to a certain GID which matches that of the
containerized application.

Closes #512

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,6 +33,7 @@ LDFLAGS :=-ldflags "-X main.HEKETI_VERSION=$(VERSION)"
 # Package target
 PACKAGE :=$(DIR)/dist/$(APP_NAME)-$(VERSION).$(GOOS).$(ARCH).tar.gz
 CLIENT_PACKAGE :=$(DIR)/dist/$(APP_NAME)-client-$(VERSION).$(GOOS).$(ARCH).tar.gz
+GOFILES=$(shell go list ./... | grep -v vendor)
 
 .DEFAULT: all
 
@@ -60,7 +61,7 @@ run: server
 	./$(APP_NAME)
 
 test: 
-	godep go test ./...
+	godep go test $(GOFILES)
 
 clean:
 	@echo Cleaning Workspace...
--- a/apps/glusterfs/app_volume.go
+++ b/apps/glusterfs/app_volume.go
@@ -19,6 +19,7 @@ package glusterfs
 import (
 	"encoding/json"
 	"fmt"
+	"math"
 	"net/http"
 
 	"github.com/boltdb/bolt"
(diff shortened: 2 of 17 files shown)


lpabon:
@heketi/dev Please review

Makefile
```diff
@@ -60,7 +61,7 @@ run: server
 	./$(APP_NAME)
 
 test: 
-	godep go test ./...
+	godep go test $(GOFILES)
```
ramkrsna: godep go test $(GOFILES), would this run all the testfiles in all the vendor packages too ?

apps/glusterfs/app_volume.go
```diff
@@ -41,6 +42,16 @@ func (a *App) VolumeCreate(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// Check group id
+	switch {
+	case msg.Gid < 0:
+		http.Error(w, "Bad group id less than zero", http.StatusBadRequest)
+		return
+	case msg.Gid >= math.MaxInt32:
```
ramkrsna: Should it be math.MaxInt64 ?

Makefile
lpabon: Actually that was the problem before.  This only runs the test in Heketi.  Makes it simple to test all of your code before submitting.

apps/glusterfs/app_volume.go
lpabon: I thought that also, but then I noticed that Kubernetes has a test against math.MaxInt32:

https://github.com/kubernetes/kubernetes/blob/master/pkg/api/validation/validation_test.go#L3238

So I decided to follow their lead.

PR #532, opened 2016-09-29
Adding TestVolumeClusterResizeByAddingNodes

ramkrsna:
I create a cluster , add nodes to the cluster and then I add devices to the node existing cluster. I populate should consume all storage, then try to create a volume and expect failure. Then add more storage. Then create a volume and use all storage, expect success. Then create another volume and expect failure.

### DIFF
--- a/apps/glusterfs/app_volume_test.go
+++ b/apps/glusterfs/app_volume_test.go
@@ -1034,3 +1034,90 @@ func TestVolumeClusterResizeByAddingDevices(t *testing.T) {
 	err = v.Create(app.db, app.executor, app.allocator)
 	tests.Assert(t, err == ErrNoSpace)
 }
+
+func TestVolumeClusterResizeByAddingNodes(t *testing.T) {
+	tmpfile := tests.Tempfile()
+	defer os.Remove(tmpfile)
+
+	// Create the app
+	app := NewTestApp(tmpfile)
+	defer app.Close()
+	router := mux.NewRouter()
+	app.SetRoutes(router)
+
+	// Setup the server
+	ts := httptest.NewServer(router)
+	defer ts.Close()
+
+	// Create a cluster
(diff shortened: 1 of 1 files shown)


apps/glusterfs/app_volume_test.go
```diff
@@ -1034,3 +1034,90 @@ func TestVolumeClusterResizeByAddingDevices(t *testing.T) {
...
+	}
+
+	// Add devices to that perticular node
+
+	n := NewNodeEntryFromRequest(noderequest)
+	n.DeviceAdd("xyz")
+	tests.Assert(t, len(n.Devices) == 1)
+
+	// Add two devices to the cluster
+	for _, nodeId := range clusterInfo.Nodes {
```
lpabon: Here the tests add devices to all nodes.  What we need is for the test to add devices to the node in 1098.

apps/glusterfs/app_volume_test.go
```diff
@@ -1034,3 +1034,90 @@ func TestVolumeClusterResizeByAddingDevices(t *testing.T) {
...
+		Hostnames: api.HostAddresses{
+			Manage:  []string{"manage"},
+			Storage: []string{"storage"},
+		},
+		Zone: 1,
+	}
+
+	// Add devices to that perticular node
+
+	n := NewNodeEntryFromRequest(noderequest)
```
lpabon: It would be better if you use the client api to add a new node.  It would return node information.

lpabon:
Also, the commit message above uses the word "I" a lot.  I would suggest instead creating a commit message describing the patch itself.  You can go ahead and edit the message.

obnoxxx:
@ramkrsna , any update on this PR?

ramkrsna:
I'll work on a newer patchset for this PR. Will update it the #461 . Having issues with the client calls, closing this patchset .

PR #537, opened 2016-10-05
api: Do not add gid information on volume info

lpabon:
GID cannot be reliably returned since the user could have changed the value
to something else.  Instead, omit the gid value in the JSON message
returned to a volume information request.


obnoxxx:
:+1:

PR #548, opened 2016-10-18
api: Better invalid path error handler

### LINKED ISSUE #539: Get a 404 error from heketi-cli when requesting a non-hex value id
```
Same applicable for cluster/device/node

#heketi-cli cluster info t
Error: 404 page not found

# heketi-cli device info t
Error: 404 page not found

#heketi-cli node info t
Error: 404 page not found
```

[BZ 1351646](https://bugzilla.redhat.com/show_bug.cgi?id=1351646)

lpabon:
When a path with an illegal id is provided by the caller (something
that does not match the regexp), the Gorilla Mux Router automatically
routed the call to http.NotFoundHandler().  This would be ok for
an API to handle, but the heketi-cli would output information which
was not very human friendly.

Instead, we now provide our own error handler which returns a nicer
error than just "404: Not found".

Closes #539

### DIFF
--- a/apps/glusterfs/app.go
+++ b/apps/glusterfs/app.go
@@ -346,8 +346,10 @@ func (a *App) SetRoutes(router *mux.Router) error {
 
 	}
 
-	return nil
+	// Set default error handler
+	router.NotFoundHandler = http.HandlerFunc(a.NotFoundHandler)
 
+	return nil
 }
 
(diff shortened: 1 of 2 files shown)


lpabon:
@heketi/dev Please review when you have time.

obnoxxx:
Patch looks good to me, :+1:

PR #556, opened 2016-10-27
build: Support full RELRO

lpabon:
Currently, the go build tool passes -z relro to the linker, which
creates a 'partial' RELRO executable.  Maybe it is a bug in the
GNU linker, but '-z relro' only creates a partial RELRO.  For
a full RELRO to be created the switch '-z now' must also be
passed.

See http://tk-blog.blogspot.com/2009/02/relro-not-so-well-known-memory.html
for more information.


lpabon:
@heketi/dev

PR #564, opened 2016-11-03
Kubernetes Deployment: Changes based on Kubernetes Integration

MohamedAshiqrh:
Master branch should make use of heketi/heketi:dev image.

### DIFF
--- a/extras/kubernetes/deploy-heketi-deployment.json
+++ b/extras/kubernetes/deploy-heketi-deployment.json
@@ -50,7 +50,7 @@
             "spec": {
                 "containers": [
                     {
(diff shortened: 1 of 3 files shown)


extras/kubernetes/deploy-heketi-deployment.json
```diff
@@ -50,7 +50,7 @@
             "spec": {
                 "containers": [
                     {
-                        "image": "heketi/heketi",
+                        "image": "heketi/heketi:dev",
```
lpabon: Removing the tags makes it so that they always get the latest stable container.

PR #574, opened 2016-11-16
remove outdated useless python example.

### LINKED ISSUE #541: Remove old useless python example
Before the python client was created, we created a simple poc example which is no longer usable and is confusing.  This needs to be removed

obnoxxx:
Closes #541


obnoxxx:
Merging since I heard no veto..

Issue #382, opened 2016-06-15
topology info: "ID not found" error when running concurrently with volume creation.


NehaRawat:
heketi-cli topology info
Error: Id not found

Logs:
[negroni] Started GET /volumes/1819834f5be2c3dce81b36e14ea440c0
[negroni] Completed 200 OK in 761.725µs
[negroni] Started GET /nodes/410b2c68b6df188f88a69003cb8a757b
[negroni] Completed 500 Internal Server Error in 6.36837ms

500 when fetching node information. As of now not sure why it's failing when running parallely.

lpabon:
If at all possible.  Save the heketi.db and attach it to this bug.  If it is in a container, you may need to attach a new host volume, then rsh into the container and copy the /var/lib/heketi/heketi.db to that host volume.

lpabon:
@NehaRawat Sounds like the test was concurrently creating and deleting volumes while you ran `heketi-cli topology info`.  Is that correct?

NehaRawat:
@lpabon I hit this issue during volume creation. Will not hit this issue if we just try to create a single volume as it returns quickly. 

Easily reproducible If we try to create volume in loop say 5-10 volumes and then run "topology info" in another terminal.

lpabon:
@NehaRawat ok, but in the loop, did you delete the volumes or just create.

NehaRawat:
@lpabon just create.

lpabon:
Great, thanks @NehaRawat

lpabon:
[BZ 1346621](https://bugzilla.redhat.com/show_bug.cgi?id=1346621)

MohamedAshiqrh:
For some reason 

DB is not returning any value, only in case of nodeId while volume create is in progress.
This is where we are hitting but not sure why we are hitting when volume create.
https://github.com/heketi/heketi/blob/master/apps/glusterfs/dbentry.go#L137

Now looking into volume create to find more...Will keep updating.

MohamedAshiqrh:
Node Info and Device Info are returning "Id not found" for valid id's, when volume create is in progress. Have created an easier setup to try out heketi code changes as pod on openshift. Not able to find anything suspicious yet.

MohamedAshiqrh:
https://github.com/MohamedAshiqrh/heketi/tree/ninfo

This is where all the test code progresses. If anyone is interested.

MohamedAshiqrh:
https://github.com/heketi/heketi/pull/575

PR #575, opened 2016-11-16
Topology Info, Node Info: "ID not found" error when running concurrently with volume creation.

MohamedAshiqrh:
The brick entries which is being updated during `volume create` causes inconsistent state ( until the transaction is over ). In this state, It returns `Err:Id not found` for bricks when issuing  `node info` | `topology info`. This fix will make `node info` and `topology info` not to depend on the bricks information. In actual scenario `node info` and `topology info` is not using the bricks anyways.

Closes:#382

### DIFF
--- a/apps/glusterfs/device_entry.go
+++ b/apps/glusterfs/device_entry.go
@@ -282,6 +282,20 @@ func (d *DeviceEntry) NewInfoResponse(tx *bolt.Tx) (*api.DeviceInfoResponse, err
 	return info, nil
 }
 
+func (d *DeviceEntry) NewInfoResponseNode(tx *bolt.Tx) (*api.DeviceInfoResponse, error) {
+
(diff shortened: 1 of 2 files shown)


MohamedAshiqrh:
review @heketi/dev

humblec:
@MohamedAshiqrh Nice catch !!! :+1:  one small suggestion to source document the newNewInfoResponseNode() func.

obnoxxx:
@MohamedAshiqrh as discussed, this patch looks clean, but it papers over the actual problem. It also removes the possibility for a future user of nodeinfo to get to the brick part. So let's try whether pr #579 helps first. We can still chose this as a plan b! :-)

obnoxxx:
This should not be necessary any more with #579 merged. Please re-propose if issues remain, @MohamedAshiqrh

PR #579, opened 2016-11-17
app: save bricks in allocBricks() to keep the db consistent

### LINKED ISSUE #382: topology info: "ID not found" error when running concurrently with volume creation.

obnoxxx:
Earlier, the bricks were only saved to the DB in volume.Create.

Closes #382

Pair-Programmed-With: Mohamed Ashiq Liyazudeen <mliyazud@redhat.com>

### DIFF
--- a/apps/glusterfs/app_volume_test.go
+++ b/apps/glusterfs/app_volume_test.go
@@ -11,6 +11,7 @@ package glusterfs
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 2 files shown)


obnoxxx:
There are issues. Gonna fix up an resubmit...

obnoxxx:
Okay... that's strange: The travis tests fail at apps/glusterfs/volume_entry_test.go:607. But on my machine it succeeds... :-o

obnoxxx:
oops, i was testing the wrong branch. the error is real.

lpabon:
And that is why it fails at https://github.com/heketi/heketi/blob/master/apps/glusterfs/volume_entry_test.go#L607 . Because the brickpath has been saved to the database before the path has been set during brick creation.

lpabon:
There also should be a new test added which replicates the issue #382 and shows how this patch fixes it.

obnoxxx:
updated the patch so that it should pass tests now.
will work on providing a unit test.... need to figure out how to best to it

obnoxxx:
Updated the PR with an additional patch that adds a test for issue #382. When the fix patch is reverted, the test fails.

obnoxxx:
@raghavendra-talur In the actual fix, we needed to keep the saving of bricks in VolumeEntry.Create(), because the CreateBricks() at line 333 actually fills the brick with info. So we need to store twice: once in allocBricks so that the device and volume structures are not pointing into nirvana, and once after CreateBricks() to store the actual brick data.

obnoxxx:
Updated the PR with fixed formatting in the test (go fmt ...)...
(Warning @lpabon : Did an amended patch for this... ;-) )

PR #589, opened 2016-11-29
tests: Update minikube to 0.12.2

lpabon:
Also, pend until all the kubernetes containers are online.  This
was probably the reason why most tests sometimes worked and sometimes
did not.

Signed-off-by: Luis Pabón <luis.pabon@coreos.com>

PR #590, opened 2016-11-29
Kube smoke test fixes

obnoxxx:
Combined work of Luis and myself.

### DIFF
--- a/tests/functional/TestKubeSmokeTest/md5sums
+++ b/tests/functional/TestKubeSmokeTest/md5sums
@@ -1,0 +1,4 @@
+8f5de36e01e8f160e91681cbc8355a64  /usr/local/bin/docker-machine
+519d29300df740a73c90d048e5df43fb  /usr/local/bin/docker-machine-driver-kvm
+eab2ea4df73dd3cbbe147954a161d10f  /usr/local/bin/kubectl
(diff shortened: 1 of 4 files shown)


obnoxxx:
Replaces PR #589 .

obnoxxx:
ok... it still fails in TesKubeSmokeTest:
'''
Step 6 : ADD ./heketi.json /etc/heketi/heketi.json
 ---> bea4f63dfab3
Removing intermediate container aa7b8d21df51
Step 7 : RUN mkdir /var/lib/heketi
 ---> Running in eab77b3fcb1f
permission denied
Container command could not be invoked.
==> ERROR: Unable to create docker container
==> FAILED TestKubeSmokeTest
==> TEARDOWN TestKubeSmokeTest
'''
This works locally for me.

lpabon:
This works locally for me also.  I think we need to do a chmod on the image that is created so that we can copy it.

obnoxxx:
@lpabon This is a different error. It happens inside the ```docker build``` command. I think what happens is that the ```mkdir /var/lib/heketi``` gets permission denied.

obnoxxx:
I think we are getting further than before. This error shows that the actual setup of minikube has succeeded. I'm going to push these changes. and work on a new patch for this new issue.

PR #592, opened 2016-11-30
tests: Skip OpenShift functional tests

lpabon:
When Heketi first started being containerized, the only tests
we had was to deploy in OpenShift.  Now with minikube, we can
do our tests with Kubernetes.  These tests were also unreliable
to deploy, and by skipping them they will provide a more stable
CI test suite.


obnoxxx:
@centos-ci ok to test

obnoxxx:
The patch works in that it skips the openshift functional tests.
I am thinking about doing it differently: I don't like renaming the directories very much. It creates a lot of churn in patches enabling and disabling.

PR #597, opened 2016-11-30
tests: Remove OpenShift Tests

lpabon:
Now that we use minikube and the latest Kubernetes, there is no
need to continue using the OpenShift tests.


obnoxxx:
@centos-ci ok to test

PR #601, opened 2016-12-05
Add daemonset envvar

ant31:
As the proposed way is now to use ds, this variable is missing

### DIFF
--- a/extras/kubernetes/heketi-deployment.json
+++ b/extras/kubernetes/heketi-deployment.json
@@ -82,6 +82,10 @@
                             "value": "y"
                           },
                           {
+                            "name": "HEKETI_KUBE_GLUSTER_DAEMONSET",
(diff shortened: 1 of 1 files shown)


obnoxxx:
Looks good.
We are in the process of moving this content to https://github.com/gluster/gluster-kubernetes.git (that code already has it), but happy to fix it here now.

obnoxxx:
@centos-ci ok to test

ant31:
@obnoxxx 
is it possible to remove any reference to 'kubernetes' example other than  https://github.com/gluster/gluster-kubernetes.git ? 

It was very confusing as I get redirected to 3-4 different version of kubernetes resources while I was following documentation (one in extra/ one in the client targz, one in gluster-kubernetes, and one external repo), and each have some difference.

PR #612, opened 2016-12-07
executors: chmod with 2775 to inherit group permission

### LINKED ISSUE #611: chmod cmd should also set S bit to enable a particular group to write in subdirs
With https://github.com/heketi/heketi/commit/b56a5e86f6c6b7f7596ae88e82652c067bc9107b we added a feature where one could add a gid to the volume to enable all users belonging to that group to be able to write into the volume.

However, it does so by changing the mode bits on the brick dirs to 775. This is not sufficient for the cause. Here is a test case which fails:

1. create a group writegroup 
2. add rtalur and vagrant to it
3. create a volume by providing writegroup's gid to heketi
4. as rtalur create files and dirs in the volume, it works
5. as vagrant creates files and dirs in the root of the volume, it works
6. as vagrant creates files and dirs in the dirs created by rtalur in step 4, it fails with permission denied although it should succeed.

raghavendra-talur:
With b56a5e8 we added a feature where one could add a gid to the volume
to enable all users belonging to that group to be able to write into the
volume.

However, it does so by changing the mode bits on the brick dirs to 775.
This is not sufficient for the cause. Here is a test case which fails:

* create a group writegroup
* add rtalur and vagrant to it
* create a volume by providing writegroup's gid to heketi
* as rtalur, create files and dirs in the volume, it works
* as vagrant, creates files and dirs in the root of the volume, it works
* as vagrant, creates files and dirs in the dirs created by rtalur in step
4, it fails with permission denied although it should succeed.

This patch fixes the above test.

Closes #611

### DIFF
--- a/tests/functional/TestSmokeTest/tests/heketi_test.go
+++ b/tests/functional/TestSmokeTest/tests/heketi_test.go
@@ -264,7 +264,7 @@ func TestHeketiCreateVolumeWithGid(t *testing.T) {
 	volInfo, err := heketi.VolumeCreate(volReq)
 	tests.Assert(t, err == nil)
 
-	// SSH into system and execute gluster command to create a snapshot
+	// SSH into system and try creating files from user belonging to gid
 	exec := ssh.NewSshExecWithKeyFile(logger, "vagrant", "../config/insecure_private_key")
 	cmd := []string{
 		fmt.Sprintf("sudo mount -t glusterfs %v /mnt", volInfo.Mount.GlusterFS.MountPoint),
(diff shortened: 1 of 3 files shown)


obnoxxx:
@centos-ci  ok to test

tests/functional/TestSmokeTest/tests/heketi_test.go
```diff
@@ -264,7 +264,7 @@ func TestHeketiCreateVolumeWithGid(t *testing.T) {
 	volInfo, err := heketi.VolumeCreate(volReq)
 	tests.Assert(t, err == nil)
 
-	// SSH into system and execute gluster command to create a snapshot
+	// SSH into system and try creating files from user belonging to gid
```
obnoxxx: why the comment change here?
it breaks rebase to pr #614 and I think is too early in this commit?

obnoxxx:
I put this into #617 , combined with #614. tested locally.

obnoxxx:
pushed in #617

PR #614, opened 2016-12-07
tests: add a test for checking if volume is group writable

raghavendra-talur:

### DIFF
--- a/pkg/utils/ssh/ssh.go
+++ b/pkg/utils/ssh/ssh.go
@@ -98,6 +98,21 @@ func NewSshExecWithKeyFile(logger *utils.Logger, user string, file string) *SshE
 	return sshexec
 }
 
+// This function requires the password string to be crypt encrypted
+func NewSshExecWithPassword(logger *utils.Logger, user string, password string) *SshExec {
+
+	sshexec := &SshExec{}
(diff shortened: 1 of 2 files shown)


pkg/utils/ssh/ssh.go
```diff
@@ -98,6 +98,21 @@ func NewSshExecWithKeyFile(logger *utils.Logger, user string, file string) *SshE
 	return sshexec
 }
 
+// This function requires the password string to be crypt encrypted
+func NewSshExecWithPassword(logger *utils.Logger, user string, password string) *SshExec {
```
obnoxxx: this deserves a separate commit (in same PR).

note that I prefer amended force-pushes over add-on patches

obnoxxx:
@centos-ci ok to test

obnoxxx:
this test with the gid patch succeeded for me.
now testing w/o gid patch

obnoxxx:
without the gid patch, it fails, as expected

@raghavendra-talur  can you do the changes I suggested and file a PR with both paches (from pr #612 and this one)?

obnoxxx:
I put this into #617 , combined with #612 . tested locally.

obnoxxx:
pushed in #617

Issue #552, opened 2016-10-25
[Integration ] Logging in Heketi 


ramkrsna:
**Problem** 
How to use openshift more broadly for CNS with respect to logging. 
- [ ] - Logging integrations between heketi and openshift, 
- [ ] - How to view gluster logs in openshift ELK stack 

**Actions**
Fluentd needs to be able to aggregate both gluster and heketi logs. 
- Gluster container Logs 
- Heketi containers Logs

ramkrsna:
- correction, there seems to be a modification instead of ELK stack to EFK stack .

ramkrsna:
- [1351535](https://bugzilla.redhat.com/show_bug.cgi?id=1351535) Need error logging on volume create failures
- [1363602](https://bugzilla.redhat.com/show_bug.cgi?id=1363602) Heketi server logs needs to be persisted

PR #613, opened 2016-12-07
Heketi logs to both stderr and heketi.log

ramkrsna:
Initial commit to see to log to a file along with the stderr works #552

PR #619, opened 2016-12-08
Heketi-Templates:Persisting logfile in the heketidbstorage volume

MohamedAshiqrh:
Requires: #613 
Closes: #552 

### DIFF
--- a/extras/openshift/templates/heketi-template.json
+++ b/extras/openshift/templates/heketi-template.json
@@ -150,6 +150,10 @@
                   {
                     "name": "db",
                     "mountPath": "/var/lib/heketi"
+                  },
+                  {
+                    "name": "db",
+                    "mountPath": "/var/log/heketi"
                   }
                 ],
                 "readinessProbe": {


MohamedAshiqrh:
@obnoxxx This has to be tested after #613 is merged.

extras/openshift/templates/heketi-template.json
```diff
@@ -150,6 +150,10 @@
                   {
                     "name": "db",
                     "mountPath": "/var/lib/heketi"
+                  },
+                  {
+                    "name": "db",
```
lpabon: Same name?  I'm not sure that works.

obnoxxx: Right ... should it be s/th like "log" ?

lpabon:
I think this change is not necessary if the log file is placed in `/var/lib/heketi` since that is the GlusterFS volume which is already mounted.  This is being fixed in #620

extras/openshift/templates/heketi-template.json
MohamedAshiqrh: name should match name in ["volumes"](https://github.com/heketi/heketi/pull/619/files#diff-db4f124381d5c61236daeb291895e5f6R177). We can have a separate volumes "entry" for the same gluster(or heketidbstorage) volume, which internally will mount the same gluster(or heketidbstorage) volume on the host and the bind mount it for the different volume entry(which I feel unnecessary). I will test this out right away. I feel heketi container should be as light as possible.

MohamedAshiqrh:
@lpabon I agree for now, but for long run it is better to have it in a place where it is expected. also for standalone heketi, it will be odd to have logs in `/var/lib/heketi/`.

lpabon:
Just like #620, this change is *not* fixing the issue the correct way.   Please see #620 for more information.

obnoxxx:
@MohamedAshiqrh is this something we should close?

MohamedAshiqrh:
Not required any more as heketi is doing things as expected.

PR #620, opened 2016-12-08
Enable Logging to a File

ramkrsna:
Logging to a specified heketi.log file .  #552 #619 
[1363602](https://bugzilla.redhat.com/show_bug.cgi?id=1363602) Heketi server logs needs to be persisted 

Signed-off-by: Ramakrishna Reddy Yekulla <rreddy@redhat.com>

### DIFF
--- a/pkg/utils/log.go
+++ b/pkg/utils/log.go
@@ -74,6 +74,8 @@ func NewLogger(prefix string, level LogLevel) *Logger {
 		l.level = level
 	}
 
+	stdout, _ = os.OpenFile("/var/log/heketi/heketi.log", os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0666)
+	//stdout = os.Stdout
 	l.critlog = log.New(stderr, prefix+" CRITICAL ", log.LstdFlags)
 	l.errorlog = log.New(stderr, prefix+" ERROR ", log.LstdFlags)
 	l.warninglog = log.New(stdout, prefix+" WARNING ", log.LstdFlags)
--- a/pkg/utils/log.go
+++ b/pkg/utils/log.go
@@ -74,14 +74,19 @@ func NewLogger(prefix string, level LogLevel) *Logger {
 		l.level = level
 	}
 
-	stdout, _ = os.OpenFile("/var/log/heketi/heketi.log", os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0666)
-	//stdout = os.Stdout}
+	fileout, err := os.OpenFile("/var/log/heketi/heketi.log", os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0666)
+	if err != nil {
+		log.Println("err")
+	}
 
-	l.critlog = log.New(stderr, prefix+" CRITICAL ", log.LstdFlags)
-	l.errorlog = log.New(stderr, prefix+" ERROR ", log.LstdFlags)
-	l.warninglog = log.New(stdout, prefix+" WARNING ", log.LstdFlags)
-	l.infolog = log.New(stdout, prefix+" INFO ", log.LstdFlags)
-	l.debuglog = log.New(stdout, prefix+" DEBUG ", log.LstdFlags)
+	mstdout := io.MultiWriter(fileout, os.Stdout)
+	mstderr := io.MultiWriter(fileout, os.Stderr)
+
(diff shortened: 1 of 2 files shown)


ramkrsna:
@heketi/dev Could you verify this patch .

obnoxxx:
@ramkrsna The unit tests fail at logging. Have you run them locally?

pkg/utils/log.go
```diff
@@ -74,6 +74,8 @@ func NewLogger(prefix string, level LogLevel) *Logger {
 		l.level = level
 	}
 
+	stdout, _ = os.OpenFile("/var/log/heketi/heketi.log", os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0666)
```
obnoxxx: @ramkrsna Logging to a file from a long-running service is not as easy as simply opening it for creation or appending and then writing to it: this way the log file will never get rotated and grow indefinitely. One usually needs some kind of signal handler to reopen the file (e.g on USR2). This way a logrotate system can work.

Logging to stderr/stdout or syslog circumvents these complications.

pkg/utils/log.go
```diff
@@ -75,7 +75,8 @@ func NewLogger(prefix string, level LogLevel) *Logger {
 	}
 
 	stdout, _ = os.OpenFile("/var/log/heketi/heketi.log", os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0666)
-	//stdout = os.Stdout
```
obnoxxx: Furthermore:

- This should not change the current behavior, but should be configurable.
- Errors from OpenFile need to be checked.

pkg/utils/log.go
```diff
@@ -74,11 +74,19 @@ func NewLogger(prefix string, level LogLevel) *Logger {
 		l.level = level
 	}
 
-	l.critlog = log.New(stderr, prefix+" CRITICAL ", log.LstdFlags)
-	l.errorlog = log.New(stderr, prefix+" ERROR ", log.LstdFlags)
-	l.warninglog = log.New(stdout, prefix+" WARNING ", log.LstdFlags)
-	l.infolog = log.New(stdout, prefix+" INFO ", log.LstdFlags)
-	l.debuglog = log.New(stdout, prefix+" DEBUG ", log.LstdFlags)
+	fileout, err := os.OpenFile("/var/log/heketi/heketi.log", os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0666)
```
lpabon: You do not want this library to assume where to write.  This should come from the caller.

obnoxxx: Indeed.

lpabon:
Could this be fixed instead in the Dockerfile? Since the log is just output to stdout, could we use `tee` and just write it to `/var/lib/heketi`?

obnoxxx:
@lpabon I was wondering and asking about the same: Can't we intercept the stdout/stderr from the docker end and put them wherever is convenient for us?

ramkrsna:
An application like Heketi should not be coupled to logging to a file. Each working process in the CNS setup along with router/middleware (negroni) should write the output stream unbuffered to STDOUT. 

It is the execution environments ( OpenShift / Kubernetes) responsibility to capture the STDOUT for viewing and archival.

ramkrsna:
https://docs.docker.com/engine/admin/logging/overview/ , All the application have to do is to write to stdout / stderr. By default Red Hat / Fedora docker images seems to have journald configured.

lpabon:
@ramkrsna I agree.  I do not think this change is needed. Heketi is already doing the right right.  I have thought about this more, and I am now sure that Heketi should _not_ save a file on the node it is running on, because it is not the kubernetes way.

Instead there should be a logstash, fluentd, or other centralized logging service to archive logs.

obnoxxx:
Heketi logs to stderr/stdout, and from our deployment templates,
the docker containers are not spawned with any special options
for the log driver (or so), so that the host can capture and
persist the logs with the default configured methods.

I.e. it is up to the host admin (docker/kubernetes/openshift)
to create an environment where the logs are persisted.
(e.g. using json-file logging in combination with an 
appropriately configured fluentd, or using journald, etc...)

If the admin wants to override the systems default settings,
we will provide instructions on how to change the templates
prior to deploying CNS to achieve this.

Hence there is nothing required to be changed in heketi code.
Heketi behaves the default way and the rest is an admin task.

==> closing

PR #621, opened 2016-12-08
Automatic Kubernetes configuration in Heketi

### LINKED ISSUE #615: Use serviceAccountName to mount the secret volume
Hi, 

In kubernetes the certificates to access the api-master are automatically mounted on:
```
/var/run/secrets/kubernetes.io/serviceaccount 
```

and to select which serviceaccount certificate to mount, the key `pod.spec.serviceAccountName` can be used: 
```
---
kind: Deployment
apiVersion: extensions/v1beta1
metadata:
  name: heketi
  labels:
    glusterfs: heketi-deployment
  annotations:
    description: Defines how to deploy Heketi
spec:
  replicas: 1
  template:
    metadata:
      name: heketi
      labels:
        name: heketi
        glusterfs: heketi-pod
    spec:
      serviceAccountName: heketi-service-account
      containers:
      - image: heketi/heketi:dev
        imagePullPolicy: Always
        name: heketi
```
    
This will remove the need to manually enter secret name. 

The change would be to look in `/var/run/secrets/kubernetes.io/serviceaccount` to get the credentials.

lpabon:
With Kubernetes 1.5 we can now use higher level APIs to be able to automatically
setup REST RPC configuration with Kubernetes. There is no longer
a need to ask Namespace, API host, or Secret from the administrator with this
change.  The configuration also sets up an HTTPS encrypted connection to the
API host, which previous versions did not.

I tried first to use go-client for Kubernetes, but it does not provide
ability to setup and execute commands on containers.  Instead, this changes
enables Heketi to use Kubernetes 1.5 APIs to automatically setup configuration
with the API host.

Godeps had to be updated from Kubernetes 1.2.alpha to 1.5.beta.2.

This change also makes it necessary to deploy Heketi *in* Kubernetes
when using kubexec.

Closes #615


client/api/go-client/client_test.go
```diff
@@ -19,13 +19,13 @@ import (
...
 	"testing"
 
-	"github.com/codegangsta/negroni"
 	"github.com/gorilla/mux"
 	"github.com/heketi/heketi/apps/glusterfs"
 	"github.com/heketi/heketi/middleware"
 	"github.com/heketi/heketi/pkg/glusterfs/api"
 	"github.com/heketi/heketi/pkg/utils"
 	"github.com/heketi/tests"
+	"github.com/urfave/negroni"
```
obnoxxx: I think this and all the vendor/ updates should be in separate patches. (not necessarily separate PR, but separate patch for sure.) We should try to avoid mixing our own code changes with vendor/ updates)

lpabon: I'll see what I can do, but that is not easy..  Godeps is a nightmare.

lpabon:
Closed because #622 splits the commits to make it easier to review and maintain

PR #622, opened 2016-12-09
Automatic Kubernetes configuration in Heketi

### LINKED ISSUE #615: Use serviceAccountName to mount the secret volume

lpabon:
With Kubernetes 1.5 we can now use higher level APIs to be able to automatically
setup REST RPC configuration with Kubernetes. There is no longer
a need to ask Namespace, API host, or Secret from the administrator with this
change.  The configuration also sets up an HTTPS encrypted connection to the
API host, which previous versions did not.

I tried first to use go-client for Kubernetes, but it does not provide
ability to setup and execute commands on containers.  Instead, this changes
enables Heketi to use Kubernetes 1.5 APIs to automatically setup configuration
with the API host.

Godeps had to be updated from Kubernetes 1.2.alpha to 1.5.beta.2.

This change also makes it necessary to deploy Heketi *in* Kubernetes
when using kubexec.

Closes #615


lpabon:
retest this please @centos-ci

lpabon:
retest this please @centos-ci

lpabon:
@obnoxxx Although the Ci system is not very stable (the VMs, not the tests themselves), when it runs, the kubernetes tests pass.

lpabon:
retest this please @centos-ci

ant31:
To deploy in one step heketi/glusterfs on kubernetes we can do the following steps: 

Deploy kubernetes resources:
   -> topology-configmap.yaml
   -> [heketi-pvc.yaml](https://github.com/kubespray/heketi-config/blob/master/templates/heketi-pvc.yaml)
   -> heketi-service-account.yaml
   -> heketi-svc.yaml
   -> heketi-glusterfs-daemonset.yaml
   -> heketi-deployment.yaml
   -> heketi-init-job.yaml

The only file the user must edit is the `topology-configmap.yaml`, other resources can have working defaults. 

In details: 
1. The heketi-deployment.yaml, is the heketi-server: 
    - It mounts the heketi-pvc, because the pvc is not bound, the pod can't be scheduled it waits.
    - It has an [init-container](https://github.com/kubernetes/kubernetes/blob/master/docs/proposals/container-init.md) that mounts also heketi-pvc, and wait for a heketi.db file exists to exist

2. The heketi-init-job.yaml is a job that does: 
    1. mount the topology configmap
    2. start a heketi-server with emptyDir to store the db
    3. run heketi-cli --load topology:  
       the job can have 2 container in the pod: 1 server | 1 client 
    4. Create a the `heketidbstorage` glusterfs-volume, the glusterfs-endpoints
    5. Create a secret from the heketi.db
    6. Create the persistentvolume [heketi-pv.yaml](https://github.com/kubespray/heketi-config/blob/master/templates/heketi-pv.yaml) resource.  

After this job, the heketi-db-secret is created and the pv/pvc are created too, so the heketi-server pod can start. The init container mount the secret and copy it if does not exist in /var/lib/heketi.

jarrpa:
@lpabon I think the Service Account should still be in its own YAML file, since it is used by both deploy-heketi and heketi pods. It shouldn't be too much to ask a user to run an additional `kubectl create -f`, as we were already.

@ant31 Did you mean for your comment to land in this PR, or in a separate issue? It doesn't seem like you're referring to anything in this PR.

jarrpa:
After a chat with @lpabon, we determined that we don't need to maintain the split Service Account. If and when we move to remove the requirement for deploy-heketi from the deployment process, we can move the Service Account definition to the heketi JSON.

screeley44:
@ant31 - wrt your comments above:

`To deploy in one step heketi/glusterfs on kubernetes we can do the following steps:`

are we at a point with this that I can give it a try...is there an example or fork that has these configs?

ant31:
@jarrpa  I'll create a separate issue for my comment, I just jumped in this thread from its title.

PR #623, opened 2016-12-12
fix: Update CentOS software and certificates

lpabon:
This may fix the issues where it sometimes yum does not errors while
installing software packages.


lpabon:
Did not fix it reliably

Issue #637, opened 2016-12-21
Heketi should be able to create 1GB volume by default


MohamedAshiqrh:
Heketi Brick minimum size is configurable in advanced options. Making heketi to create 1GB volume as default.

obnoxxx:
closed by #642

PR #638, opened 2016-12-21
Change BrickMinSize to 1GB enabling heketi to create 1GB volumes

MohamedAshiqrh:
Closes: #637 

### DIFF
--- a/apps/glusterfs/app_volume_test.go
+++ b/apps/glusterfs/app_volume_test.go
@@ -166,11 +166,19 @@ func TestVolumeCreateInvalidSize(t *testing.T) {
 }
 
 func TestVolumeCreateSmallSize(t *testing.T) {
-	tmpfile := tests.Tempfile()
-	defer os.Remove(tmpfile)
 
-	// Create the app
-	app := NewTestApp(tmpfile)
+	dbfile := tests.Tempfile()
+	defer os.Remove(dbfile)
+	os.Setenv("HEKETI_EXECUTOR", "mock")
+	defer os.Unsetenv("HEKETI_EXECUTOR")
+
+	data := []byte(`{
+		"glusterfs" : {
+			"db" : "` + dbfile + `",
+			"brick_min_size_gb" : 4
+		}
+	}`)
+	app := NewApp(bytes.NewReader(data))
 	defer app.Close()
 	router := mux.NewRouter()
 	app.SetRoutes(router)
--- a/apps/glusterfs/app_test.go
+++ b/apps/glusterfs/app_test.go
@@ -82,7 +82,7 @@ func TestAppAdvsettings(t *testing.T) {
 			"allocator" : "simple",
 			"db" : "` + dbfile + `",
 			"brick_max_size_gb" : 1024,
-			"brick_min_size_gb" : 1,
+			"brick_min_size_gb" : 4,
 			"max_bricks_per_volume" : 33
 		}
 	}`)
@@ -93,11 +93,12 @@ func TestAppAdvsettings(t *testing.T) {
 	}()
 
 	app := NewApp(bytes.NewReader(data))
+	defer app.Close()
 	tests.Assert(t, app != nil)
 	tests.Assert(t, app.conf.Executor == "mock")
 	tests.Assert(t, BrickMaxNum == 33)
 	tests.Assert(t, BrickMaxSize == 1*TB)
-	tests.Assert(t, BrickMinSize == 1*GB)
+	tests.Assert(t, BrickMinSize == 4*GB)
 }
 
 func TestAppLogLevel(t *testing.T) {
--- a/apps/glusterfs/limits.go
+++ b/apps/glusterfs/limits.go
@@ -11,7 +11,7 @@ package glusterfs
 
 var (
 	// Default limits
-	BrickMinSize = uint64(4 * GB)
+	BrickMinSize = uint64(1 * GB)
 	BrickMaxSize = uint64(4 * TB)
 	BrickMaxNum  = 32
 )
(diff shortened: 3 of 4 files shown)


MohamedAshiqrh:
@obnoxxx I am not able to create a request for volume less than 1GB. As the size parameter is interger. Do you think it is a good idea to make size a float? I think we dont have a choice just wanted to know is there a way around.

apps/glusterfs/app_volume_test.go
```diff
@@ -166,12 +166,20 @@ func TestVolumeCreateInvalidSize(t *testing.T) {
...
-	defer app.Close()
+	dbfile := tests.Tempfile()
+	defer os.Remove(dbfile)
+	os.Setenv("HEKETI_EXECUTOR", "mock")
+	defer os.Unsetenv("HEKETI_EXECUTOR")
+
+	data := []byte(`{
+		"glusterfs" : {
+			"db" : "` + dbfile + `",
+			"brick_min_size_gb" : 2
```
obnoxxx: Suggestion: you can use 4 here (instead of 2). Then you don't need to change the test below, hence reducing the size of the patch! :-)

apps/glusterfs/volume_durability_test.go
```diff
@@ -77,41 +77,41 @@ func TestNoneDurability(t *testing.T) {
 	r := &NoneDurability{}
 	r.SetDurability()
 
-	gen := r.BrickSizeGenerator(100 * GB)
+	gen := r.BrickSizeGenerator(50 * GB)
```
obnoxxx: Instead of reducing the initial size, from 100 to 50 here, why not just ad another block below? reduced the diff and makes the change more obvious?

apps/glusterfs/volume_durability_test.go
```diff
@@ -77,41 +77,41 @@ func TestNoneDurability(t *testing.T) {
...
 
 	// Gen 5
 	sets, brick_size, err = gen()
 	tests.Assert(t, err == nil)
 	tests.Assert(t, sets == 16)
-	tests.Assert(t, brick_size == 6400*1024)
+	tests.Assert(t, brick_size == 3200*1024)
 	tests.Assert(t, 1 == r.BricksInSet())
 
 	// Gen 6
```
obnoxxx: would just add this here:
```
 	sets, brick_size, err = gen()
 	tests.Assert(t, err == nil)
 	tests.Assert(t, sets == 16)
	tests.Assert(t, brick_size == 3200*1024)
 	tests.Assert(t, 1 == r.BricksInSet())
 
       // Gen7
```

Then you don't need to adapt all the hunks above...

apps/glusterfs/volume_durability_test.go
```diff
@@ -128,27 +128,27 @@ func TestDisperseDurability(t *testing.T) {
 	r.Data = 8
 	r.Redundancy = 3
 
-	gen := r.BrickSizeGenerator(200 * GB)
+	gen := r.BrickSizeGenerator(100 * GB)
```
obnoxxx: Same comment as with other test: leave the 200 here and add another block at the end. (Gen 5)

apps/glusterfs/volume_durability_test.go
```diff
@@ -176,41 +176,41 @@ func TestReplicaDurabilityGenerator(t *testing.T) {
 	r := &VolumeReplicaDurability{}
 	r.Replica = 2
 
-	gen := r.BrickSizeGenerator(100 * GB)
+	gen := r.BrickSizeGenerator(50 * GB)
```
obnoxxx: Same cooment. leave 100 and add a block at the end

apps/glusterfs/app_volume_test.go
MohamedAshiqrh: Sure agreed.

apps/glusterfs/volume_durability_test.go
MohamedAshiqrh: I tired that, It is failing with different error(not ErrMinimumBrickSize).

```
 	sets, brick_size, err = gen()
 	tests.Assert(t, err == nil)
 	tests.Assert(t, sets == 16)
	tests.Assert(t, brick_size == 3200*1024)
 	tests.Assert(t, 1 == r.BricksInSet())
 
       // Gen7
```

It is crossing the 32 max brick count and showing that error.

apps/glusterfs/volume_durability_test.go
MohamedAshiqrh: same as above

apps/glusterfs/volume_durability_test.go
MohamedAshiqrh: same as above.

apps/glusterfs/volume_durability_test.go
MohamedAshiqrh: comment below.

apps/glusterfs/volume_durability_test.go
MohamedAshiqrh: We are not asserting for that error.

obnoxxx: Well, then we should do :-)
Also go additional steps further down, as needed.

Alternatively, we could also change the config that we run the test on to have 4 GB BrickMinSize ... but I don't think we should do that.

apps/glusterfs/volume_durability_test.go
obnoxxx: I'll show you what I mean in a minute or so ...

obnoxxx:
@MohamedAshiqrh here is the modification of your patch that I meant:

https://github.com/obnoxxx/heketi/commit/80dc2df1881c376b0815999a9b6824969c20d326

The diff is not smaller for the generator tests, but I think it is more obvious what's going on.

obnoxxx:
amend commit is broken ;-) (squashed with previous commit, which was a merged commit)

MohamedAshiqrh:
Closing this opened a new PR: https://github.com/heketi/heketi/pull/641

PR #643, opened 2016-12-23
tests: use "become" instead of deprecated "sudo: True" in ansible

obnoxxx:


obnoxxx:
Flakey functional test env...

Issue #630, opened 2016-12-16
Use glide instead of Godeps


lpabon:
This will make it possible to *not* check in the massive `vendor` directory.

PR #655, opened 2017-01-12
Initial commit using glide instead of godeps

ramkrsna:
This resolves the issue in #630 .  The following tools were used ::
```
go get -u github.com/Masterminds/glide
go get github.com/sgotti/glide-vc
```
The `glide.yaml` file is generated from the init command shipped from the tool. Then we run the  glide-vc --only-code --no-tests to delete all unnecessary files from vendor.


humblec:
@ramkrsna we need to update the developer wiki as well with some info on this.

MohamedAshiqrh:
@ramkrsna Nice work. Haven't reviewed just appreciating the work.

ramkrsna:
Although the glide-vc has reduced the footprint of the project, it has removed the license files of the dependencies as it is an "unused" file along with the test files of the dependencies. I need to figure out a way to  preserve the  license files @heketi/dev @lpabon , do suggest the changes required.

obnoxxx:
@centos-ci ok to test

raghavendra-talur:
@ramkrsna @heketi/dev If you look at https://github.com/Masterminds/glide#q-should-i-check-vendor-into-version-control it says you don't need to check in vendor, so we don't need to worry about missing license files. I would recommend that we remove vendor dir from git check-ins.

ramkrsna:
@heketi/dev @raghavendra-talur I have made the necessary changes. we do not require the vendor directory in this commit. You could take a look at the new patch and send in the comments.

lpabon:
@ramkrsna After this change is merged you will need to talk to the gluster infra team to update the scripts that setup the CI system.

lpabon:
Travis was successful because it still uses Godep: https://travis-ci.org/heketi/heketi/jobs/192991662#L201 .  It needs to be updated to remove godep.

lpabon:
@ramkrsna Any updates?

ramkrsna:
I have considered all the suggestions and sent in a new commit in #706

PR #666, opened 2017-02-03
[WIP] Glide Changes 

ramkrsna:
- This change removes the makefiles to work with glide
- travis.yaml is using glide dependency instead of godep


client/cli/go/glide.yaml
```diff
@@ -0,0 +1,15 @@
+package: github.com/heketi/heketi/client/cli/go
+import:
+- package: github.com/heketi/heketi
```
lpabon: Is this package necessary?  Pulling the source?  I could see this causing issues if someone adds a file.

lpabon:
Also, could you please put the removal of all the Godeps onto its own commit?  It makes it much easier to review.

client/cli/go/glide.yaml
ramkrsna: Ack. Will remove this package. I have been having other issues with the Makefile. The client heketi package has its own dependencies; it is trying to create a vendor file, glide.yaml and a lock file. Any idea how to tackle that issue.

ramkrsna:
I have considered all the suggestions and sent in a new commit in #706

PR #671, opened 2017-02-07
Backup database to Kubernetes Secret

### LINKED ISSUE #529: Make use of 'init containers' to perform pre configuration tasks of heketi pod.
At present we make use of 'deploy-heketi' pod which perform initialization tasks before heketi pod comes in picture.  From OSE 3.3 we have something called 'init containers' which are designed for this kind of jobs. It will be really cool, if we can make use of init containers in Heketi deployment process.

### LINKED ISSUE #631: Deploy heketi in a single step/command
I've worked a little on this topic, only few changes are required to be able to deploy heketi in a single step. 
related to:  https://github.com/gluster/gluster-kubernetes/issues/39
The following description is up to discussion / changes but the main idea is we start the heketi-server and let it wait for the pvc to be bound/initialized and topology is loaded automatically from a configmap.

To deploy in one step heketi/glusterfs on kubernetes we can do the following steps:

Deploy kubernetes resources:
-> topology-configmap.yaml
-> heketi-pvc.yaml
-> heketi-service-account.yaml
-> heketi-svc.yaml
-> heketi-glusterfs-daemonset.yaml
-> heketi-deployment.yaml
-> heketi-init-job.yaml

The only file the user must edit is the topology-configmap.yaml, other resources can have working defaults.

In details:

The heketi-deployment.yaml, is the heketi-server:

It mounts the heketi-pvc, because the pvc is not bound, the pod can't be scheduled it waits.
It has an init-container that mounts also heketi-pvc, and wait for a heketi.db file exists to exist
The heketi-init-job.yaml is a job that does:

- mount the topology configmap
- start a heketi-server with emptyDir to store the db
- run heketi-cli --load topology:
- the job can have 2 container in the pod: 1 server | 1 client
- Create a the heketidbstorage glusterfs-volume, the glusterfs-endpoints
- Create a secret from the heketi.db
- Create the persistentvolume heketi-pv.yaml resource.
-> After this job, the heketi-db-secret is created and the pv/pvc are created too, so the heketi-server pod can start. The init container (or the job) mount the secret and copy it if does not exist in /var/lib/heketi

lpabon:
This change enables Heketi to automatically backup the database to a
Kubernetes secret.  It does this by adding a new middleware which
determines if the database needs to be backed up or not by checking the
request and its result.

With this change, there is no longer a need to deploy Heketi twice or
create a GlusterFS volume to store the database.

A future patch will add functional tests once the functional test
framework is updated.

How to test:
Change the image in extras/kubernetes/heketi-deploy.json to
`lpabon/heketi:secretdb`, and the value of HEKETI_EXECUTOR to `mock`
then deploy.  Insert to fake systems using heketi-cli (you can use the
topology-sample.json file in client/cli/go).  Now edit deployment and
change the value of snapshot limit.  This will cause Kubernetes to
redeploy Heketi.  Check the new Heketi server has the data that was
inserted.

Closes #529
Closes #631


lpabon:
@heketi/dev

MohamedAshiqrh:
@lpabon Wow this looks awesome. I just read the commit, I will try this out in a setup tomorrow and share my reviews. nice work.

apps/glusterfs/app_middleware.go
```diff
@@ -0,0 +1,83 @@
...
+	// the Golang http ResponseWriter does not provide access to
+	// the HttpStatus.
+	responsew := negroni.NewResponseWriter(w)
+	next(responsew, r)
+
+	// Only backup for POST and PUT
+	if !a.isAsyncDone(responsew, r) &&
+		r.Method == http.MethodGet {
+		return
+	}
```
jarrpa: The indentation of the above section seems wrong, or at least somewhat misleading.

extras/kubernetes/heketi-deployment.json
```diff
@@ -1,119 +1,142 @@
...
-  "spec": {
-    "selector": {
-        "name": "heketi"
+  "items": [
+    {
+      "kind": "Secret",
+      "apiVersion": "v1",
+      "metadata": {
+        "name": "heketi-db-backup",
+        "labels": {
```
jarrpa: Do we need these labels for anything?

lpabon:
@jarrpa 
1. Yeah, we have to create a secret first so that the mount is satisfied and the pod is started by Kubernetes.
1. No, the mounting is done by Kubernetes, and the copying of the database is only done by the shell script during startup.  All backupdb.go is doing is "putting" the db in a secret over REST.

extras/kubernetes/heketi-deployment.json
lpabon: Maybe to search later, but nothing really _needed_

apps/glusterfs/app_middleware.go
lpabon: ah, Yeah, I can fix that.

lpabon: Plus the comment is "old" and incorrect.

jarrpa:
@lpabon I'm confused, then. Why is backupdb.go trying to create a secret rather than just updating it to begin with? Also, do we need to hardcode the secret name or can that be a configuration parameter?

extras/kubernetes/heketi-deployment.json
jarrpa: I'd say don't include the labels yet if we don't need them, they're unnecessarily very specific and it would make the secret JSON simpler if they were omitted.

lpabon:
@jarrpa It creates if you delete it for some reason.  Other than that, it updates it.  We need to hardcode it because it must match the deployment name.  If we had it configurable, it may not match the deployment volume name and may block forever waiting for the name.

lpabon:
@jarrpa But I think you make a good point and should probably be configurable as long as the administrator know what they are doing.  I'll make it configurable through the config file or env variable.

jarrpa:
LGTM!

lpabon:
Sweet thanks @jarrpa and @MohamedAshiqrh

PR #676, opened 2017-02-14
Remove Device: Used to distribute all the bricks from device to other devices

MohamedAshiqrh:
Important:
We only support this feature for Replicate and Distribute Replicate Volume.
Refer: https://gluster.readthedocs.io/en/latest/Administrator%20Guide/Managing%20Volumes/#replace-brick

Removes all the brick from the device and start these bricks on other
devices based on allocator. Heal is triggered automatially for replicate volumes
on replace brick.

Allocate and create new brick to replace.
It stops the brick to be replaced, If it is not already down(kill the brick process).
Then gluster replace brick which will replace the brick with new one and also starts the heals.

If other nodes does not have sufficient storage then this command should fail.

Pair-Programmed-With: Mohamed Ashiq Liyazudeen mliyazud@redhat.com

### DIFF
--- a/apps/glusterfs/allocator_simple.go
+++ b/apps/glusterfs/allocator_simple.go
@@ -179,6 +179,10 @@ func (s *SimpleAllocator) GetNodes(clusterId, brickId string) (<-chan string,
 	// Get the list of devices for this brick id
 	devicelist, err := s.getDeviceList(clusterId, brickId)
 
+	for _, tester := range devicelist {
+		logger.Debug("Allocator Device ID: %v, Brick ID: %v", tester.deviceId, brickId)
+	}
+
 	if err != nil {
 		errc <- err
 		close(device)
--- a/apps/glusterfs/app.go
+++ b/apps/glusterfs/app.go
@@ -290,6 +290,11 @@ func (a *App) SetRoutes(router *mux.Router) error {
 			Method:      "POST",
 			Pattern:     "/devices/{id:[A-Fa-f0-9]+}/state",
 			HandlerFunc: a.DeviceSetState},
+		rest.Route{
+			Name:        "DeviceRemove",
+			Method:      "POST",
+			Pattern:     "/devices/{id:[A-Fa-f0-9]+}/remove",
+			HandlerFunc: a.DeviceRemove},
 
 		// Volume
 		rest.Route{
--- a/apps/glusterfs/app.go
+++ b/apps/glusterfs/app.go
@@ -143,6 +143,13 @@ func NewApp(configIo io.Reader) *App {
 				return err
 			}
 
+			// Handle Upgrade Changes
+			err = app.Upgrade(tx)
+			if err != nil {
+				logger.LogError("Unable to Upgrade Changes")
+				return err
+			}
+
 			return nil
 
 		})
@@ -188,6 +195,16 @@ func (a *App) setLogLevel(level string) {
 	case "debug":
 		logger.SetLevel(utils.LEVEL_DEBUG)
 	}
+}
+
+// Upgrade Path to update all the values for new API entries
+func (a *App) Upgrade(tx *bolt.Tx) error {
+	err := AddVolumeIdInBrickEntry(tx)
+	if err != nil {
+		logger.LogError("VolumeId add to BrickEntry failed")
+		return err
+	}
+	return nil
 }
 
 func (a *App) setAdvSettings() {
@@ -372,4 +389,39 @@ func (a *App) Backup(w http.ResponseWriter, r *http.Request) {
 func (a *App) NotFoundHandler(w http.ResponseWriter, r *http.Request) {
 	logger.Warning("Invalid path or request %v", r.URL.Path)
 	http.Error(w, "Invalid path or request", http.StatusNotFound)
+}
+
+func AddVolumeIdInBrickEntry(tx *bolt.Tx) error {
+	clusters, err := ClusterList(tx)
+	if err != nil {
+		return err
+	}
+	for _, cluster := range clusters {
+		clusterEntry, err := NewClusterEntryFromId(tx, cluster)
+		if err != nil {
+			return err
+		}
+		for _, volume := range clusterEntry.Info.Volumes {
+			volumeEntry, err := NewVolumeEntryFromId(tx, volume)
+			if err != nil {
+				return err
+			}
+			for _, brick := range volumeEntry.Bricks {
+				brickEntry, err := NewBrickEntryFromId(tx, brick)
+				if err != nil {
+					return err
+				}
+				if brickEntry.Info.VolumeId == "" {
+					brickEntry.Info.VolumeId = volume
+					err = brickEntry.Save(tx)
+					if err != nil {
+						return err
+					}
+				} else {
+					break
+				}
+			}
+		}
+	}
+	return nil
 }
--- a/apps/glusterfs/app_device.go
+++ b/apps/glusterfs/app_device.go
@@ -154,6 +154,56 @@ func (a *App) DeviceAdd(w http.ResponseWriter, r *http.Request) {
 
 }
 
+func (a *App) DeviceRemove(w http.ResponseWriter, r *http.Request) {
+
+	logger.Info("entered deviceremove")
+
+	vars := mux.Vars(r)
+	id := vars["id"]
+
+	// Check request
+	var device *DeviceEntry
+	err := a.db.View(func(tx *bolt.Tx) error {
+		var err error
+		// Access device entry
+		device, err = NewDeviceEntryFromId(tx, id)
+		if err == ErrNotFound {
+			http.Error(w, err.Error(), http.StatusNotFound)
+			return err
+		} else if err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return logger.Err(err)
+		}
+		return nil
+	})
+	if err != nil {
+		return
+	}
+	// Check and remove all bricks from device
+	logger.Info("Check and remove all bricks from device %v", device.Info.Id)
+	a.asyncManager.AsyncHttpRedirectFunc(w, r, func() (string, error) {
+		var err error
+
+		// Check if we can remove the device
+		if device.IsDeleteOk() {
+			logger.Info("Device is clean to delete or remove")
+			return "", nil
+		}
+
+		err = device.Remove(a.db, a.executor, a.allocator)
+		if err != nil {
+			if err == ErrNoReplacement {
+				http.Error(w, device.NoReplacementDeviceString(), http.StatusInternalServerError)
+				logger.LogError(device.NoReplacementDeviceString())
+			}
+			return "", err
+		}
+		logger.Info("exit async")
+		return "", nil
+
+	})
+}
+
 func (a *App) DeviceInfo(w http.ResponseWriter, r *http.Request) {
 
 	// Get device id from URL
--- a/apps/glusterfs/brick_entry.go
+++ b/apps/glusterfs/brick_entry.go
@@ -12,6 +12,7 @@ package glusterfs
 import (
 	"bytes"
 	"encoding/gob"
+
 	"github.com/boltdb/bolt"
 	"github.com/heketi/heketi/executors"
 	"github.com/heketi/heketi/pkg/glusterfs/api"
@@ -36,7 +37,7 @@ func BrickList(tx *bolt.Tx) ([]string, error) {
 }
 
 func NewBrickEntry(size, tpsize, poolMetadataSize uint64,
-	deviceid, nodeid string, gid int64) *BrickEntry {
+	deviceid, nodeid string, gid int64, volumeid string) *BrickEntry {
 
 	godbc.Require(size > 0)
 	godbc.Require(tpsize > 0)
@@ -51,6 +52,7 @@ func NewBrickEntry(size, tpsize, poolMetadataSize uint64,
 	entry.Info.Size = size
 	entry.Info.NodeId = nodeid
 	entry.Info.DeviceId = deviceid
+	entry.Info.VolumeId = volumeid
 
 	godbc.Ensure(entry.Info.Id != "")
 	godbc.Ensure(entry.TpSize == tpsize)
--- a/apps/glusterfs/volume_entry_allocate.go
+++ b/apps/glusterfs/volume_entry_allocate.go
@@ -10,7 +10,10 @@
 package glusterfs
 
 import (
+	"fmt"
+
 	"github.com/boltdb/bolt"
+	"github.com/heketi/heketi/executors"
 	"github.com/heketi/heketi/pkg/utils"
 )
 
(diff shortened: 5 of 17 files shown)


MohamedAshiqrh:
Please DO NOT REVIEW. It is still breaking. Will update on the fixes and all things to do in the PR so that everyone is aware of the progress(and also can share their thoughts).

lpabon:
@MohamedAshiqrh @raghavendra-talur I encourage you guys to try things out as PoCs, but on big changes like this you should first create a design document describing the changes and showing how it will deal with failures.  We can discuss that here in Github or over email.

apps/glusterfs/app.go
```diff
@@ -291,6 +291,11 @@ func (a *App) SetRoutes(router *mux.Router) error {
 			Method:      "POST",
 			Pattern:     "/devices/{id:[A-Fa-f0-9]+}/state",
 			HandlerFunc: a.DeviceSetState},
+		rest.Route{
```
lpabon: The model for device state changes is to use `/devices/{id..}/state`. It is then handled by the Node and Device state changes as seen [in node_entry.go](https://github.com/heketi/heketi/blob/master/apps/glusterfs/node_entry.go#L248), where the value that should be used to fail a node or device is `api.EntryStateFailed`.

apps/glusterfs/volume_entry_allocate.go
```diff
@@ -63,6 +64,165 @@ func (v *VolumeEntry) allocBricksInCluster(db *bolt.DB,
...
+					return err
+				}
+				logger.Info("Remove node %v from allocator", node.Info.Id)
+				err = node.removeAllDisksFromRing(tx, allocator)
+				if err != nil {
+					return err
+				}
+			}
+		}
+		defer func() {
```
lpabon: This may not work since the db.Update() is called in a db.Update().  It should definitely be done another way.

apps/glusterfs/volume_entry_allocate.go
```diff
@@ -63,6 +64,165 @@ func (v *VolumeEntry) allocBricksInCluster(db *bolt.DB,
 	}
 }
 
+func (v *VolumeEntry) replaceBrickInVolume(db *bolt.DB, executor executors.Executor,
```
lpabon: This function is _the_ most important part of the fix.  It needs to make sure that it picks a new brick from the ring that is on a _different_ zone than the ones already used and/or in the same zone if there is only one zone.  The algorithm used here should be discussed first.

apps/glusterfs/app.go
MohamedAshiqrh: @lpabon yeah agreed. so you are suggesting that we can use `/devices/{id:[A-Fa-f0-9]+}` instead of `/devices/{id:[A-Fa-f0-9]+}/remove` with method "POST". Then yes we can do it as we do not require a EntryState.

apps/glusterfs/volume_entry_allocate.go
MohamedAshiqrh: Sure.

apps/glusterfs/volume_entry_allocate.go
MohamedAshiqrh: done.

MohamedAshiqrh:
@heketi/maintainers @heketi/dev The latest update of patch works as expected.
Zone is not yet considered. Will add the changes soon now replace brick is working as expected.

apps/glusterfs/volume_entry_allocate.go
MohamedAshiqrh: We can discuss more about this if you want in mail. 
http://lists.gluster.org/pipermail/heketi-devel/2017-February/000064.html

MohamedAshiqrh: @lpabon ^^^

raghavendra-talur:
@heketi/dev @heketi/maintainers @lpabon We were thinking about the problem of finding the right replacement for the brick. As heketi does not store the brick set information along with the volume or brick information, it is proving to be little tricky. We have two possible solutions:
1. Query the volume info in this function and determine on the equivalent of setList of VolumeCreate function. This seems doable and without breaking any backward compatibility.
2. Store the brick set information while we do a volume create. This however would mean that only those volumes that are created this version onwards are eligible for volume operations like replace brick etc. I know we can also do a one time query of gluster volumes and update the db with brick set info, but I am not inclined to do that.

If we do choose option 1, we don't really have to change anything in the allocator. Please let us know your thoughts.

Also, would you be open to taking this patch in as-is and considering the problem of "finding the right brick replacement" as a bug which we would fix in a week's time.

MohamedAshiqrh:
@heketi/admin @heketi/dev I have added the volumeId to brickEntry and also made the changes to test. I feel the first option can be really helpful as We can support old volumes too. @lpabon Let us know what can be done.

MohamedAshiqrh:
Also there was a small bug which was causing the http to not return as the DB was in transaction lock. The db was opened in view and replaceBrick was called. Inside replaceBrick we again opened the db with update so the transaction was hung in spurious cases. This is fixed in the updated commit.

MohamedAshiqrh:
@heketi/dev @heketi/admin @lpabon Addressed all your comments, The Patch is clean and open for Review.

@raghavendra-talur and I worked on the logic of algorithm and Decisions were taken on the Algorithm based on the support for older volumes.

Prerequisites, Done:
We now added VolumeId in BrickEntry and VolumeInfo Executor call which will return Whole information of volume from gluster Itself(instead of saving the brick peer(brickset), we generate the brick peers from this information). 

How does this work:

For a Device to be remove.
First If the Device is Empty then Return ok to remove.
Else
Get the bricklist for bricks in device to be removed and its appropriate volumeEntrylist for bricks. 
Call Replace brick for a volume with the brickId.

In Replace Brick Logic:
1)First we Find the BrickSet(a set in which brick belongs, For Example in Distribute-Replicate 2x3 [A(A1,A2,A3), B(B1,B2,B3)], B2 is on set B) in which the brick to be replaced is present.
Reason to find this is we should not place the brick with another brick of same set(which will cause Quorum to be met if one node is down and also not a good design).
2) Call the allocator to give out devices for the same cluster.
3)Ignore the Device IF:
         a)Same Device to be removed
         b)Device belongs to same Node where one of the other bricks in Set is present
4) With above logic We can still use the logic of simpleAllocator ring to decide the brick placement with single Zone and Multiple zones.
5) On Failure returns Err and In case of NoSpaceError, We Respond Replacementnotfound.

Note:
Few basic tests added for New VolumeId for BrickEntry and all the failure based on executor.SimpleVolumeInfo change from executor.VolumeInfo has been fixed. 
Kept Device Remove modular so that can be used for Node Remove.

To Be Done: 
Tests to be Added.

Let us know If we missed anything. Open for Reviews.

MohamedAshiqrh:
Looked into the Failing tests and fixed the Patch Hopefully the latest will pass the test.

obnoxxx:
@centos-ci test this, please

executors/sshexec/volume.go
```diff
@@ -228,3 +228,101 @@ func (s *SshExecutor) checkForSnapshots(host, volume string) error {
 
 	return nil
 }
+
+func (s *SshExecutor) VolumeInfo(host string, volume string) (*executors.SingleVolumeInfo, error) {
```
obnoxxx: Could this possibly be added in a separate patch, in order to keep the patches a little smaller?

raghavendra-talur:
@obnoxxx thanks for the review. Will address the comments soon.

apps/glusterfs/allocator_simple_test.go
```diff
@@ -176,7 +176,7 @@ func TestSimpleAllocatorInitFromDbWithOfflineDevices(t *testing.T) {
 		// devices are added to the ring
 		node, err := cluster.NodeEntryFromClusterIndex(tx, 0)
 		tests.Assert(t, err == nil)
-		nodeId = node.Info.Id
+		//	nodeId := node.Info.Id
```
raghavendra-talur: @MohamedAshiqrh I need your help here. I don't remember doing this change, is it necessary?

executors/mockexec/mock.go
```diff
@@ -9,9 +9,7 @@
 
 package mockexec
 
-import (
-	"github.com/heketi/heketi/executors"
-)
+import "github.com/heketi/heketi/executors"
```
raghavendra-talur: Editor defaults, cleaning up.

executors/sshexec/volume.go
raghavendra-talur: Done.

apps/glusterfs/allocator_simple.go
```diff
@@ -179,6 +180,10 @@ func (s *SimpleAllocator) GetNodes(clusterId, brickId string) (<-chan string,
 	// Get the list of devices for this brick id
 	devicelist, err := s.getDeviceList(clusterId, brickId)
 
+	for _, tester := range devicelist {
+		logger.Info("Allocator Device ID: %v, Brick ID: %v", tester.deviceId, brickId)
+	}
```
raghavendra-talur: Also, shall I move this to Debug level?

apps/glusterfs/allocator_simple_test.go
raghavendra-talur: Ok, I think I got it. nodeId is unused, hence the removal?

MohamedAshiqrh: @raghavendra-talur It is not necessary. better to revert all the changes in the file as We are not depending on this.

apps/glusterfs/allocator_simple.go
obnoxxx: Yeah, seems to call out for "Debug" ?

apps/glusterfs/allocator_simple_test.go
obnoxxx: yeah, please remove unrelated changes. If there is a case for them, better do additional patches.

apps/glusterfs/allocator_simple.go
```diff
@@ -10,8 +10,9 @@
 package glusterfs
 
 import (
-	"github.com/boltdb/bolt"
 	"sync"
+
+	"github.com/boltdb/bolt"
```
raghavendra-talur: Editor again, standard packages and github packages are separated by a space and standard packages should come first. Trying to revert this change.

obnoxxx: @raghavendra-talur if there is a case for this change -- make a separate patch (and maybe separate PR).

apps/glusterfs/brick_create.go
```diff
@@ -57,6 +57,7 @@ func createDestroyConcurrently(db *bolt.DB,
 }
 
 func CreateBricks(db *bolt.DB, executor executors.Executor, brick_entries []*BrickEntry) error {
+	logger.Info("In Create Brick")
```
raghavendra-talur: done

apps/glusterfs/app_volume.go
```diff
@@ -167,7 +167,7 @@ func (a *App) VolumeList(w http.ResponseWriter, r *http.Request) {
 
 	var list api.VolumeListResponse
 
-	// Get all the cluster ids from the DB
+	// Get all the volume ids from the DB
```
raghavendra-talur: done

apps/glusterfs/app.go
raghavendra-talur: @lpabon , @MohamedAshiqrh and I discussed this.

We liked that usage of api.* to indicate what to be done on the node/device instead of using another handler and REST URL. However, I don't feel "removed" fits in the hierarchy of states  "online" and "offline". Online and offline are mutually exclusive states and adding "removed" to the same hierarchy would corrupt it. A device would then could be in both "online" and "removed" state. Also, we feel that "failed" state is different than "removed". In remove we are just ensuring that all bricks from the device are moved away. It would be immediately available for new volumes.

In summary, we think, remove is an operation and not a change of state. Let us know what you think.

apps/glusterfs/app.go
```diff
@@ -190,6 +197,15 @@ func (a *App) setLogLevel(level string) {
 	}
 }
 
+// Upgrade Path to update all the values for new API entries
+func (a *App) Upgrade(tx *bolt.Tx) error {
+	err := VolumeIdInBrickEntry(tx)
+	if err != nil {
+		logger.LogError("VolumeId add to BrickEntry failed")
+	}
```
raghavendra-talur: need to return error here
change function name to "FillVolumeIdInBrickEntry"

apps/glusterfs/brick_entry.go
```diff
@@ -12,6 +12,7 @@ package glusterfs
 import (
 	"bytes"
 	"encoding/gob"
+
```
raghavendra-talur: This is recommended practice. One empty line before referencing non-standard packages.

obnoxxx:
@centos-ci retest

obnoxxx:
As far as I can tell, all of @lpabon 's requests have been addressed.
I am going to merge.
Awesome work, @MohamedAshiqrh and @raghavendra-talur ! 👍

apps/glusterfs/app.go
lpabon: What does "It would be immediately available for new volumes." mean?

raghavendra-talur: @lpabon  A new volume create request might pick this device again.

lpabon: @raghavendra-talur Why would we use the device again if the brick data was moved?  Why move it then?

apps/glusterfs/volume_entry_allocate.go
```diff
@@ -63,6 +66,209 @@ func (v *VolumeEntry) allocBricksInCluster(db *bolt.DB,
...
+		if err != nil {
+			return err
+		}
+		oldBrickNode, err := NewNodeEntryFromId(tx, oldBrickEntry.Info.NodeId)
+		if err != nil {
+			return err
+		}
+
+		// Determine the setlist by getting data from Gluster
+		vinfo, err := executor.VolumeInfo(oldBrickNode.ManageHostName(), v.Info.Name)
```
lpabon: Nicely done, but encapsulate line 116 to 138 in a function.

apps/glusterfs/volume_entry_allocate.go
```diff
@@ -63,6 +66,209 @@ func (v *VolumeEntry) allocBricksInCluster(db *bolt.DB,
...
+		var slicestartindex int
+		var foundbrickset bool
+		var brick executors.Brick
+		setlist := make([]*BrickEntry, 0)
+		for slicestartindex = 0; slicestartindex <= len(vinfo.Bricks.Bricks)-v.Durability.BricksInSet(); slicestartindex = slicestartindex + v.Durability.BricksInSet() {
```
lpabon: This needs a detailed explanation of what it is doing.

lpabon: What is Bricks.Bricks?

executors/executor.go
```diff
@@ -69,5 +73,56 @@ type VolumeRequest struct {
...
+	DistCount       int      `xml:"distCount"`
+	StripeCount     int      `xml:"stripeCount"`
+	ReplicaCount    int      `xml:"replicaCount"`
+	ArbiterCount    int      `xml:"arbiterCount"`
+	DisperseCount   int      `xml:"disperseCount"`
+	RedundancyCount int      `xml:"redundancyCount"`
+	Type            int      `xml:"type"`
+	TypeStr         string   `xml:"typeStr"`
+	Transport       int      `xml:"transport"`
+	Bricks          Bricks
```
lpabon: Use another name for the structure

apps/glusterfs/volume_entry_allocate.go
```diff
@@ -63,6 +66,209 @@ func (v *VolumeEntry) allocBricksInCluster(db *bolt.DB,
...
+			setlist = make([]*BrickEntry, 0)
+			for _, brick = range vinfo.Bricks.Bricks[slicestartindex : slicestartindex+v.Durability.BricksInSet()] {
```
lpabon: Use a function which is more descriptive and easier to understand for v.info.Bricks.Bricks[...]...

apps/glusterfs/volume_entry_allocate.go
lpabon: This section would be easier if the bricks-in-set ids are saved in the db.  This would be the backup model.

Instead provide an "Upgrade()" call like the one above for volumes which do not have BricksInSet in the database, and fill in that information  Then in this function, just use the database.

apps/glusterfs/volume_entry_allocate.go
```diff
@@ -63,6 +66,209 @@ func (v *VolumeEntry) allocBricksInCluster(db *bolt.DB,
...
+				brickentry, err := v.getEntryfromBrickName(tx, brick.Name)
+				if err != nil {
+					return err
```
lpabon: We should log here `logger.LogError("Nice explanation: %v", err)`

apps/glusterfs/volume_entry_allocate.go
```diff
@@ -63,6 +66,209 @@ func (v *VolumeEntry) allocBricksInCluster(db *bolt.DB,
...
+				}
+				if brickentry.Id() == oldBrickId {
+					foundbrickset = true
+				} else {
+					setlist = append(setlist, brickentry)
```
lpabon: Nice method

apps/glusterfs/volume_entry_allocate.go
```diff
@@ -63,6 +66,209 @@ func (v *VolumeEntry) allocBricksInCluster(db *bolt.DB,
...
+				}
+			}
+			if foundbrickset {
+				break
+			}
+		}
+		if !foundbrickset {
+			return err
```
lpabon: logger.LogErr

apps/glusterfs/volume_entry_allocate.go
```diff
@@ -63,6 +66,209 @@ func (v *VolumeEntry) allocBricksInCluster(db *bolt.DB,
...
+		}
+
+		for _, brickInSet := range setlist {
+			logger.Info("setlist is %v", brickInSet)
```
lpabon: You can print it all using this: `logger.Info("Bricks in set %+v", setlist)`

apps/glusterfs/volume_entry_allocate.go
```diff
@@ -63,6 +66,209 @@ func (v *VolumeEntry) allocBricksInCluster(db *bolt.DB,
...
+		}
+
+		logger.Info("Ask allocator for the list of devices")
+		// Check the ring for devices to place the brick
+		deviceCh, done, errc := allocator.GetNodes(v.Info.Cluster, newBrickId)
+		defer func() {
+			close(done)
+		}()
+
+		for deviceId := range deviceCh {
```
lpabon: There is a lot of copied code here from the allocator.  Please redesign to minimize the copied code.

apps/glusterfs/volume_entry_allocate.go
```diff
@@ -63,6 +66,209 @@ func (v *VolumeEntry) allocBricksInCluster(db *bolt.DB,
...
+		var newBrickId string
+		newId := false
+		for !newId {
+			newBrickId = utils.GenUUID()
+			if !utils.SortedStringHas(v.BricksIds(), newBrickId) {
+				newId = true
+			}
+		}
+
+		logger.Info("Ask allocator for the list of devices")
```
lpabon: Remove this logging statement, it is confusing.  Or reword with information for the administrator

apps/glusterfs/volume_entry_allocate.go
```diff
@@ -63,6 +66,209 @@ func (v *VolumeEntry) allocBricksInCluster(db *bolt.DB,
...
+			return nil, err
+		}
+		if brickname == fmt.Sprintf("%v:%v", nodeEntry.Info.Hostnames.Storage[0], brickentry.Info.Path) {
+			return brickentry, nil
+		}
+	}
+	return nil, ErrNotFound
+}
+
+func (v *VolumeEntry) replaceBrickInVolume(db *bolt.DB, executor executors.Executor,
```
lpabon: This function is too large.  Please reorganize into one or two more functions.  Feel free to create a new xxx_replace.go file to place all these functions in.

apps/glusterfs/volume_entry_allocate.go
```diff
@@ -63,6 +66,209 @@ func (v *VolumeEntry) allocBricksInCluster(db *bolt.DB,
...
+		}
+
+		for _, brickInSet := range setlist {
+			logger.Info("setlist is %v", brickInSet)
+		}
+
+		//Create an Id for new brick
+		var newBrickId string
+		newId := false
+		for !newId {
```
lpabon: I don't think this is needed.  There none of this code in any other part of Heketi.  You can assume there will be no collisions.

apps/glusterfs/volume_entry_allocate.go
```diff
@@ -63,6 +66,209 @@ func (v *VolumeEntry) allocBricksInCluster(db *bolt.DB,
...
+
+			// Do not allow a device from the same node to be
+			// in the set
+			deviceOk := true
+			for _, brickInSet := range setlist {
+				if brickInSet.Info.NodeId == newDeviceEntry.NodeId {
+					deviceOk = false
+				}
+			}
+
```
lpabon: **Must have:**

Add a check here for Zone.  If the devices in the brick set not in the same zone, the new device must also not be in the same zone.

apps/glusterfs/volume_entry_allocate.go
```diff
@@ -63,6 +66,209 @@ func (v *VolumeEntry) allocBricksInCluster(db *bolt.DB,
...
+			logger.Info("Volume entry: Device ID: %v,Node ID: %v", newDeviceEntry.Id, newDeviceEntry.NodeId)
+			if !deviceOk {
+				continue
+			}
+			logger.Info("Got the device with id %v", newDeviceEntry.Id())
```
lpabon: Delete this log because we still do not know if the device will be accepted.

lpabon:
First I want to add that this is really solid work.  The algorithm has one bug, but other than that it is very sound.  Unfortunately due to the number of issues shown, the code will be reverted.  Please, fix the above issues and resubmit the code.  This change is very important.

@MohamedAshiqrh @raghavendra-talur

apps/glusterfs/app.go
```diff
@@ -290,6 +307,11 @@ func (a *App) SetRoutes(router *mux.Router) error {
 			Method:      "POST",
 			Pattern:     "/devices/{id:[A-Fa-f0-9]+}/state",
 			HandlerFunc: a.DeviceSetState},
+		rest.Route{
+			Name:        "DeviceRemove",
+			Method:      "POST",
+			Pattern:     "/devices/{id:[A-Fa-f0-9]+}/remove",
```
lpabon: Another way it to use `DeviceDelete` to be able to delete a device and re-arrange all of its bricks.

apps/glusterfs/volume_entry_allocate.go
raghavendra-talur: @lpabon If I am not wrong, this is already done in allocator code. Even in allocBricks we don't check for zones and rely on allocator for that.

lpabon: You have brought up a good point, but the allocator does not do that.  Which actually, is a bigger issue because this is also a problem with the normal allocation of bricks, where they could go into the same zone.

The allocator creates an array of the following form for a topology of three zones: [z1d0n0 z2d0n0, z3d0n0, z1d1n0, z2d1n0, z3d1n0, z1d0n1, z3d1n0, z3d1n0, and so on..].  So if z1d0n0 is the disk that needs to be replaced, it may choose z3d0n0.  But in the set for z1d0n0, it could have [z1d0n0, z2d1n0, z3d2n1], and if it chose z3d0n0 for replacing z1d0n0, now it has two drives in the same zone.  Instead it should make sure to go through the drives looking _first_ for drives _not_ in a zone already in the set.  Then if it does not find one because they have no space, then we may have to pick one in a zone which is already used as a fallback.

This sounds like it could be an issue also in the normal allocation, so solving it here we may use it as a fix for the normal allocation.

I'm not opposed to pushing this to its own fix after this if you want were we can concentrate the fix for both allocations in that one fix and provide the appropriate tests.  Up to you.

apps/glusterfs/volume_entry_allocate.go
raghavendra-talur: I am hesitating to populate db with this much of info because I want those changes to be properly designed. I request this change be taken in in current form. As it isn't a API change, it can be done in subsequent patches without breaking backward compatibility.

apps/glusterfs/app.go
raghavendra-talur: This is also available in pr #710

apps/glusterfs/app.go
raghavendra-talur: done

apps/glusterfs/app_device.go
```diff
@@ -154,6 +154,56 @@ func (a *App) DeviceAdd(w http.ResponseWriter, r *http.Request) {
...
+			return logger.Err(err)
+		}
+		return nil
+	})
+	if err != nil {
+		return
+	}
+	// Check and remove all bricks from device
+	logger.Info("Check and remove all bricks from device %v", device.Info.Id)
+	a.asyncManager.AsyncHttpRedirectFunc(w, r, func() (string, error) {
```
raghavendra-talur: done

apps/glusterfs/app_device.go
```diff
@@ -154,6 +154,56 @@ func (a *App) DeviceAdd(w http.ResponseWriter, r *http.Request) {
...
+		var err error
+
+		// Check if we can remove the device
+		if device.IsDeleteOk() {
+			logger.Info("Device is clean to delete or remove")
+			return "", nil
+		}
+
+		err = device.Remove(a.db, a.executor, a.allocator)
```
raghavendra-talur: done with the functional change. Need more opinion on *naming*

apps/glusterfs/volume_entry_allocate.go
raghavendra-talur: I agree, changed the member name to be more descriptive. Will convert to function in refactor PR.

apps/glusterfs/volume_entry_allocate.go
raghavendra-talur: yes, will do.

Issue #402, opened 2016-06-30
 Device delete does not delete vg directory created under "/var/lib/heketi/mounts"


NehaRawat:
Device delete does not delete the vg directory created under /var/lib/heketi/mounts

lpabon:
https://bugzilla.redhat.com/show_bug.cgi?id=1351454

PR #686, opened 2017-02-21
Remove the VG directory on Device remove

MohamedAshiqrh:
Device Delete should check if the vg directory present or not and if present then remove the directory.

Closes: #402 


obnoxxx:
@centos-ci test this, please

obnoxxx:
LGTM. merging.

But we should think about whether we can write a test to verify that the directory really vanishes... @MohamedAshiqrh

PR #688, opened 2017-02-23
Remove redundant comments

obnoxxx:
Working on the code, I found a lot of redundant/superfluous comments that state exactly the obvious thing (what is done right after that). sometimes even falsely (copy and paster error?) Here are two patches to remove these from two code files I worked on...

### DIFF
--- a/executors/sshexec/volume.go
+++ b/executors/sshexec/volume.go
@@ -49,8 +49,11 @@ func (s *SshExecutor) VolumeCreate(host string,
 		maxPerSet = 1
 	}
 
-	// There could many, many bricks which could make the command line
-	// too long.  Instead, create the volume first, then add each brick set.
+	// There could many, many bricks, which could render a single command
+	// line that creates the volume with all the bricks too long.
+	// Therefore, we initially create the volume with the first brick set
+	// only, and then add each brick set in one subsequent command.
+
 	for _, brick := range volume.Bricks[:inSet] {
 		cmd += fmt.Sprintf("%v:%v ", brick.Host, brick.Path)
 	}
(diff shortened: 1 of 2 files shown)


executors/sshexec/volume.go
```diff
@@ -51,23 +49,18 @@ func (s *SshExecutor) VolumeCreate(host string,
...
 	// There could many, many bricks which could make the command line
 	// too long.  Instead, create the volume first, then add each brick set.
 	for _, brick := range volume.Bricks[:inSet] {
 		cmd += fmt.Sprintf("%v:%v ", brick.Host, brick.Path)
 	}
 
-	// Initialize the commands with the create command
 	commands := []string{cmd}
 
-	// Now add all the commands to add the bricks
```
raghavendra-talur: I request this one line to stay as a continuation of comments on lines 55,56. I know it is self explanatory but it is a "different way of doing things" and would be good to have a comment to get attention of developer.

obnoxxx: Thanks for the comment @raghavendra-talur .
I still that this individual comment is bogus.
Instead, I have expanded the comment block explaining our strategy,
and separated it from the for-loop by an empty line, indicating that it
does not relate to the next line or block but to the following commands,
each of which are imho pretty idiomatic and natural to read.
(I just want to avoid the pattern
```
  // now add commands to add broicks
  s.createAddBrickCommands(...)
```
which is kind-of the definition of redundancy ;-)

obnoxxx:
@raghavendra-talur additional patch added. please check if you like it better. :-)

obnoxxx:
thanks @raghavendra-talur.

PR #701, opened 2017-03-02
open db in read only mode when FS is read only

ramkrsna:
This patch is a PoC to overcome the limitations of Boltdb in which, the BoltDb does not detect a read-only filesystem, hence heketi does not open db in read only mode in such cases. 
As a workaround we can detect read-only filesystem in heketi and handle it.

Signed-off-by: Ramakrishna Reddy Yekulla <rreddy@redhat.com>

### DIFF
--- a/apps/glusterfs/app.go
+++ b/apps/glusterfs/app.go
@@ -52,6 +52,22 @@ type App struct {
 	xo *mockexec.MockExecutor
 }
 
+func checkBoltDbWrite(app *App) (err error) {
+		err = app.db.Update(func(tx *bolt.Tx) error {
+			// Create Cluster Bucket
+			_, err := tx.CreateBucketIfNotExists([]byte("dummy"))
+			if err != nil {
+				logger.LogError("Unable to create dummy cluster bucket in DB")
+				return err
+			}
+
+			return nil
+		})
+
+		return  err
+}
+
+
 // Use for tests only
 func NewApp(configIo io.Reader) *App {
 	app := &App{}
@@ -98,7 +114,20 @@ func NewApp(configIo io.Reader) *App {
 		logger.Warning("Unable to open database.  Retrying using read only mode")
 
 		// Try opening as read-only
-		app.db, err = bolt.Open(dbfilename, 0666, &bolt.Options{
+		app.db, err = bolt.Open(dbfilename, 0600, &bolt.Options{
+			ReadOnly: true,
+		})
+		if err != nil {
+			logger.LogError("Unable to open database: %v", err)
+			return nil
+		}
+		app.dbReadOnly = true
+	} else if nil != checkBoltDbWrite(app) {
+		logger.Warning("write failed. Retrying using read only mode")
+                app.db.Close()
+
+		// Try opening as read-only
+		app.db, err = bolt.Open(dbfilename, 0600, &bolt.Options{
 			ReadOnly: true,
 		})
 		if err != nil {


ramkrsna:
@obnoxxx @lpabon @heketi/dev please suggest if we are going in the right direction to solve the limitation of BoltDB.

apps/glusterfs/app.go
```diff
@@ -98,7 +114,20 @@ func NewApp(configIo io.Reader) *App {
 		logger.Warning("Unable to open database.  Retrying using read only mode")
 
 		// Try opening as read-only
-		app.db, err = bolt.Open(dbfilename, 0666, &bolt.Options{
+		app.db, err = bolt.Open(dbfilename, 0600, &bolt.Options{
```
humblec: According to 'https://github.com/boltdb/bolt/#read-only-mode , it should be '0666' .

apps/glusterfs/app.go
```diff
@@ -98,7 +114,20 @@ func NewApp(configIo io.Reader) *App {
...
+			ReadOnly: true,
+		})
+		if err != nil {
+			logger.LogError("Unable to open database: %v", err)
+			return nil
+		}
+		app.dbReadOnly = true
+	} else if nil != checkBoltDbWrite(app) {
```
humblec: This `if` condition looks weird to me. Need some cleanup.

apps/glusterfs/app.go
raghavendra-talur: The mode of the file is used for creation of the db file. It does decide with what permissions it will open it in.

apps/glusterfs/app.go
raghavendra-talur: :) https://en.wikipedia.org/wiki/Yoda_conditions

apps/glusterfs/app.go
```diff
@@ -98,7 +114,20 @@ func NewApp(configIo io.Reader) *App {
...
+		logger.Warning("write failed. Retrying using read only mode")
+                app.db.Close()
+
+		// Try opening as read-only
+		app.db, err = bolt.Open(dbfilename, 0600, &bolt.Options{
```
MohamedAshiqrh: I think we should do 666 here. It worked for me. can you add more detail why not 600?

apps/glusterfs/app.go
MohamedAshiqrh: I am not completely sure about open permission here.

MohamedAshiqrh:
> ./heketi --config=/heketi-var.json      
> Heketi v4.0.0-22-g6816149-pr_701
> [heketi] INFO 2017/03/02 19:28:35 Loaded mock executor
> [heketi] WARNING 2017/03/02 19:28:35 write failed. Retrying using read only mode
> [heketi] INFO 2017/03/02 19:28:35 Loaded simple allocator
> [heketi] INFO 2017/03/02 19:28:35 GlusterFS Application Loaded
> Listening on port 3333
> [negroni] Started GET /clusters
> [negroni] Completed 200 OK in 113.429µs
> [negroni] Started POST /clusters
> [negroni] Completed 0  in 16.087µs
> [negroni] Started GET /clusters
> [negroni] Completed 200 OK in 62.061µs

looks good.

MohamedAshiqrh:
LGTM. overall good. just not clear on permission. Might be silly confusion I have. 

I have tried with db permission 0444 and gluster volume in readonly.

I checked for regression expected the heketi.db with proper permission It worked \o/.

@lpabon @obnoxxx Please take a look.

apps/glusterfs/app.go
raghavendra-talur: Again, 666 would also work, but why should we create the db with more wider access if not required?

apps/glusterfs/app.go
lpabon: Good point guys.  I'm not sure the purpose of changing the value if there is no need to.

lpabon:
I'm confused on the requirement for this change: `the BoltDb does not detect a read-only filesystem`.

Does that mean that the `open()` call returns successfully for RDRW, but the filesystem is running in read-only mode?  Isn't this a bug in the file system then?  Does this really need to be fixed here?

If the file system is running in read-only mode, I would think that an open call with RDRW flag should return a failure

obnoxxx:
open() with O_RDWR on a RO-filesystem returns -1 with errno == EROFS
open() with O_RDWR on a RO file returns -1 with error == EACCESS

ramkrsna:
@obnoxxx @lpabon This would infer that there is a bug on GlusterFS . @heketi/dev  Any idea if there is already a bug opened on GlusterFS for this case for other products?

humblec:
@ramkrsana can you  remount a local fs where u mounted db path in read only mode and try to start heketi

PR #708, opened 2017-03-07
Add a doc/design/ directory for future purposes

obnoxxx:
This could become a place for design documents and design proposals.
Following the kubernetes example, we could file PRs against (new) documents in this folder as Design-PRs for discussing designs for new features.

### DIFF
--- a/doc/design/README.md
+++ b/doc/design/README.md
@@ -1,0 +1,3 @@
+# Heketi Design Documents and Proposals
+
+This directory contains Heketi design documents and accepted design proposals.


obnoxxx:
I would like to add a design for the block support, and thought this might be a good mechanism to do design reviews. What do people think. @lpabon ?

ramkrsna:
Should this be part of the Heketi project, We could have another repository in the Heketi organization for all the Blueprints, Design PR's with some formal guideline around the proposals.

obnoxxx:
@ramkrsna Thanks for the comments. My thought is I would not like to add more process than needed, keeping the hurdle low to get involved. I think it's more likely for people to spot design prs directly in the heketi repo. Maybe we should not name the directory doc/design-proposals/ but rather doc/design.

Also, I would not put a formal guideline for a process. Let's rehearse this in a few cases is my thought. If we need a formal process then we can still establish it later. Heketi is a small project and hopefully does not need too elaborate processes.

my 2 cents...

raghavendra-talur:
My opinion is to have the design docs to be in same repo. If the feature does not make it for a release, we should delete the design doc when we do the branching.

obnoxxx:
Updated the PR to have the directory as doc/design/ instead of doc/design-proposals/. After all, after a design PR is merged this will contain the accepted design.

lpabon:
I have had the experience with both.  In OpenStack they add the docs to the same repo.  In Kubernetes is to add them to the _features_ repo.  I think that this project is small enough to keep the requests on the same repo.  Just my thought.

obnoxxx:
It seems that consensus has been reached (talked to @ramkrsna on irc). Merging this for now.

PR #710, opened 2017-03-09
Remove device

raghavendra-talur:
We support this feature only for Replicate and Distribute Replicate Volume.
Refer: https://gluster.readthedocs.io/en/latest/Administrator%20Guide/Managing%20Volumes/#replace-brick

Removes all the brick from the device and start these bricks on other
devices based on allocator. Heal is triggered automatially for replicate volumes
on replace brick.

Allocate and create new brick to replace.
It stops the brick to be replaced, If it is not already down(kill the brick process).
Then gluster replace brick which will replace the brick with new one and also starts the heals.

If other nodes does not have sufficient storage then this command should fail.

Pair-Programmed-With: Mohamed Ashiq Liyazudeen mliyazud@redhat.com

### DIFF
--- a/apps/glusterfs/app.go
+++ b/apps/glusterfs/app.go
@@ -199,11 +199,37 @@ func (a *App) setLogLevel(level string) {
 
 // Upgrade Path to update all the values for new API entries
 func (a *App) Upgrade(tx *bolt.Tx) error {
-	err := AddVolumeIdInBrickEntry(tx)
+
+	err := ClusterEntryUpgrade(tx)
 	if err != nil {
-		logger.LogError("VolumeId add to BrickEntry failed")
+		logger.LogError("Failed to upgrade db for cluster entries")
 		return err
 	}
+
+	err = NodeEntryUpgrade(tx)
+	if err != nil {
+		logger.LogError("Failed to upgrade db for node entries")
+		return err
+	}
+
+	err = VolumeEntryUpgrade(tx)
+	if err != nil {
+		logger.LogError("Failed to upgrade db for volume entries")
+		return err
+	}
+
+	err = DeviceEntryUpgrade(tx)
+	if err != nil {
+		logger.LogError("Failed to upgrade db for device entries")
+		return err
+	}
+
+	err = BrickEntryUpgrade(tx)
+	if err != nil {
+		logger.LogError("Failed to upgrade db for brick entries")
+		return err
+	}
+
 	return nil
 }
 
@@ -389,39 +415,4 @@ func (a *App) Backup(w http.ResponseWriter, r *http.Request) {
 func (a *App) NotFoundHandler(w http.ResponseWriter, r *http.Request) {
 	logger.Warning("Invalid path or request %v", r.URL.Path)
 	http.Error(w, "Invalid path or request", http.StatusNotFound)
-}
-
-func AddVolumeIdInBrickEntry(tx *bolt.Tx) error {
-	clusters, err := ClusterList(tx)
-	if err != nil {
-		return err
-	}
-	for _, cluster := range clusters {
-		clusterEntry, err := NewClusterEntryFromId(tx, cluster)
-		if err != nil {
-			return err
-		}
-		for _, volume := range clusterEntry.Info.Volumes {
-			volumeEntry, err := NewVolumeEntryFromId(tx, volume)
-			if err != nil {
-				return err
-			}
-			for _, brick := range volumeEntry.Bricks {
-				brickEntry, err := NewBrickEntryFromId(tx, brick)
-				if err != nil {
-					return err
-				}
-				if brickEntry.Info.VolumeId == "" {
-					brickEntry.Info.VolumeId = volume
-					err = brickEntry.Save(tx)
-					if err != nil {
-						return err
-					}
-				} else {
-					break
-				}
-			}
-		}
-	}
-	return nil
 }
--- a/apps/glusterfs/app.go
+++ b/apps/glusterfs/app.go
@@ -143,6 +143,13 @@ func NewApp(configIo io.Reader) *App {
 				return err
 			}
 
+			// Handle Upgrade Changes
+			err = app.Upgrade(tx)
+			if err != nil {
+				logger.LogError("Unable to Upgrade Changes")
+				return err
+			}
+
 			return nil
 
 		})
@@ -188,6 +195,16 @@ func (a *App) setLogLevel(level string) {
 	case "debug":
 		logger.SetLevel(utils.LEVEL_DEBUG)
 	}
+}
+
+// Upgrade Path to update all the values for new API entries
+func (a *App) Upgrade(tx *bolt.Tx) error {
+	err := AddVolumeIdInBrickEntry(tx)
+	if err != nil {
+		logger.LogError("VolumeId add to BrickEntry failed")
+		return err
+	}
+	return nil
 }
 
 func (a *App) setAdvSettings() {
@@ -290,6 +307,11 @@ func (a *App) SetRoutes(router *mux.Router) error {
 			Method:      "POST",
 			Pattern:     "/devices/{id:[A-Fa-f0-9]+}/state",
 			HandlerFunc: a.DeviceSetState},
+		rest.Route{
+			Name:        "DeviceRemove",
+			Method:      "POST",
+			Pattern:     "/devices/{id:[A-Fa-f0-9]+}/remove",
+			HandlerFunc: a.DeviceRemove},
 
 		// Volume
 		rest.Route{
@@ -367,4 +389,39 @@ func (a *App) Backup(w http.ResponseWriter, r *http.Request) {
 func (a *App) NotFoundHandler(w http.ResponseWriter, r *http.Request) {
 	logger.Warning("Invalid path or request %v", r.URL.Path)
 	http.Error(w, "Invalid path or request", http.StatusNotFound)
+}
+
+func AddVolumeIdInBrickEntry(tx *bolt.Tx) error {
+	clusters, err := ClusterList(tx)
+	if err != nil {
+		return err
+	}
+	for _, cluster := range clusters {
+		clusterEntry, err := NewClusterEntryFromId(tx, cluster)
+		if err != nil {
+			return err
+		}
+		for _, volume := range clusterEntry.Info.Volumes {
+			volumeEntry, err := NewVolumeEntryFromId(tx, volume)
+			if err != nil {
+				return err
+			}
+			for _, brick := range volumeEntry.Bricks {
+				brickEntry, err := NewBrickEntryFromId(tx, brick)
+				if err != nil {
+					return err
+				}
+				if brickEntry.Info.VolumeId == "" {
+					brickEntry.Info.VolumeId = volume
+					err = brickEntry.Save(tx)
+					if err != nil {
+						return err
+					}
+				} else {
+					break
+				}
+			}
+		}
+	}
+	return nil
 }
--- a/apps/glusterfs/app.go
+++ b/apps/glusterfs/app.go
@@ -333,11 +333,6 @@ func (a *App) SetRoutes(router *mux.Router) error {
 			Method:      "POST",
 			Pattern:     "/devices/{id:[A-Fa-f0-9]+}/state",
 			HandlerFunc: a.DeviceSetState},
-		rest.Route{
-			Name:        "DeviceRemove",
-			Method:      "POST",
-			Pattern:     "/devices/{id:[A-Fa-f0-9]+}/remove",
-			HandlerFunc: a.DeviceRemove},
 
 		// Volume
 		rest.Route{
--- a/apps/glusterfs/app_device.go
+++ b/apps/glusterfs/app_device.go
@@ -154,51 +154,6 @@ func (a *App) DeviceAdd(w http.ResponseWriter, r *http.Request) {
 
 }
 
-func (a *App) DeviceRemove(w http.ResponseWriter, r *http.Request) {
-
-	vars := mux.Vars(r)
-	id := vars["id"]
-
-	// Check request
-	var device *DeviceEntry
-	err := a.db.View(func(tx *bolt.Tx) error {
-		var err error
-		// Access device entry
-		device, err = NewDeviceEntryFromId(tx, id)
-		if err == ErrNotFound {
-			http.Error(w, err.Error(), http.StatusNotFound)
-			return err
-		} else if err != nil {
-			http.Error(w, err.Error(), http.StatusInternalServerError)
-			return logger.Err(err)
-		}
-		return nil
-	})
-	if err != nil {
-		return
-	}
-	// Check and remove all bricks from device
-	a.asyncManager.AsyncHttpRedirectFunc(w, r, func() (string, error) {
-		var err error
-
-		// Check if we can remove the device
-		if device.IsDeleteOk() {
-			return "", nil
-		}
-
-		err = device.Remove(a.db, a.executor, a.allocator)
-		if err != nil {
-			if err == ErrNoReplacement {
-				http.Error(w, err.Error(), http.StatusInternalServerError)
-				logger.LogError("Unable to delete device [%v] as no device was found to replace it", device)
-			}
-			return "", err
-		}
-		return "", nil
-
-	})
-}
-
 func (a *App) DeviceInfo(w http.ResponseWriter, r *http.Request) {
 
 	// Get device id from URL
@@ -367,35 +322,31 @@ func (a *App) DeviceSetState(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, "request unable to be parsed", 422)
 		return
 	}
+	logger.Info("entered set state state is %+v, device is %v", msg, id)
 
 	// Set state
-	err = a.db.Update(func(tx *bolt.Tx) error {
-		device, err := NewDeviceEntryFromId(tx, id)
+	a.asyncManager.AsyncHttpRedirectFunc(w, r, func() (string, error) {
+		err = a.db.Update(func(tx *bolt.Tx) error {
+			device, err := NewDeviceEntryFromId(tx, id)
+			if err != nil {
+				logger.LogError("Did not find device requested %v", id)
+				return err
+			}
+
+			// Set state
+			err = device.SetState(tx, a.executor, a.allocator, msg.State)
+			if err != nil {
+				return err
+			}
+			return nil
+		})
 		if err == ErrNotFound {
 			http.Error(w, "Id not found", http.StatusNotFound)
-			return err
+			return "", err
 		} else if err != nil {
 			http.Error(w, err.Error(), http.StatusInternalServerError)
-			return err
+			return "", err
 		}
-
-		// Set state
-		err = device.SetState(tx, a.allocator, msg.State)
-		if err != nil {
-			http.Error(w, err.Error(), http.StatusBadRequest)
-			return err
-		}
-
-		// Save new state
-		err = device.Save(tx)
-		if err != nil {
-			http.Error(w, err.Error(), http.StatusInternalServerError)
-			return err
-		}
-
-		return nil
+		return "", nil
 	})
-	if err != nil {
-		return
-	}
 }
--- a/apps/glusterfs/app_device.go
+++ b/apps/glusterfs/app_device.go
@@ -189,8 +189,8 @@ func (a *App) DeviceRemove(w http.ResponseWriter, r *http.Request) {
 		err = device.Remove(a.db, a.executor, a.allocator)
 		if err != nil {
 			if err == ErrNoReplacement {
-				http.Error(w, device.NoReplacementDeviceString(), http.StatusInternalServerError)
-				logger.LogError(device.NoReplacementDeviceString())
+				http.Error(w, err.Error(), http.StatusInternalServerError)
+				logger.LogError("Unable to delete device [%v] as no device was found to replace it", device)
 			}
 			return "", err
 		}
--- a/apps/glusterfs/app_device.go
+++ b/apps/glusterfs/app_device.go
@@ -314,6 +314,7 @@ func (a *App) DeviceSetState(w http.ResponseWriter, r *http.Request) {
 	// Get the id from the URL
 	vars := mux.Vars(r)
 	id := vars["id"]
+	var device *DeviceEntry
 
 	// Unmarshal JSON
 	var msg api.StateRequest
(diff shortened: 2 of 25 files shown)


raghavendra-talur:
This does not have the API changes yet. I wanted to send out the simpler stuff for review first.

raghavendra-talur:
This PR now has the complete changes for API. Also the unit tests have been fixed.

What is remaining and will be done is different PRs
1. Refactor of big function replaceBrickInVolume()
2. Improve/fix the allocator code to account for zones while picking the brick devices

@lpabon @MohamedAshiqrh @humblec @ramkrsna @jarrpa  PR ready for review.

apps/glusterfs/app.go
```diff
@@ -190,6 +197,42 @@ func (a *App) setLogLevel(level string) {
 	}
 }
 
+// Upgrade Path to update all the values for new API entries
+func (a *App) Upgrade(tx *bolt.Tx) error {
```
lpabon: :+1:

apps/glusterfs/app_device.go
```diff
@@ -322,35 +322,31 @@ func (a *App) DeviceSetState(w http.ResponseWriter, r *http.Request) {
...
+	a.asyncManager.AsyncHttpRedirectFunc(w, r, func() (string, error) {
+		err = a.db.Update(func(tx *bolt.Tx) error {
+			device, err := NewDeviceEntryFromId(tx, id)
+			if err != nil {
+				logger.LogError("Did not find device requested %v", id)
+				return err
+			}
+
+			// Set state
+			err = device.SetState(tx, a.executor, a.allocator, msg.State)
```
lpabon: Transferring the `tx` instead of the `db` here is causing concurrency issues.  For example, this change introduces locks and unlocks in the concurrent brick creation were there were none before.  It looks to me that the Update() call can chage the state first in the `tx` context, then the db context can be freed.  This will let the bricks all be executed in parallel, which is one of the benefits of Heketi.  By introducing locks (I haven't checked yet), but it may create an issue on many nodes and volumes because there will no longer be parallelism.

apps/glusterfs/brick_create.go
```diff
@@ -34,11 +34,13 @@ func createDestroyConcurrently(db *bolt.DB,
 		sg.Add(1)
 		go func(b *BrickEntry) {
 			defer sg.Done()
+			sg.Lock()
```
lpabon: This is not good, because this essentially would make Heketi take a long time on large number of parallel bricks. Instead, review how to use the existing infrastructure from the state change call.  You can also use "db" key locks in the way the Register/Unregister work.  For example, if you are "deleting" a brick, you cannot change its state until the state change has fully happened.  You can use the db to place it in "XXXING" state until it gets to "XXX" state.

apps/glusterfs/brick_entry.go
```diff
@@ -51,6 +52,7 @@ func NewBrickEntry(size, tpsize, poolMetadataSize uint64,
 	entry.Info.Size = size
 	entry.Info.NodeId = nodeid
 	entry.Info.DeviceId = deviceid
+	entry.Info.VolumeId = volumeid
```
lpabon: :+1:

apps/glusterfs/brick_entry.go
```diff
@@ -164,28 +160,22 @@ func (b *BrickEntry) Create(db *bolt.DB, executor executors.Executor) error {
 	return nil
 }
 
-func (b *BrickEntry) Destroy(db *bolt.DB, executor executors.Executor) error {
+func (b *BrickEntry) Destroy(tx *bolt.Tx, executor executors.Executor) error {
```
lpabon: :-1:

apps/glusterfs/brick_entry.go
```diff
@@ -239,3 +229,46 @@ func (b *BrickEntry) DestroyCheck(db *bolt.DB, executor executors.Executor) erro
...
+				return err
+			}
+			for _, brick := range volumeEntry.Bricks {
+				brickEntry, err := NewBrickEntryFromId(tx, brick)
+				if err != nil {
+					return err
+				}
+				if brickEntry.Info.VolumeId == "" {
+					brickEntry.Info.VolumeId = volume
+					err = brickEntry.Save(tx)
```
lpabon: Nice.  In a way think of the db context as something that can be accomplished immediately, without going to the executor.

apps/glusterfs/device_entry.go
```diff
@@ -243,12 +260,23 @@ func (d *DeviceEntry) SetState(tx *bolt.Tx,
 			if err != nil {
 				return err
 			}
+			d.State = s
+			err = d.Save(tx)
```
lpabon: If it is already in offline state, why save the state again?

apps/glusterfs/device_entry.go
```diff
@@ -243,12 +260,23 @@ func (d *DeviceEntry) SetState(tx *bolt.Tx,
...
+			if err != nil {
+				return err
+			}
 		case api.EntryStateFailed:
-			// Only thing to do here is to set the state
+
+			err := d.Remove(tx, e, a)
```
lpabon: This is the most important call.

It looks like this function could instead be passed `db` like volume.Create(), and let it pick, when and when not to stay in context.

apps/glusterfs/device_entry.go
```diff
@@ -393,3 +421,72 @@ func (d *DeviceEntry) poolMetadataSize(tpsize uint64) uint64 {
...
+func (d *DeviceEntry) Remove(tx *bolt.Tx,
+	executor executors.Executor,
+	allocator Allocator) (e error) {
+	type brickToReplace struct {
+		brickId     string
+		volumeEntry *VolumeEntry
+	}
+
+	// If the device has no bricks, just change the state and we are done
+	if d.IsDeleteOk() {
```
lpabon: :+1:

apps/glusterfs/device_entry.go
```diff
@@ -393,3 +421,72 @@ func (d *DeviceEntry) poolMetadataSize(tpsize uint64) uint64 {
...
+			return err
+		}
+		bricksToReplace = append(bricksToReplace, brickToReplace{
+			brickId:     brickId,
+			volumeEntry: volumeEntry,
+		})
+	}
+	// Move bricks now
+	for _, brick := range bricksToReplace {
+		logger.Info("Replacing brick %v on device %v on node %v",
```
lpabon: :+1:

apps/glusterfs/volume_entry.go
```diff
@@ -330,15 +330,24 @@ func (v *VolumeEntry) Create(db *bolt.DB,
 	}()
 
 	// Create the bricks on the nodes
-	err = CreateBricks(db, executor, brick_entries)
+	err = db.Update(func(tx *bolt.Tx) error {
```
lpabon: These changes are not necessary

apps/glusterfs/volume_entry_allocate.go
```diff
@@ -63,6 +66,187 @@ func (v *VolumeEntry) allocBricksInCluster(db *bolt.DB,
...
+		return err
+	}
+
+	// Determine the setlist by getting data from Gluster
+	vinfo, err := executor.VolumeInfo(oldBrickNode.ManageHostName(), v.Info.Name)
+	var slicestartindex int
+	var foundbrickset bool
+	var brick executors.Brick
+	setlist := make([]*BrickEntry, 0)
+	for slicestartindex = 0; slicestartindex <= len(vinfo.Bricks.BrickList)-v.Durability.BricksInSet(); slicestartindex = slicestartindex + v.Durability.BricksInSet() {
```
lpabon: This needs a comment explaining what it is doing.

apps/glusterfs/volume_entry_allocate.go
```diff
@@ -63,6 +66,187 @@ func (v *VolumeEntry) allocBricksInCluster(db *bolt.DB,
...
+		}
+		v.BrickAdd(newBrickEntry.Id())
+		v.removeBrickFromDb(tx, oldBrickEntry)
+		err = v.Save(tx)
+		if err != nil {
+			logger.Err(err)
+			return err
+		}
+
+		logger.Info("replaced brick:%v on node:%v at path:%v with brick:%v on node:%v at path:%v",
```
lpabon: :+1:

pkg/utils/statusgroup.go
```diff
@@ -15,6 +15,7 @@ import (
 
 type StatusGroup struct {
 	wg      sync.WaitGroup
+	lock    sync.Mutex
```
lpabon: Please remove.

apps/glusterfs/device_entry.go
raghavendra-talur: This is being moved from offline to online.

apps/glusterfs/app_device.go
obnoxxx: Right... Using locks reduces parallelism. But the parallelism is also problematic. @raghavendra-talur can you explain in more detail, what problem you are solving with the introduction of the locks?

raghavendra-talur:
Ok, I did the mistake of assuming that all failures of unit tests are shown in a run. There are more fixes in tests to come.

@lpabon, I will update this PR with removal of tx and passing db.  It would be helpful if you could reply to the email in heketi-devel about the same as there are a few more points in email.

apps/glusterfs/brick_entry.go
raghavendra-talur: I like this idea. Will add this part to Upgrade() function, but as a different PR.

apps/glusterfs/volume_entry_allocate.go
raghavendra-talur: Handled in next PR which refactors this file.

raghavendra-talur:
@lpabon Changed the db related code to pass around db and have the transaction for smallest possible time.

Thanks a lot for the awesome unit tests you have, they increase developer confidence so much!

pkg/utils/statusgroup.go
raghavendra-talur: done

lpabon:
@raghavendra-talur @MohamedAshiqrh 

During my verification (I haven't reviewed yet) I found this issue:

```
[heketi] INFO 2017/03/14 10:47:08 Replacing brick 3001c8c571b7f9dd9ea3bd2eaa8c8b5e on device /dev/sdg on node d7b757b29f5dc42031b8a109d1d8c639
[heketi] ERROR 2017/03/14 10:47:08 /src/github.com/heketi/heketi/apps/glusterfs/volume_entry_allocate.go:154: Unable to find brick set for brick 3001c8c571b7f9dd9ea3bd2eaa8c8b5e, db is possibly corrupt
[heketi] ERROR 2017/03/14 10:47:08 /src/github.com/heketi/heketi/apps/glusterfs/device_entry.go:495: Failed to replace brick device %v on node %v on %!(EXTRA string=3001c8c571b7f9dd9ea3bd2eaa8c8b5e, string=/dev/sdg, string=d7b757b29f5dc42031b8a109d1d8c639)
[asynchttp] INFO 2017/03/14 10:47:08 Completed job 01307705074284bba7937d4e5b7f62ad in 725.793µs
[negroni] Started GET /queue/01307705074284bba7937d4e5b7f62ad
[negroni] Completed 500 Internal Server Error in 147.849µs
```

When I ran the command:

```
$ ./heketi-cli device remove 77b254c70f71931f4a2e4b8675b2bd06
Error: Failed to replace brick device %v on node %v on %!(EXTRA string=3001c8c571b7f9dd9ea3bd2eaa8c8b5e, string=/dev/sdg, string=d7b757b29f5dc42031b8a109d1d8c639)
```

There is also a printf() error on that.

How I created the test:

```
$ ./heketi --config=heketi.json  <-- Where the config was copied from /etc/heketi.json and edited the db location to the local directory
```

Then in another window I ran:

```
$ cd client/cli/go
 1012  export HEKETI_CLI_SERVER=http://localhost:8080
 1013  ./heketi-cli topology info
 1014  ls
 1015  ./heketi-cli topology load --json=topology-sample.json
 1016  ./heketi-cli topology info
 1017  ./heketi-cli volume create --size=4096
 1018  ./heketi-cli volume create --size=1024
 1019  ./heketi-cli topology info
 1020  ./heketi-cli volume create --size=1024
 1021  ./heketi-cli volume create --size=200
 1022  ./heketi-cli topology info
 1023  ./heketi-cli volume create --size=150
 1024  ./heketi-cli volume create --size=100
 1025  ./heketi-cli volume create --size=50
 1026  ./heketi-cli topology info
 1027  ./heketi-cli volume create --size=200
 1028  ./heketi-cli topology info
 1029  ./heketi-cli volume create --size=500
 1030  ./heketi-cli topology info
 1031  ./heketi-cli volume list
 1032  ./heketi-cli volume info 0842109914247beba852643f1c773b7c
 1033  ./heketi-cli volume delete 9c12a01499217306717cf4bfb2c63c12
 1034  ./heketi-cli topology info
 1035  ls
 1036  vim topology-sample.json 
 1037  ./heketi-cli topology load --json=topology-sample.json
 1038  ./heketi-cli topology info
 1039  ./heketi-cli device -h
 1040  ./heketi-cli device remove -h
 1041  ./heketi-cli device remove 77b254c70f71931f4a2e4b8675b2bd06
 1042  ./heketi-cli device remove -h
 1043  ./heketi-cli topology info
 1044  ./heketi-cli device disable  77b254c70f71931f4a2e4b8675b2bd06
 1045  ./heketi-cli topology info
 1046  ./heketi-cli device remove 77b254c70f71931f4a2e4b8675b2bd06
```

I have also attached the gzipped db file for use.
[heketi.db.gz](https://github.com/heketi/heketi/files/841712/heketi.db.gz)

lpabon:
You probably want to write or add to the functional smoke tests `tests/functional/TestSmokeTest/tests` to make it easier to test.

lpabon:
@raghavendra-talur @MohamedAshiqrh One more.  I tested with Mock executor.  Sounds like that could be the issue also.

lpabon:
I'll test again with gluster

raghavendra-talur:
@lpabon Yes, you are right. Mock version of VolumeInfo is returning nil, which leads to this. I always tested with sshexec and real gluster setup. The difficulty here is with determining the number of bricks. 

Thinking of ways to solve that.

raghavendra-talur:
@lpabon @obnoxxx Added 3 more commits.

2 are log related and required.
Mock related is just to prove that it was a mock error. I am trying to find a better way to Mock.

lpabon:
@raghavendra-talur @MohamedAshiqrh 

Ok, I now tried to verify using sshexec (specifically, I used TestSmokeTest) and found an error.

How to reproduce:

First I started a new test in TestSmokeTest/test:

```go
func TestPull(t *testing.T) {
    teardownCluster(t)
    setupCluster(t)
}
```

Then bring up the vagrant setup using `TestSmokeTest/vagrant`.  Once up, I ran the heketi server and used the config file in `TestSmokeTest/config`.

Once that is setup, I can go into `TestSmokeTest/test` and run the following command to setup the topology: `go test -tags functional -v -run TestPull`

Ok, Now, I ran the following commands using heketi-cli:

```
 1006  export HEKETI_CLI_SERVER=http://localhost:8080
 1007  ./heketi-cli get nodes
 1008  ./heketi-cli topology info
 1009  ./heketi-cli volume create --size=750
 1010  ./heketi-cli volume create --size=1024
 1011  ./heketi-cli topology info
 1012  ./heketi-cli volume create --size=2048
 1013  ./heketi-cli volume create --size=1024
 1014  ./heketi-cli topology info
 1015  ./heketi-cli device disable b8970b9ea78025bd7b75eae967910ebb
 1016  ./heketi-cli topology info
 1017* ./heketi-cli device remove b8970b9ea78025bd7b75
 1018  ./heketi-cli topology info
 1019  ./heketi-cli device online  b8970b9ea78025bd7b75eae967910ebb
 1020  ./heketi-cli device online b8970b9ea78025bd7b75eae967910ebb
 1021  ./heketi-cli device enable  b8970b9ea78025bd7b75eae967910ebb
 1022  ./heketi-cli topology info
 1023  ./heketi-cli device offline 36c67deb96c8da5815e72f64ad15d97b
 1024  ./heketi-cli device disable 36c67deb96c8da5815e72f64ad15d97b
 1025  ./heketi-cli topology info
 1026  ./heketi-cli device remove 36c67deb96c8da5815e72f64ad15d97b
```

Then got error:

```
$ ./heketi-cli device remove 36c67deb96c8da5815e72f64ad15d97b
Error: Failed to replace brick device %v on node %v on %!(EXTRA string=4fbc9c2da6f7e5892f5b587922772ff9, string=/dev/vdf, string=fc06b29dc9544bdbf60a863bc938c9e3)
```

I attached log file and db file
[logs.txt](https://github.com/heketi/heketi/files/842336/logs.txt)
[heketi.db.gz](https://github.com/heketi/heketi/files/842337/heketi.db.gz)

lpabon:
I highly recommend creating a functional test in TestSmokeTest (like I did called TestPull()) which creates many volumes, then finds a device to remove with a single brick and multiple bricks.

raghavendra-talur:
@lpabon I wrote a similar test in TestSmokeTest and it seems to be working for me. Will upload it, checking what is different between mine and yours. I do have the latest commits though.

raghavendra-talur:
@lpabon Got the error, sending tests and a fix.

raghavendra-talur:
@lpabon The bug that Luis observed is fixed. It would happen if there are two or more bricks of same volume reside on the same device. I am not able to write a testcase to match exact error. 

@MohamedAshiqrh @obnoxxx Please have a look at the last commit.

tests/functional/TestSmokeTest/tests/heketi_test.go
```diff
@@ -300,3 +300,92 @@ func TestHeketiCreateVolumeWithGid(t *testing.T) {
...
+			tests.Assert(t, err == nil)
+
+			for _, device := range nodeInfo.DevicesInfo {
+				logger.Info("deviceInfo %+v %v", device.Id, device.DeviceInfo.Storage.Used)
+			}
+		}
+	}
+
+
+	//stateReq := &api.StateRequest{}
```
lpabon: I think the tests are commented out.

lpabon:
@raghavendra-talur You can go through the devices on the topology and get a list of bricks on the device.  If the device has two or more bricks, then disable it and remove it.

See https://github.com/heketi/heketi/wiki/API#device_info

raghavendra-talur:
@lpabon Update with a test. Found one more bug while writing the test case, fixed it too. Please review.

@obnoxxx @MohamedAshiqrh @humblec @jarrpa @ramkrsna Need more eyes on this. Please review.

MohamedAshiqrh:
@raghavendra-talur on it 👍 .

obnoxxx:
@raghavendra-talur thanks!! Since we collaborated on the last stretch of the effort, creating the test case and fixing the remaining pieces in the code, I already have a very good opinion of the current state, but will take one more overall look later.

apps/glusterfs/device_entry.go
```diff
@@ -393,3 +433,73 @@ func (d *DeviceEntry) poolMetadataSize(tpsize uint64) uint64 {
...
+		}
+		logger.Info("Replacing brick %v on device %v on node %v", brickEntry.Id(), d.Id(), d.NodeId)
+		err = volumeEntry.replaceBrickInVolume(db, executor, allocator, brickEntry.Id())
+		if err != nil {
+			return logger.Err(err)
+		}
+	}
+
+	// Get new entry for the device because db would have changed
+	// replaceBrickInVolume calls functions that change device state in db
```
lpabon: I think the comment means: "Set state on to failed on removed device", right?

apps/glusterfs/volume_entry_allocate.go
```diff
@@ -63,6 +66,226 @@ func (v *VolumeEntry) allocBricksInCluster(db *bolt.DB,
...
+
+		// Do not allow a device from the same node to be
+		// in the set
+		deviceOk := true
+		for _, brickInSet := range setlist {
+			if brickInSet.Info.NodeId == newDeviceEntry.NodeId {
+				deviceOk = false
+			}
+		}
+
```
lpabon: Create a new issue to track the needed zone logic that should be added here.  That way we won't forget

lpabon:
:tada: Passed verification! 

Quick question.  If a `kill -9 <pid>` has been sent to the process on gluster, but then a replacement brick has not been found because there is no more space, what happens then?

apps/glusterfs/volume_entry_allocate.go
raghavendra-talur: Created #716 for zone logic.
Created #717 for moving the brick set determination logic to Upgrade() part.

apps/glusterfs/device_entry.go
raghavendra-talur: Yup, added that line.

apps/glusterfs/app_device.go
```diff
@@ -324,33 +324,24 @@ func (a *App) DeviceSetState(w http.ResponseWriter, r *http.Request) {
...
-		} else if err != nil {
-			http.Error(w, err.Error(), http.StatusInternalServerError)
-			return err
-		}
-
-		// Set state
-		err = device.SetState(tx, a.allocator, msg.State)
+	a.asyncManager.AsyncHttpRedirectFunc(w, r, func() (string, error) {
+		var device *DeviceEntry
+		err = a.db.View(func(tx *bolt.Tx) error {
```
raghavendra-talur: done

MohamedAshiqrh:
@lpabon kill on the gluster birck process is called only after finding the replacement brick. executor function call happens with two bricks one old and one new. Technically if the device is bad the brick process will be down, If it is up for some reason we do kill -9 on that process. 

Do you have any case where this can happen, For me I don't see executor being called before finding space for new brick.

obnoxxx:
@lpabon In volumeEntry.replaceBrickInVolume(), we first get new replacement bricks. Only if we found a new brick, we continue with executor.VolumeReplaceBrick() which does the kill of the old brick server and then calls gluster volume replace-brick. So I think we are safe?

apps/glusterfs/volume_entry_allocate.go
```diff
@@ -105,6 +106,10 @@ func (v *VolumeEntry) replaceBrickInVolume(db *bolt.DB, executor executors.Execu
 	var oldBrickNodeEntry *NodeEntry
 	var newBrickNodeEntry *NodeEntry
 
+	if api.DurabilityDistributeOnly == v.Info.Durability.Type {
+		return fmt.Errorf("replace brick is not supported for volume durability type %v", v.Info.Durability.Type)
```
obnoxxx: good point!

obnoxxx:
the changes LGTM now 👍 . Good stuff, @raghavendra-talur

apps/glusterfs/device_entry.go
```diff
@@ -476,7 +476,7 @@ func (d *DeviceEntry) Remove(db *bolt.DB,
 		logger.Info("Replacing brick %v on device %v on node %v", brickEntry.Id(), d.Id(), d.NodeId)
 		err = volumeEntry.replaceBrickInVolume(db, executor, allocator, brickEntry.Id())
 		if err != nil {
-			return logger.Err(err)
+			return logger.Err(fmt.Errorf("Failed to remove device, error: %v", err))
```
lpabon: FYI, this works, but you could have used logger.LogError() to do the same in one function.

apps/glusterfs/volume_entry_allocate.go
lpabon: Nice find.

PR #714, opened 2017-03-14
return the status detail of a volume

harryge00:
Currently, the db only contains the basic information on a volume. Users cannot get details such like the available size and total size of a device. So I add an API which can run `gluster --mode=script volume status vol detail --xml` on the remote machine through ssh, and get the result back as json.


executors/sshexec/volume.go
```diff
@@ -142,6 +142,25 @@ func (s *SshExecutor) VolumeDestroy(host string, volume string) error {
 	return nil
 }
 
+func (s *SshExecutor) VolumeStatusDetail(host string, volume string) (string, error) {
```
raghavendra-talur: This functionality is also implemented in PR #710 , I suggest you wait for that patch to be merged to avoid conflicts.

harryge00:
@raghavendra-talur Thanks. I have not seen your PR. I will close this.Hope it can merge soon.

Issue #32, opened 2015-06-24
Volumes should have state


lpabon:
Volumes should show what state they are in: Started, Created, Stopped, Zombie (tried delete, but didn't work)

phlogistonjohn:
This issue has not had any activity in years. Plus, I don't think the current architecture is compatible with this intent of this issue. Closing.

PR #718, opened 2017-03-17
error out replace brick if self heal is in progress

raghavendra-talur:
When the brick to be replaced is a "source" brick in a replicate or disperse volume, we should fail the replace brick. This should be ideally handled in Gluster but no harm in having extra check in Heketi.

### DIFF
--- a/apps/glusterfs/volume_entry_allocate.go
+++ b/apps/glusterfs/volume_entry_allocate.go
@@ -147,6 +147,7 @@ func (v *VolumeEntry) replaceBrickInVolume(db *bolt.DB, executor executors.Execu
 	var foundbrickset bool
 	var brick executors.Brick
 	setlist := make([]*BrickEntry, 0)
+	var onlinePeerBrickCount = 0
 	// BrickList in volume info is a slice of all bricks in volume
 	// We loop over the slice in steps of BricksInSet()
 	// If brick to be replaced is found in an iteration, other bricks in that slice form the setlist
@@ -171,6 +172,43 @@ func (v *VolumeEntry) replaceBrickInVolume(db *bolt.DB, executor executors.Execu
 	if !foundbrickset {
 		logger.LogError("Unable to find brick set for brick %v, db is possibly corrupt", oldBrickEntry.Id())
 		return ErrNotFound
+	}
+
+	// Get self heal status for this brick's volume
+	healinfo, err := executor.HealInfo(node, v.Info.Name)
+	if err != nil {
+		return err
+	}
+	for _, brickHealStatus := range healinfo.Bricks.BrickList {
+		// Gluster has a bug that it does not send Name for bricks that are down.
+		// Skip such bricks; it is safe because it is not source if it is down
+		if brickHealStatus.Name == "information not available" {
+			continue
+		}
+		iBrickEntry, err := v.getBrickEntryfromBrickName(db, brickHealStatus.Name)
+		if err != nil {
+			return fmt.Errorf("Unable to determine heal status of brick")
+		}
+		if iBrickEntry.Id() == oldBrickEntry.Id() {
+			// If we are here, it means the brick to be replaced is
+			// up and running. We need to ensure that it is not a
+			// source for any files.
+			if brickHealStatus.NumberOfEntries != "-" &&
+				brickHealStatus.NumberOfEntries != "0" {
+				return fmt.Errorf("Cannot replace brick %v as it is source brick for data to be healed", iBrickEntry.Id())
+			}
+		}
+		for _, brickInSet := range setlist {
+			if brickInSet.Id() == iBrickEntry.Id() {
+				onlinePeerBrickCount++
+			}
+		}
+	}
+	if onlinePeerBrickCount < v.Durability.QuorumBrickCount() {
+		return fmt.Errorf("Cannot replace brick %v as only %v of %v "+
+			"required peer bricks are online",
+			oldBrickEntry.Id(), onlinePeerBrickCount,
+			v.Durability.QuorumBrickCount())
 	}
 
 	//Create an Id for new brick
--- a/executors/executor.go
+++ b/executors/executor.go
@@ -26,6 +26,7 @@ type Executor interface {
 	VolumeExpand(host string, volume *VolumeRequest) (*Volume, error)
 	VolumeReplaceBrick(host string, volume string, oldBrick *BrickInfo, newBrick *BrickInfo) error
 	VolumeInfo(host string, volume string) (*Volume, error)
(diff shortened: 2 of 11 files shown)


raghavendra-talur:
@heketi/maintainers This is still work in progress. I need to add tests, but would like your review if the approach is right or not.

obnoxxx:
@raghavendra-talur The approach looks good and the code clean enough. But as you said: it's WIP, tests are missing.

raghavendra-talur:
@obnoxxx yes, WIP tag because of no tests. I will add a test,

apps/glusterfs/volume_entry_allocate.go
```diff
@@ -131,6 +131,29 @@ func (v *VolumeEntry) replaceBrickInVolume(db *bolt.DB, executor executors.Execu
 		return err
 	}
 
+	// Get self heal status for this brick's volume
+	healinfo, err := executor.HealInfo(oldBrickNodeEntry.ManageHostName(), v.Info.Name)
```
humblec: Getting heal info only from one node may be not sufficient, cant we have a fallback ?

apps/glusterfs/volume_entry_allocate.go
```diff
@@ -131,6 +131,29 @@ func (v *VolumeEntry) replaceBrickInVolume(db *bolt.DB, executor executors.Execu
...
+	if err != nil {
+		return err
+	}
+	for _, brickHealStatus := range healinfo.Bricks.BrickList {
+		// Gluster has a bug that it does not send Name for bricks that are down.
+		// Skip such bricks; it is safe because it is not source if it is down
+		if brickHealStatus.Name == "information not available" {
```
humblec: Are there any status available than this ?

executors/sshexec/volume.go
```diff
@@ -312,3 +312,32 @@ func (s *SshExecutor) VolumeReplaceBrick(host string, volume string, oldBrick *e
...
+		HealInfo executors.HealInfo `xml:"healInfo"`
+	}
+
+	command := []string{
+		fmt.Sprintf("gluster --mode=script volume heal %v info --xml", volume),
+	}
+
+	output, err := s.RemoteExecutor.RemoteCommandExecute(host, command, 10)
+	if err != nil {
+		return nil, fmt.Errorf("Unable to get heal info of volume name: %v", volume)
```
humblec: you can remove "name" from the above log.

executors/sshexec/volume.go
```diff
@@ -312,3 +312,32 @@ func (s *SshExecutor) VolumeReplaceBrick(host string, volume string, oldBrick *e
...
+	}
+	var healInfo CliOutput
+	err = xml.Unmarshal([]byte(output[0]), &healInfo)
+	if err != nil {
+		return nil, fmt.Errorf("Unable to determine heal info of volume name: %v", volume)
```
humblec: same here

lpabon:
@raghavendra-talur Sounds like this is a state of the Volume. If the volume is self-healing, then I'm assuming that no other state can change can occur until either canceled or completed, right?

obnoxxx:
@lpabon yes, it's a state of the volume, and I guess no other state change should occur, except for possibly remove? Not sure about expansion.

lpabon:
Closes #32

lpabon:
We also need some tests. Any status @raghavendra-talur  ?

raghavendra-talur:
@lpabon Added tests, ready for review.

While adding tests, found a bug in Gluster which affects heketi. Filed a bug in Gluster https://bugzilla.redhat.com/show_bug.cgi?id=1473026 and accounted for the same in heketi.

@obnoxxx @MohamedAshiqrh @jarrpa @humblec @ramkrsna Please have a look.

raghavendra-talur:
@heketi/dev ping

lpabon:
@raghavendra-talur Sweet! What else is needed. What do you suggest we do with this PR?

raghavendra-talur:
@lpabon nothing else pending here. Just review and merge.

executors/executor.go
```diff
@@ -26,6 +26,7 @@ type Executor interface {
 	VolumeExpand(host string, volume *VolumeRequest) (*Volume, error)
 	VolumeReplaceBrick(host string, volume string, oldBrick *BrickInfo, newBrick *BrickInfo) error
 	VolumeInfo(host string, volume string) (*Volume, error)
+	HealInfo(host string, volume string) (*HealInfo, error)
```
lpabon: Do not return GlusterFS specific XML data back to Heketi. The executor should interpret the information and the it should return what the result is, instead of letting the caller figure it out.

raghavendra-talur: @lpabon We need to match the db data with gluster data. So I should either send db data to executor or get executor data to app_* part. I don't see any way to reconcile data without doing so.
Any suggestions?

obnoxxx: @raghavendra-talur : I think @lpabon means not to return the XML data raw, but have the executor extract all info we need into an internal structure (defined by heketi/your needs). Does not mean to lose any data.

lpabon: Correct. The executor's job is to abstract any of the fine details of the call. This is letting the caller deal with raw XML data. Instead let the executor look at the XML data and return a simple struct of what the call is supposed to provide.

jarrpa: @lpabon I was trying to have a look at this today, and I don't understand your request. it looks like there's already a function that extracts the relevant data into a structure? https://github.com/heketi/heketi/pull/718/files#diff-2320cf68f8e9328f659fe584a9a33620

lpabon: @jarrpa I checked it again, and since this is information about the system, it could be that at some time we look at other values in the structure. 

I remove this request. It is ok as it is.

lpabon:
@raghavendra-talur please resolve the conflict.

@heketi/dev If the team is ok with this change it LGTM.

lpabon:
@raghavendra-talur @obnoxxx I would like to see this pass the CI functional tests after the conflict is resolve. Also, @raghavendra-talur could you create a new issue for Release 6 to create a functional test for this change to test it? Thanks.

obnoxxx:
@lpabon since @raghavendra-talur is not available currently, I rebased the branch and pushed a new PR in #834 . PTAL there...

obnoxxx:
@lpabon, the straight rebase in PR #834 passed the travis tests and the ci functional tests. hence, i'm going to merge according to your LGTM above.

obnoxxx:
@lpabon also created the issue #835 for creating a functional test.

obnoxxx:
it's done via merging #834 and creating #835.

lpabon:
:+1:

PR #719, opened 2017-03-17
have "Id:" as prefix for each cluster id

raghavendra-talur:
We have "Id:" as prefix for Id in most of the cli outputs.
Adding it to cluster list output too for easy parsing.


obnoxxx:
LGTM

PR #724, opened 2017-03-22
sshexec: Environmental variable support for config

MohamedAshiqrh:
sshexec config values to support environmental variable inputs.
We have the same for Kubeexec. This patch will enable to same for sshexec. 

Signed-off-by: Mohamed Ashiq Liyazudeen <mliyazud@redhat.com>

PR #726, opened 2017-03-23
Suggestion - External Gluster from OCP POC [DO NOT MERGE]

mangirdaz:
Opening this just to start the discussion about the validity of this solution and if we can agree on something will rebase and do it properly as per agreement. 

I can see a big need to have heketi running in OCP/Origin and manage external/already existing Gluster installation. 

Did POC where SSH keys are mounted as a secret together with a configuration file. 
Modified Image to have SSHD running in the image.
Tested with external gluster and storageclasses and all things works. 

I would like to see this kind of configuration as a separate image with sshd installed, as to have it the main one would increase possible attack vector for this image if user does not use ssh executor. 

Any ideas how this could fit to wider Heketi + Gluster = CNC storage roadmap?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -114,3 +114,12 @@ release: darwin_amd64_dist linux_arm64_dist linux_arm_dist linux_amd64_dist
 
 .PHONY: server client test clean name run version release \
         darwin_amd64_dist linux_arm_dist linux_amd64_dist linux_arm64_dist
+
+custombuild:
+	@cp heketi ./extras/docker/ci/ 
+	@cp ./extras/docker/fromsource/heketi-start.sh ./extras/docker/ci/ 
+	@cp ./extras/docker/fromsource/heketi.json ./extras/docker/ci/ 
+	@cp ./extras/docker/fromsource/adhoc-setup.sh ./extras/docker/ci/ 
+	cd ./extras/docker/ci ;\
+	docker build --rm --tag mangirdas/heketi:dev . ;\
+	docker push mangirdas/heketi:dev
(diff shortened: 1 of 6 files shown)


lpabon:
@mangirdaz I think this is a great idea. I think you are on the right path. What does the @heketi/dev think?

MohamedAshiqrh:
@mangirdaz great job. @lpabon Looks good, have not given a try of this. https://github.com/heketi/heketi/pull/724 can be used, but I dont see a place where the conf file changes being done for the heketi to start as sshexec.

obnoxxx:
@mangirdaz indeed, as @MohamedAshiqrh mentioned, PR https://github.com/heketi/heketi/pull/724 made it possible to have heketi inside kubernetes control and outside gluster. You can see how this is used in the gluster-kubernetes project in PR https://github.com/gluster/gluster-kubernetes/pull/281 .

Note that heketi can today *not* manage pre-existing clusters, but asumes that it gets the IPs of raw nodes which it turns into a trusted storage pool itself. That is irresepctive of whether the gluster nodes are inside or outside of kubernetes.

mangirdaz:
K. So we can close this one.

obnoxxx:
@mangirdaz thanks for your inputs! Please try the gluster-kubernetes project as a means for deploying heketi / gluster. It has a test environement for kubernetes. (I have also put one for origin up as a PR at some point, but that's probably outdated and needs to be rebased and fixed).

And if your idea is not fully covered, please keep up the discussion! :-)

Issue #707, opened 2017-03-07
Setting volume parameters via Heketi


andrewlove:
Is there a way to set GlusterFS volume parameters via Heketi? For example:
```cluster.consistent-metadata on
performance.write-behind off
performance.quick-read off
performance.readdir-ahead off```

Ideally I want to set these when I make a dynamic volume claim in OpenShift. heketi-cli would be good too.

MohamedAshiqrh:
@andrewlove We are working on something similar.

See https://github.com/heketi/heketi/pull/751 .

This can add more options like encryption, small-file perf, etc options

humblec:
@andrewlove it will be `in` soon . I will provide an update as soon as its in :).

Issue #711, opened 2017-03-09
volume create request should have a volume option field with it.


humblec:
At the moment, heketi dont have a field for specifying gluster volume options. Its good to have a new member so that the client or caller can pass volume options in the request.

obnoxxx:
I think this is not necessarily the right apporach:
One of heketi's principles is that it shields the lower level details from the caller.
Handing out low-level volume options to the caller of heketi seems to be against these principles. (similar to allowing to specifiy an explicit cluster ID.)

We might think about having some higher level options that call out for optimizing a volume config for certain workloads (small file  / large file / etc).

@humblec can you elaborate on why you would want a volume options field?

humblec:
@obnoxxx u r fast :) .  I completely agree on the optimization or QOS based setting done by heketi its own. I am mainly looking for volume options like `client.ssl ` , `server.ssl` ..etc which has to be set on volumes to support encryption. What we can do here is, kind of whitelisted volume options in heketi and enable it if the user request comes in for those and discard the option if it does not belong to the allowed list of options.

obnoxxx:
Ok, that makes sense. But we should still think about wrapping the gluster-options into some higher-level settings in heketi. like 'encryptable' or 'encrypted', etc. not sure what's the best approach here, just requesting to think about whether the path of least resistance is right...

jarrpa:
@obnoxxx I actually agree with @humblec on this one. If we try to wrap everything, it becomes heketi's job to keep pace with any changes to Gluster's options. I see nothing wrong with allowing an advanced user access to the lower-level functionality, and in fact think it's regressive to block it entirely.

obnoxxx:
I withdraw my concerns raised before.
Exposing the options to the outside still seems like a small layer violation, but it will allow us to quickly reach certain effects without further code changes.

alikhajeh1:
I've seen https://github.com/heketi/heketi/issues/711 and just wondering how (or if) you guys see Heketi integrating into Openshift/K8 to allow users to define the volume options via labels on PVCs. I'm not sure if cluster admins would want users to edit all options, so maybe they can define a whitelist of options that can be changed by users.

Our use case is for ElasticSearch, which requires `cluster.consistent-metadata on` on a gluster volume. As it stands, we're asking users to add a label to their PVC at the time of creation, then have to watch for the labels and use `oc exec gluster_pod` to update that gluster volume option after it's been created.

alikhajeh1:
[...]

Issue #171, opened 2015-08-29
Need better error descriptions when on ssh failures


lpabon:
That way the caller does not see `Process exited with: 255. Reason was:  ()`

lpabon:
@qe Please take a look at this bug.

lpabon:
@heketi/qe Please take a look at this bug

PR #751, opened 2017-05-02
Set Gluster Volume options based on user input.

humblec:
At present there is no method to provide volume options via
    heketi commandline or via api client. With this patch we provide
    an option to user for specifying various volume options.
    
    For eg#:
    
    '
    $ heketi-cli volume create --size=100 --durability=none --options="performance.rda-cache-limit 10MB","performance.nl-cache-positive-entry no"
    '
    
    Issue ##711
    
    Signed-off-by: Mohammed Ashiq mliyazud@redhat.com
    Signed-off-by: Humble Chirammal hchiramm@redhat.com

### DIFF
--- a/apps/glusterfs/volume_entry_create.go
+++ b/apps/glusterfs/volume_entry_create.go
@@ -97,6 +97,7 @@ func (v *VolumeEntry) createVolumeRequest(db *bolt.DB,
 	// Setup volume information in the request
 	vr.Name = v.Info.Name
 	v.Durability.SetExecutorVolumeRequest(vr)
+	vr.GlusterVolumeOptions = v.GlusterVolumeOptions
 
 	return vr, sshhost, nil
 }
--- a/apps/glusterfs/volume_entry_create.go
+++ b/apps/glusterfs/volume_entry_create.go
@@ -11,11 +11,12 @@ package glusterfs
 
 import (
 	"fmt"
+	"strings"
+
 	"github.com/boltdb/bolt"
 	"github.com/heketi/heketi/executors"
 	"github.com/heketi/heketi/pkg/utils"
 	"github.com/lpabon/godbc"
-	"strings"
 )
 
 func (v *VolumeEntry) createVolume(db *bolt.DB,
--- a/client/cli/go/cmds/volume.go
+++ b/client/cli/go/cmds/volume.go
@@ -23,20 +23,21 @@ import (
 )
 
 var (
-	size           int
-	volname        string
-	durability     string
-	replica        int
-	disperseData   int
-	redundancy     int
-	gid            int64
-	snapshotFactor float64
-	clusters       string
-	expandSize     int
-	id             string
-	kubePvFile     string
-	kubePvEndpoint string
-	kubePv         bool
+	size                 int
+	volname              string
+	durability           string
+	replica              int
+	disperseData         int
+	redundancy           int
+	gid                  int64
+	snapshotFactor       float64
+	clusters             string
+	expandSize           int
+	id                   string
+	kubePvFile           string
+	kubePvEndpoint       string
+	kubePv               bool
+	glusterVolumeOptions string
 )
 
 func init() {
@@ -78,6 +79,9 @@ func init() {
 			"\n\ton any of the configured clusters which have the available space."+
 			"\n\tProviding a set of clusters will ensure Heketi allocates storage"+
 			"\n\tfor this volume only in the clusters specified.")
+	volumeCreateCommand.Flags().StringVar(&glusterVolumeOptions, "gluster-volume-options", "",
+		"\n\tOptional: Comma separated list of volume options which can be set on the volume."+
+			"\n\tIf omitted, Heketi will set no volume option for the volume.")
 	volumeCreateCommand.Flags().BoolVar(&kubePv, "persistent-volume", false,
 		"\n\tOptional: Output to standard out a persistent volume JSON file for OpenShift or"+
 			"\n\tKubernetes with the name provided.")
@@ -126,6 +130,9 @@ var volumeCreateCommand = &cobra.Command{
   * Create a 100GB erasure coded 8+3 volume with 25GB snapshot storage:
       $ heketi-cli volume create --size=100 --durability=disperse --snapshot-factor=1.25 \
         --disperse-data=8 --redundancy=3
+
+  * Create a 100GB distributed volume which supports performance related volume options.
+      $ heketi-cli volume create --size=100 --durability=none --gluster-volume-options="performance.rda-cache-limit 10MB","performance.nl-cache-positive-entry no"
 `,
 	RunE: func(cmd *cobra.Command, args []string) error {
 		// Check volume size
@@ -153,6 +160,12 @@ var volumeCreateCommand = &cobra.Command{
 		req.Durability.Replicate.Replica = replica
 		req.Durability.Disperse.Data = disperseData
 		req.Durability.Disperse.Redundancy = redundancy
+
+
+		// Check volume options
+		if glusterVolumeOptions != "" {
+			req.GlusterVolumeOptions = strings.Split(glusterVolumeOptions, ",")
+		}
 
 		// Set group id if specified
 		if gid != 0 {
--- a/client/cli/go/cmds/volume.go
+++ b/client/cli/go/cmds/volume.go
@@ -146,21 +146,18 @@ var volumeCreateCommand = &cobra.Command{
 			return fmt.Errorf("Missing endpoint")
 		}
 
-		// Check clusters
-		var clusters_ []string
-		if clusters != "" {
-			clusters_ = strings.Split(clusters, ",")
-		}
-
 		// Create request blob
 		req := &api.VolumeCreateRequest{}
 		req.Size = size
-		req.Clusters = clusters_
 		req.Durability.Type = api.DurabilityType(durability)
 		req.Durability.Replicate.Replica = replica
 		req.Durability.Disperse.Data = disperseData
 		req.Durability.Disperse.Redundancy = redundancy
 
+		// Check clusters
+		if clusters != "" {
+			req.Clusters = strings.Split(clusters, ",")
+		}
 
 		// Check volume options
 		if glusterVolumeOptions != "" {
--- a/executors/sshexec/volume.go
+++ b/executors/sshexec/volume.go
@@ -62,6 +62,8 @@ func (s *SshExecutor) VolumeCreate(host string,
 
 	commands = append(commands, s.createAddBrickCommands(volume, inSet, inSet, maxPerSet)...)
 
+	commands = append(commands, s.createVolumeOptionsCommand(volume)...)
+
 	commands = append(commands, fmt.Sprintf("gluster --mode=script volume start %v", volume.Name))
 
 	_, err := s.RemoteExecutor.RemoteCommandExecute(host, commands, 10)
@@ -153,6 +155,23 @@ func (s *SshExecutor) VolumeDestroyCheck(host, volume string) error {
 	}
 
 	return nil
+}
+
+func (s *SshExecutor) createVolumeOptionsCommand(volume *executors.VolumeRequest) []string {
+	godbc.Require(len(volume.GlusterVolumeOptions) > 0)
+
+	commands := []string{}
+	var cmd string
+
+	// Go through all the Options and create volume set command
+	for _, volOption := range volume.GlusterVolumeOptions {
+		if volOption != "" {
+			cmd = fmt.Sprintf("gluster --mode=script volume set %v %v", volume.Name, volOption)
+			commands = append(commands, cmd)
+		}
+
+	}
+	return commands
 }
 
 func (s *SshExecutor) createAddBrickCommands(volume *executors.VolumeRequest,
--- a/pkg/glusterfs/api/types.go
+++ b/pkg/glusterfs/api/types.go
@@ -148,12 +148,13 @@ type VolumeDurabilityInfo struct {
 
 type VolumeCreateRequest struct {
 	// Size in GB
-	Size       int                  `json:"size"`
-	Clusters   []string             `json:"clusters,omitempty"`
-	Name       string               `json:"name"`
-	Durability VolumeDurabilityInfo `json:"durability,omitempty"`
-	Gid        int64                `json:"gid,omitempty"`
-	Snapshot   struct {
+	Size                 int                  `json:"size"`
+	Clusters             []string             `json:"clusters,omitempty"`
+	Name                 string               `json:"name"`
+	Durability           VolumeDurabilityInfo `json:"durability,omitempty"`
+	Gid                  int64                `json:"gid,omitempty"`
+	GlusterVolumeOptions []string             `json:"glustervolumeoptions,omitempty"`
+	Snapshot             struct {
 		Enable bool    `json:"enable"`
 		Factor float32 `json:"factor"`
 	} `json:"snapshot"`
--- a/tests/functional/TestSmokeTest/tests/heketi_test.go
+++ b/tests/functional/TestSmokeTest/tests/heketi_test.go
@@ -523,3 +523,36 @@ func TestHeketiVolumeExpandWithGid(t *testing.T) {
 	tests.Assert(t, err == nil, "Brick found with different Gid")
 
 }
+
+func TestHeketiVolumeCreateWithOptions(t *testing.T) {
+	// Setup the VM storage topology
+	teardownCluster(t)
+	setupCluster(t, 2, 2)
+	defer teardownCluster(t)
+
+	// Create a volume
+	volReq := &api.VolumeCreateRequest{}
+	volReq.Size = 10
+	volReq.Durability.Type = api.DurabilityReplicate
+	volReq.Durability.Replicate.Replica = 2
+	volReq.Snapshot.Enable = true
+	volReq.Snapshot.Factor = 1.5
+	volReq.GlusterVolumeOptions = []string{"performance.rda-cache-limit 10MB","performance.nl-cache-positive-entry no"}
+
+	// Set to the vagrant gid
+	volReq.Gid = 2333
+
+	// Create the volume
+	volInfo, err := heketi.VolumeCreate(volReq)
+	tests.Assert(t, err == nil)
+	tests.Assert(t, len(volInfo.GlusterVolumeOptions) > 0)
+
+	// SSH into system and check volume options.
+	vagrantexec := ssh.NewSshExecWithKeyFile(logger, "vagrant", "../config/insecure_private_key")
+	cmd := []string{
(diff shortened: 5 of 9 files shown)


humblec:
@lpabon @obnoxxx The tests are pending, however please share your thought about this design.

humblec:
@heketi/dev please review.

pkg/glusterfs/api/types.go
```diff
@@ -153,6 +153,7 @@ type VolumeCreateRequest struct {
 	Name       string               `json:"name"`
 	Durability VolumeDurabilityInfo `json:"durability,omitempty"`
 	Gid        int64                `json:"gid,omitempty"`
+	Options    []string             `json:"options,omitempty"`
```
lpabon: New type does not change the old API, this is ok.

executors/sshexec/volume.go
```diff
@@ -62,9 +62,16 @@ func (s *SshExecutor) VolumeCreate(host string,
 
 	commands = append(commands, s.createAddBrickCommands(volume, inSet, inSet, maxPerSet)...)
 
+	commands = append(commands, s.createVolumeOptionsCommand(volume)...)
+
 	commands = append(commands, fmt.Sprintf("gluster --mode=script volume start %v", volume.Name))
 
+	for _, c := range commands {
```
lpabon: Please remove this debugging for loop information

pkg/glusterfs/api/types.go
humblec: Yep.

executors/sshexec/volume.go
humblec: Sure.  Will address this in next patch set.

humblec:
Thanks @lpabon for the review.!

apps/glusterfs/app_volume.go
```diff
@@ -95,6 +95,18 @@ func (a *App) VolumeCreate(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
+	for _, volumeOption := range msg.Options {
+		switch volumeOption {
+		case "encryption":
```
humblec: @obnoxxx This design is based on the discussion happend here https://github.com/heketi/heketi/issues/711. If we dont maintain a whitelist, heketi is not in control to validate the options or modify accordingly. There may be options which are conflicting each other. The intelligence of parsing it and validating those scenarios can be done in heketi if we follow this path. User may not know those kind of scenarios and they may pass all at same time. Thats why I thought we can have a kind of validation in heketi. But I am ready to hear other thoughts.

obnoxxx: @humblec There is not much of a discussion about a whitelist in #711 actually. In fact the only argument to be read there is that it would be *bad* to maintain an additional wrapping layer in heketi, because it would require us to keep pace with gluster changes and options in heketi. (see @jarrpa 's comment https://github.com/heketi/heketi/issues/711#issuecomment-285668834).

I originally thought a higher level set of options that wrap a couple of gluster options would be good to have, and you said you wanted a whitelist of gluster options instead. Then José kind of convinced me that we should not add a complicated wrapping layer but mainly hand options through verbatim. But the current patch does actually not add a whitelist, but it adds a higher level option "encryption" which will translate into setting ```server.ssl``` and ```client.ssl``` to on.

I am not convinced yet that this is the way to go. I think there are several aspects to adding options:

* convenience and simplicity of use
* safety
* maintenance
* flexibility

Going with higher level wrapping and heketi-level checking will:

* make it convenient, simple to use
* will be safe to use in a sense that heketi prevents inconsistent setting
* will make maintenance much harder (gluster changes imply heketi changes)
* will be totally unflexible (need to do code changes for every new option we want to pass)

On the other hand side, just passing options through to gluster verbatim will:

* be less convenient to use (need to write out all options)
* require the admin to know more about gluster
* will be only slightly less safe to use (checking done by gluster)
* will require zero maintenance (gluster option changes do NOT require heketi changes)
* will be very flexible (using new options does NOT require heketi changes)

Based on these, I would now vote for doing the verbatim passing-through.
Minimal code change with maximum flexibility. KISS.

Of course we might think about adding some specific high-level / heketi-level
options like "encryption" in parallel. Just need to decide about the names and
about which takes precedence over what...

humblec: @obnoxxx fine.  I started  #171 by a thought of  having  free flow options from user to gluster via heketi, then by following other comments, I got a feel that we have to wrap it and develop a logic in heketi based on that.   'encryption' option was proposed in this PR by following this. I would say the current patch does have a whitelist, ie if the option does not fall into the known options we return with an error. I could see this logic as a whitelist. Whitelist or further case analysis ( using `switch`) was performed in heketi to bring  additional intelligence on these options and also to filter  different options provided by user,  for eg# if we already know there are issues in gluster when using using few options together, we can return with an error. But again, if everyone think we can do passthrough of options without wrapping or without any case analysis, I am fine with this. I will change the PR on this regard.

obnoxxx: @humblec, yes initially I felt we should do wrapping with higher-level heketi options, but then, with arguments by you and @jarrpa, and a recent real-world case that I ran across, I was slowly becoming convinced that a verbatim pass-through might be the best option.

And your current patch is not a simple whitelist, in the sense that it has a list of options which to pass through verbatim, but it is a higher level wrapping of some allowed combination of options (currently only the encryption ones). So it's more complicated than a simple whiltelist (adding convenience for the user but taking flexibility).

In the previous comment, I tried to list the pros and cons of the two approaches, and gave my preference, based on this newer perspective. I don't think anymore that we should have that additional intelligence in heketi. But I would like to hear other people's opinions before we change code again... @jarrpa implicitly voted for the simple pass-through - is that still valid? @humblec what do you think according to the analysis above? @lpabon @raghavendra-talur @MohamedAshiqrh .... any opinions?

jarrpa: I fully support a blanket pass-through of volume options with no whitelist. We should allow advanced users who know what they're doing to exercise their judgment. Then in `createVolumeOptionsCommand()` you can just generate a series of `gluster volume set` commands.

I will say I also like the idea of having convenience names for more common combinations of options. So for this I'd suggest still keeping the switch statement to watch for specific keywords, but have the default case be the direct pass-through.

humblec:
@lpabon @obnoxxx Please review at the earliest. Thanks !

client/cli/go/cmds/volume.go
```diff
@@ -145,10 +152,17 @@ var volumeCreateCommand = &cobra.Command{
 			clusters_ = strings.Split(clusters, ",")
 		}
 
+		// Check clusters
+		var volumeOptions_ []string
```
MohamedAshiqrh: Done.

MohamedAshiqrh: Thanks @jarrpa

executors/sshexec/volume.go
MohamedAshiqrh: done :)

apps/glusterfs/app_volume.go
MohamedAshiqrh: @humblec and  I have addressed this comment. check for the heketi-cli volume create -h. 
We can have our own list of strings(like "encryption") which chooses options related to string or they can pass the option directly. volume create is failed if they mentioned a wrong volume option. 
Added functional test.

executors/sshexec/volume.go
```diff
@@ -155,6 +158,33 @@ func (s *SshExecutor) VolumeDestroyCheck(host, volume string) error {
...
+			switch volOption {
+			case "encryption":
+				cmd = fmt.Sprintf("gluster --mode=script volume set %v server.ssl on", volume.Name)
+				commands = append(commands, cmd)
+
+				cmd = fmt.Sprintf("gluster --mode=script volume set %v client.ssl on", volume.Name)
+				commands = append(commands, cmd)
+			default:
+				cmd = fmt.Sprintf("gluster --mode=script volume set %v %v", volume.Name, volOption)
+				commands = append(commands, cmd)
```
obnoxxx: Thanks for adding the pass-through, as discussed!

executors/sshexec/volume.go
```diff
@@ -155,6 +158,33 @@ func (s *SshExecutor) VolumeDestroyCheck(host, volume string) error {
...
+
+	commands := []string{}
+	var cmd string
+
+	// Go through all the Options and create volume set command
+	for _, volOption := range volume.Options {
+		if volOption != "" {
+			switch volOption {
+			case "encryption":
+				cmd = fmt.Sprintf("gluster --mode=script volume set %v server.ssl on", volume.Name)
```
obnoxxx: I am not 100% comfortable with adding the wrapping heketi-specific options through the same command-line option as the pass-through options. Isn't it dangerous? What if gluster adds an "encryption" option that means an entirely different thing?

I.e. if we add heketi-specific wrapping-options, should we enable them either with a specific cmdline option like "--encryption" or else add two cmline parameters, one (e.g. --options) for the wrapping options and one (e.g. --gluster-options) for the pass-through options? I know it makes things more complicated, but seems cleaner.

This is not a veto (yet), still thinking about what would be the cleanest approach, like to be convinced...

@jarrpa @raghavendra-talur ... what do you think?

tests/functional/TestSmokeTest/tests/heketi_test.go
```diff
@@ -523,3 +523,36 @@ func TestHeketiVolumeExpandWithGid(t *testing.T) {
 	tests.Assert(t, err == nil, "Brick found with different Gid")
 
 }
+
+func TestHeketiVolumeCreateWithOptions(t *testing.T) {
```
obnoxxx: Thanks for adding a test!

executors/sshexec/volume.go
humblec: @obnoxxx can we discuss about the options later and do more enhancements with a follow up patch ?

obnoxxx: @humblec my comments were not about future enhancements, but about reasons why I would be reluctant to merge this patch as is.

I think we need to discuss the options (the names of the cmdline options, and whether to trigger the heketi-level options from the same heketi-cli cmdline option as the pass-through ones or from a different one) now, before merging anything, because otherwise we might create non-backward-compatible updates in the future.

I'd  be happy to merge a patch that does only the passing-through for now, and later, after design agreement has been reached, add the heketi-level convenience-wrappers.

But also for the passing-through we need to stop briefly and think about the naming. "--options" just seems to be too ambiguous. We should use something like --volume-options or --gluster-volume-options or --passthrough-options or so.

And the splitting up I'd still request. Now it would be 2 patches:
1. the prep patch with the clusters variable
2. the passthrough patch with the agreed name (currently proposed as --options)

Cheers - Michael

raghavendra-talur:
IMO,
1. Adding intelligence of so many options that gluster has, into heketi is a bad idea. I fully support the pass through option.
2. I don't like overloading of --options either, heketi interpreting some values for it and not interpreting is a bad idea. It will cause problems.
3. If heketi has some gluster option that it wants to interpret, it should come in as a new key, like we have durability type in volume create.

There is one thing I haven't thought about yet, which is cluster wide options. They are unfortunately overloaded in gluster volume set and it is dangerous to let them pass through in heketi.

humblec:
@obnoxxx comments are addressed. No more wrapping enabled in this patch. The options string has been changed to `--gluster-volume-options`. Please review at the earliest.

executors/sshexec/volume.go
```diff
@@ -155,6 +158,26 @@ func (s *SshExecutor) VolumeDestroyCheck(host, volume string) error {
...
+
+	godbc.Require(len(volume.GlusterVolOptions) > 0)
+
+	commands := []string{}
+	var cmd string
+
+	// Go through all the Options and create volume set command
+	for _, volOption := range volume.GlusterVolOptions {
+		if volOption != "" {
+
```
obnoxxx: extra empty line

executors/sshexec/volume.go
```diff
@@ -155,6 +158,26 @@ func (s *SshExecutor) VolumeDestroyCheck(host, volume string) error {
 	return nil
 }
 
+func (s *SshExecutor) createVolumeOptionsCommand(volume *executors.VolumeRequest) []string {
+
```
obnoxxx: extra empty line

executors/sshexec/volume.go
```diff
@@ -155,6 +158,26 @@ func (s *SshExecutor) VolumeDestroyCheck(host, volume string) error {
...
+	var cmd string
+
+	// Go through all the Options and create volume set command
+	for _, volOption := range volume.GlusterVolOptions {
+		if volOption != "" {
+
+			cmd = fmt.Sprintf("gluster --mode=script volume set %v %v", volume.Name, volOption)
+			commands = append(commands, cmd)
+		}
+
```
obnoxxx: extra empty line

executors/sshexec/volume.go
```diff
@@ -62,9 +62,12 @@ func (s *SshExecutor) VolumeCreate(host string,
 
 	commands = append(commands, s.createAddBrickCommands(volume, inSet, inSet, maxPerSet)...)
 
+	commands = append(commands, s.createVolumeOptionsCommand(volume)...)
+
 	commands = append(commands, fmt.Sprintf("gluster --mode=script volume start %v", volume.Name))
 
 	_, err := s.RemoteExecutor.RemoteCommandExecute(host, commands, 10)
+
```
obnoxxx: extra empty line.
also, this hunk has nothing to do with the change of this commit.

client/cli/go/cmds/volume.go
```diff
@@ -139,21 +146,24 @@ var volumeCreateCommand = &cobra.Command{
 			return fmt.Errorf("Missing endpoint")
 		}
 
-		// Check clusters
-		var clusters_ []string
-		if clusters != "" {
-			clusters_ = strings.Split(clusters, ",")
-		}
-
```
obnoxxx: I still would have liked to see this cleanup as part of a preliminary / preparatory patch, because it is a code cleanup and not directly related to the contents of this commit.

apps/glusterfs/volume_entry_create.go
```diff
@@ -11,11 +11,12 @@ package glusterfs
 
 import (
 	"fmt"
+	"strings"
+
 	"github.com/boltdb/bolt"
 	"github.com/heketi/heketi/executors"
 	"github.com/heketi/heketi/pkg/utils"
 	"github.com/lpabon/godbc"
-	"strings"
```
obnoxxx: this reordering of lines (due to gofmt ???) is independent of the change of this commit and should not be sneaked in. might be a separate commit (in the same PR if you like, or in a different PR).

client/cli/go/cmds/volume.go
```diff
@@ -23,20 +23,21 @@ import (
...
+	redundancy           int
+	gid                  int64
+	snapshotFactor       float64
+	clusters             string
+	expandSize           int
+	id                   string
+	kubePvFile           string
+	kubePvEndpoint       string
+	kubePv               bool
+	glusterVolumeOptions string
```
obnoxxx: Thanks for choosing a longer, more descriptive name for the options variable.
I still see a slight inconsistency: in the cli it's called glusterVolumeOptions / --gluster-volume-options, while in the api it's glusterVolOptions. Why that difference? Shouldn't we make them read the same, according to the principle of least surprise? (At least for the outside facing names cli and api.)

executors/sshexec/volume.go
humblec: Removed.

executors/sshexec/volume.go
humblec: Removed.

executors/sshexec/volume.go
humblec: Removed.

client/cli/go/cmds/volume.go
humblec: Yes, made it consistent ie `glustervolumeoptions`

humblec:
Splitting the commits into 3.

apps/glusterfs/volume_entry_create.go
humblec: Yeah, its goformat string.

humblec:
@obnoxxx please review:)

client/cli/go/cmds/volume.go
obnoxxx: thanks!

obnoxxx:
Thanks @humblec for splitting the patches and unifying the names!
This looks good to me now.
I think @lpabon 's requests have all been addressed, but we should give him an opportunity to have a final look. I'll try to reach him...

humblec:
Thanks all !

obnoxxx:
@lpabon it would be really great to get your (hopefully final) review, because we just now still have the chance of landing a patch in kubernetes to make use of this new API in time for 1.7 ... :-)

humblec:
@lpabon Can you please review this PR? :)

lpabon:
Everything is looking good. Just missing two unit tests. One at app_volume_test.go to test the API receiving the new option, and another in volume_entry_test.go to check that the entry in the database saves if the information (if it is to be saved.)

lpabon:
FYI, I also really agree with @raghavendra-talur 's comment:

>IMO,
>
>Adding intelligence of so many options that gluster has, into heketi is a bad idea. I fully support the pass through option.
>I don't like overloading of --options either, heketi interpreting some values for it and not interpreting is a bad idea. It will cause problems.
>If heketi has some gluster option that it wants to interpret, it should come in as a new key, like we have durability type in volume create.
>There is one thing I haven't thought about yet, which is cluster wide options. They are unfortunately overloaded in gluster volume set and it is dangerous to let them pass through in heketi.

We need to be careful how many "dials" we add to the API. We should always keep it as simple as possible.

obnoxxx:
@lpabon fair point about the unit tests. We should add them.

lpabon:
@humblec Do you think you can add unit tests in the next few weeks?

humblec:
@lpabon Definitely yes! Thanks for review. `One at app_volume_test.go to test the API receiving the new option,` -> This is already covered here though https://github.com/heketi/heketi/pull/751/files#diff-d8ea7c84c6f0d484faec4dcc9bea4f67R548

lpabon:
Closes #707

humblec:
@lpabon @obnoxxx requested unit tests are added . Please review 👍

obnoxxx:
I think the code looks good now.
Just one small nitpick:
The last patch that adds the unit tests, also makes one small fix to apps/glusterfs/volume_entry.go .
This should probably go into the first commit.

joejulian:
:tada:

MohamedAshiqrh:
\o/
nice :+1:

humblec:
\\o// 
Awesome!! Thanks all !!!

maikotz:
Nice work, thanks! 

Is there a way to set default options in the heketi json config? 
Would be nice to have for k8s storage auto provisioning.

Issue #748, opened 2017-04-19
the readme.md in extras/kubernetes/ is  not correspond to real


4admin2root:
for example 
there is no file named gluster-daemonset.json
and where is deploy-heketi-deployment.json

MohamedAshiqrh:
@4admin2root hey we have a [repo](https://github.com/gluster/gluster-kubernetes) which is more preferable for the kubernetes and openshift deployments as we have mentioned in the heketi [wiki](https://github.com/heketi/heketi/wiki/Kubernetes-Integration#overview).

lpabon:
Duplicate of #755

PR #762, opened 2017-05-16
Update README.md to point to gluster-kubernetes.

MohamedAshiqrh:
Point to gluster-kubernetes for the artifacts of kubernetes and openshift which is required for deployment.

Closes: #748

### DIFF
--- a/extras/kubernetes/README.md
+++ b/extras/kubernetes/README.md
@@ -1,7 +1,7 @@
 # Overview
 Kubernetes templates for Heketi and Gluster. The following documentation is setup
 to deploy the containers in Kubernetes.  It is not a full setup.  For full
-documentation, please visit the Heketi wiki page.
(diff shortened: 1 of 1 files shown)


obnoxxx:
@MohamedAshiqrh this does not fully fix #748. At least PR #765 is needed as well. But it seems that the whole document was rendered inconsistent in the last few updates and needs to be overhauled.

MohamedAshiqrh:
@obnoxxx agreed. We need #765 too. So how can we do that. We can merge #765 then we can add more fixes here and make the doc up to date?

obnoxxx:
@MohamedAshiqrh neither of the patches can be merged as is, imho, because they do fix or improve individual lines but still leave the whole document inconsistent. I was asking for a larger overhaul / reword of the text. This could be done in an update to this (or the other) PR or simply as a new PR.

lpabon:
@MohamedAshiqrh any status?

lpabon:
It would be great to also have all the possible environment variables documented.

lpabon:
I have see no updates to this PR. I am moving it out of Release 5.

PR #767, opened 2017-05-23
Revert command typos 

andersla:
Revert cp command typos that were introduced in commit c8cb940c285f2c225d12405870587eeb61916754
The redirect symbol > in the copy command makes the script fail and breaks the dockerimage: heketi/heketi:dev
This pull request fixes this error.


obnoxxx:
Thanks for the patch and for catching this this quickly...
And shame on me for insufficient review... Ouch!
We should have tests for this...

LGTM

Issue #455, opened 2016-08-12
Godep restore fails ..


humblec:
```
# cd .; git clone https://code.google.com/p/go-decimal-inf.exp /home/user/gocode/work/src/speter.net/go/exp/math/dec/inf
Cloning into '/home/user/gocode/work/src/speter.net/go/exp/math/dec/inf'...
fatal: repository 'https://code.google.com/p/go-decimal-inf.exp/' not found
godep: error downloading dep (speter.net/go/exp/math/dec/inf): exit status 128
godep: Error downloading some deps. Aborting restore and check.
```

humblec:
@lpabon have u seen this failure recently? Looks like the repo has been moved.

lpabon:
Do you need to run godep restore?  OpenShift's godep uses repo commits from its own repos with names for generic repos.  For example, kubernetes commit hashes for the hashes for github.com/openshift/kubernetes, even though they use them for k8s.io/kubernetes

humblec:
@lpabon I was trying to point out the godep restore failure especially with the https://code.google.com/p/go-decimal-inf.exp repo.  Eventhough I dont really want this, it should work seemlessly, Isnt it ?.

lpabon:
I agree it should if openshift did the right thing, but they don't.  Checkout https://github.com/openshift/origin/blob/master/hack/godep-save.sh

lpabon:
Maybe we could create a `hack/godep-save.sh` just like them.

lpabon:
Here is another: https://github.com/openshift/origin/blob/master/hack/godep-restore.sh

humblec:
@lpabon yeah, may be a hack/godep-save.sh is required in heketi. Because now heketi's dependency chain (heketi->openshift->k8s..etc)  is really heavy and difficult to manage :).

ramkrsna:
Do we have to check the dependency from Heketi -> OpenShift -> Kubernetes, or just the openshift < - > Heketi would suffice ?

lpabon:
@ramkrsna openshift to Heketi is enough, because openshift has a dep on Kube.

ramkrsna:
@heketi/dev @obnoxxx Could we explore the option of vendorizing heketi . i.e  As of Go 1.6, it will first look at the dependancies in the .vendor folder, if it does not find dependencies  in there, then it will look for the source directories in the GOPATH

lpabon:
[...]

PR #769, opened 2017-05-24
Move to use glide instead of godeps

### LINKED ISSUE #630: Use glide instead of Godeps

lpabon:
This is required because we needed to update the Kubernetes
APIs to use 1.6+. The biggest issue is that Kuberentes pulled
a lot of the APIs out of the tree and moved them to two different
repos: apimachinery, and client-go. This would have been a nightmare
to do in GoDeps, so I finally decided to do this in glide.

This will enable a new change for development, allowing '--kubeconfig'
to be passed to Heketi, enabling it to connect to a Kuberentes cluster
from the developers laptop.

I'm no git expert, so I couldn't figure out how to make the hated
vendor directory go into its own git commit, sorry for the pain.

Closes #630


lpabon:
Closes #455

executors/kubeexec/kubeexec_test.go
```diff
@@ -13,7 +13,7 @@ import (
 	"os"
 	"testing"
 
-	"k8s.io/kubernetes/pkg/client/restclient"
+	restclient "k8s.io/client-go/rest"
```
MohamedAshiqrh: nice.

extras/docker/fromsource/Dockerfile
```diff
@@ -11,6 +11,8 @@ ENV GOPATH=$BUILD_HOME/golang
 ENV PATH=$GOPATH/bin:$PATH
 ENV HEKETI_BRANCH="master"
 
+RUN curl https://glide.sh/get | sh
+
 # install dependencies, build and cleanup
 RUN mkdir $BUILD_HOME $GOPATH && \
```
MohamedAshiqrh: This is not a big change but we can reduce a layer.

RUN curl https://glide.sh/get | sh && \
mkdir $BUILD_HOME $GOPATH && \ 
.
.
.

MohamedAshiqrh:
This is good. I verified this and works as expected for me. I deleted the vendor directory and then did glide install -v. it was showing some packages missing before that.

obnoxxx:
First comments:

1. It is indeed good to see checked-in vendor/ go.
2. This is not documented. I had to look into travis.yml to find out that I had to do "dnf install glide" followed by "glide install -v" in order to be able to run make... ==> it needs to be documented. 
3. The fact that make does not work out of the box is unfortunate. I modified the Makefile to run glide install before running go build, but we should do it more cleverly, because this way, the time for make is longer. Can we detect whether glide install has been run? Is there a way to update the "glide install"?
4. I also experimented to have make use "go build -i" instead of "go build" to reduce the time needed for "make". Running "glide install -v" prior to "go make -i" greatly reduced the usefulness of "-i". :-(

lpabon:
@obnoxxx Great points. I'll make those changes.

raghavendra-talur:
@obnoxxx I don't understand your comment "It is indeed good to see checked-in vendor/ go."

@lpabon Changes look good.

obnoxxx:
@raghavendra-talur i wanted to express that I don't like the checked-in vendor directory. And I like the fact that this patch removes it.

lpabon:
@obnoxxx @raghavendra-talur @MohamedAshiqrh It works now just by typing `make`. I changed the Travis instructions to use this method from now on.

obnoxxx:
@lpabon looks good.

The only thing I don't like in the makefile is the download of glide.
I don't like flooding of the system with manual dowloads for something
that can also be installed via system packages (rpms / debs / ...).

Apart from that: Nice!

obnoxxx:
I'm inclined to merge this PR as is, and think about a possible follow-up patch for the makefile to address my above comment. I like to have this in rather now than late. :-)

obnoxxx:
LGTM

PR #779, opened 2017-06-13
change failed to removed

raghavendra-talur:

### DIFF
--- a/pkg/glusterfs/api/types.go
+++ b/pkg/glusterfs/api/types.go
@@ -28,7 +28,7 @@ const (
 	EntryStateUnknown EntryState = ""
 	EntryStateOnline  EntryState = "online"
 	EntryStateOffline EntryState = "offline"
-	EntryStateFailed  EntryState = "failed"
+	EntryStateFailed  EntryState = "removed"
 )
 
 type DurabilityType string


pkg/glusterfs/api/types.go
```diff
@@ -28,7 +28,7 @@ const (
 	EntryStateUnknown EntryState = ""
 	EntryStateOnline  EntryState = "online"
 	EntryStateOffline EntryState = "offline"
-	EntryStateFailed  EntryState = "failed"
+	EntryStateFailed  EntryState = "removed"
```
lpabon: Please also change the variable throughout the code `EntryStateFailed` to `EntryStateRemoved` for completeness

raghavendra-talur: @lpabon I realised now that EntryStateFailed is a api. Do you really want to change it? It will break backward compatibility.

humblec:
@heketi/dev  Please review

pkg/glusterfs/api/types.go
lpabon: True. I think for that reason, this PR cannot be changed. (API: https://github.com/heketi/heketi/blob/master/pkg/glusterfs/api/types.go#L31 ).

Instead we can always change the string shown in the CLI

lpabon:
Closing this PR since it would change the existing API.

PR #780, opened 2017-06-13
db dump

raghavendra-talur:
this dumps the whole db as json and also provides a way to create a new db using a json input


lpabon:
We discussed at the last meeting how this is a good idea, but needs to be implemented as a new tool accessing the database directly.

PR #781, opened 2017-06-14
make: Do not automatically install glide.

lpabon:
Instead ask the user to install it using their preferred method.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,10 +57,13 @@ heketi: glide.lock vendor
 
 server: heketi
 
-vendor: glide.lock
+vendor:
 ifndef GLIDEPATH
-	echo "Installing Glide"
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -57,10 +57,13 @@ heketi: glide.lock vendor
 
 server: heketi
 
-vendor: glide.lock
```
obnoxxx: Why the removal of glide.lock from the deps of vendor?

obnoxxx:
@lpabon process-wise, it is kind of weird that you just merged the patch immediately after filing the PR without waiting for review. (Not that we have a formal rule (yet) for that though.) The patch as such looks fine, except for the one question posted above, which you can certainly clarify. :-)

PR #784, opened 2017-06-19
Ignore vendor directory from git status

MohamedAshiqrh:
We can ignore vendor directory as glide install will take care of the directory.


obnoxxx:
LGTM

PR #785, opened 2017-06-19
clean Makefile and avoid skipping heketi building

MohamedAshiqrh:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,10 +52,10 @@ name:
 package:
 	@echo $(PACKAGE)
 
-heketi: glide.lock vendor
-	go build $(LDFLAGS) -o $(APP_NAME)
+heketi: server
 
-server: heketi
+server: glide.lock vendor
+	go build $(LDFLAGS) -o $(APP_NAME)
 
 vendor:
 ifndef GLIDEPATH


MohamedAshiqrh:
@heketi/maintainers PTAL

@obnoxxx This will help to build faster, can we merge this to avoid build delay.

MohamedAshiqrh:
@raghavendra-talur can you review?

MohamedAshiqrh:
@lpabon PTAL

obnoxxx:
@MohamedAshiqrh thanks. I have to say that I am doubtful about this patch. Will follow up later...

lpabon:
LGTM. Seems to do the same. Travis confirmed it works. I'll let @obnoxxx comment.

obnoxxx:
I said I'm skeptical about the patch because it seems conceptually weird. It just exchanges the roles of server and heketi in the Makefile. Why does it work? It works because "server" is a .PHONY target, hence it gets executed always, in contrast to "heketi" which is only executed when the the heketi file is older than the deps. In a real Makefile/buildsystem, there would be dependencies to all files needed for the build (i.e. in particular the source files, and not only glide.lock and vendor). But here this does not exist.

By changing the dependency and rule roles, so that heketi depends on the .PHONY server target, heketi gets always built. This is not wrong as such. But idiomatically, it "feels wrong" to have the "meta target" server depend on the concrete file target heketi. In (the theoretical) case we would want to add more server components to the server target again, we'd need to revise this patch again.

So I currently see one conceptually more satisfying alternative solution:

* make the heketi target .PHONY ==> Proposed in PR #788

Since we don't have dependency tracking for heketi sources, this is only
fair. And it achieves the same effect that is patch achieves implicitly.

On top of that, we might think about removing the "server" target altogether.

lpabon:
@obnoxxx good points. I'll let you and @MohamedAshiqrh decide what to do.

MohamedAshiqrh:
Closing this as #788 fixes this in right way. @obnoxxx thanks for the explanation.

PR #787, opened 2017-06-19
Allow disperse to be 2+1

fascinatedcow:
I was getting `Error: Invalid dispersion combination: 2+1`, but after checking the gluster docs and testing it out, I found out this combination is actually valid and is the one I would like to use in my 3-node cluster.

### DIFF
--- a/apps/glusterfs/app_volume.go
+++ b/apps/glusterfs/app_volume.go
@@ -83,6 +83,7 @@ func (a *App) VolumeCreate(w http.ResponseWriter, r *http.Request) {
 		d := msg.Durability.Disperse
 		// Place here correct combinations
 		switch {
+		case d.Data == 2 && d.Redundancy == 1:
 		case d.Data == 4 && d.Redundancy == 2:
 		case d.Data == 8 && d.Redundancy == 3:
 		case d.Data == 8 && d.Redundancy == 4:


lpabon:
@heketi/dev Is this a legal configuration for EC?

fascinatedcow:
Hi Luis,

Yep. Check here: https://gluster.readthedocs.io/en/latest/Administrator%20Guide/Setting%20Up%20Volumes/#creating-dispersed-volumes

> It's important to note that a configuration with 3 bricks and redundancy 1 will have less usable space (66.7% of the total physical space) than a configuration with 10 bricks and redundancy 1 (90%).

That example implies that 3 bricks are possible.

And:

> The implementation of erasure codes in GlusterFS limits the redundancy to a value smaller than #Bricks / 2 (or equivalently, redundancy * 2 < #Bricks).

1 * 2 < 3, so we're still within this boundary.

Furthermore, it works and doesn't complain about it not being an optimal configuration.

Actually, thinking about it there are many other possible configurations that aren't supported by heketi due to the way the validation is done. What if you want more than 8 bricks? Is there a particular reason for limiting it to only 3 possible configurations?

Let me re-submit with what I think are the limits (need to go and research it a bit first) and we'll go from there?

Kind regards,

Matt

MohamedAshiqrh:
@lpabon yeah, 2+1 is a valid EC configuration. Confirmed with Ashish who works on EC in gluster.

@fascinatedcow thanks for the PR.

lpabon:
@fascinatedcow @MohamedAshiqrh Thanks!

PR #788, opened 2017-06-20
Build: make sure heketi is always built.

obnoxxx:
The heketi sources are not tracked in the Makefile's
dependencies. So only by making the heketi target
phony we can ensure that heketi gets rebuilt when
any sources have changed.


obnoxxx:
This is an alternative to PR #785 . I think it is conceptually more correct. @lpabon @MohamedAshiqrh

obnoxxx:
thx @MohamedAshiqrh

PR #792, opened 2017-06-23
python: set heketi log level to "debug" in tests

obnoxxx:

### DIFF
--- a/client/api/python/test/unit/heketi.json
+++ b/client/api/python/test/unit/heketi.json
@@ -34,6 +34,8 @@
       "user": "sshuser"
     },
 
+    "loglevel" : "debug",
+
     "_db_comment": "Database file name",
     "db": "heketi.db"
   }


client/api/python/test/unit/heketi.json
```diff
@@ -34,6 +34,8 @@
       "user": "sshuser"
     },
 
+    "loglevel" : "debug",
```
raghavendra-talur: This is something which I am not sure about:
1. it is good to have debug logs ready when a test fails
2. it might however mask some race conditions as debug logs add enough delay to operations.

I am a little inclined towards default log level in tests too. Let developers run in debug level the second time when a test fails.

@heketi/dev what do others think?

lpabon:
I'm I wrong or is debug already the default?

obnoxxx:
@lpabon hmmm, that is actually an interesting question. From looking at the code, it seems that there is no compiled-in default. But the *default* config file (etc/heketi.json) has the default loglevel as debug. It make me feel uncomfortable when there are no compiled-in defaults. Will need to check the code for other config defaults.

lpabon:
retest this please

obnoxxx:
@raghavendra-talur should we close this?

obnoxxx:
Clarified it. The default log level in the app is INFO, it's initialized here:

https://github.com/heketi/heketi/blob/master/apps/glusterfs/app.go#L43

PR #795, opened 2017-06-26
fix/go-cli: set default HEKETI_CLI_SERVER

jsvisa:
if `HEKETI_CLI_SERVER` is not set, make `http://localhost:8080` as the default one

### DIFF
--- a/client/cli/go/cmds/root.go
+++ b/client/cli/go/cmds/root.go
@@ -36,11 +36,10 @@ type Options struct {
 }
 
 var RootCmd = &cobra.Command{
-	Use:   "heketi-cli",
-	Short: "Command line program for Heketi",
-	Long:  "Command line program for Heketi",
-	Example: `  $ export HEKETI_CLI_SERVER=http://localhost:8080
-  $ heketi-cli volume list`,
+	Use:     "heketi-cli",
+	Short:   "Command line program for Heketi",
+	Long:    "Command line program for Heketi",
+	Example: `  $ heketi-cli volume list`,
 	Run: func(cmd *cobra.Command, args []string) {
 		if version {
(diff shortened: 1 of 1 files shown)


lpabon:
@jsvisa  This looks great, but could you add this to the help screen also? Thanks!

bradleyfalzon:
Love this, but what about also logging the fact you're defaulting to `defaultCliServer` (and potentially how to change it)?

jsvisa:
@lpabon help message is updated.

jsvisa:
@bradleyfalzon the priority is `command arg -s` > `HEKETI_CLI_SERVER` > `defaultCliServer`

bradleyfalzon:
Oh I meant from the UX point of view, if I run multiple Heketi servers for different clusters and I do a `cluster list` without an exported server config, the user might not know which server they're connecting to.

```
[root@instance-1 ~]# heketi-cli cluster list
No server specified using http://localhost:8080
Clusters:
635e47ef8dfc1ab2c24b9390665d1ad8
```

Although since documenting the default server in help message, I retract my comment, I think that's enough.

jsvisa:
If no server or ENV is specified, the `http://localhost:8080` is the default one

bradleyfalzon:
Yes, but I meant the user didn't know that, but since documenting the default in the help message, it's clearer.

jsvisa:
IMO, this is redundancy, the user need to know which endpoint was using, if not specified, the default one is using. @lpabon What's your opinion

bradleyfalzon:
Yes I agree it's now redundant now that it's being documented in the help message, sorry for the confusion.

lpabon:
I think this is clean, thanks! @jsvisa and @bradleyfalzon

PR #797, opened 2017-06-26
Fix/create without volume options

jsvisa:
`GlusterVolumeOptions` is an optional argument, if this assert is acquired, the default volume create manner will lead the heketi server to crash.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,8 @@ tags
 log
 *.sw?
 *~
+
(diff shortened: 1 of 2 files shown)


jsvisa:
And the CI runs the mock executor, so the sshexec can't reached forever, so the CI always success.

lpabon:
Thanks @jsvisa !

Issue #756, opened 2017-05-05
Device size not updated if disk is expanded


ae-dg:
When using VMs for the gluster nodes drive sizes can be changed. For example in RHEV it is possible to expand the drive without even restarting the virtual machine.

After increasing the size of the virtual disk and running pvresize on the virtual disk the volume group shows an increase in the free space.  However heketi doesn't notice and still shows the original size of the drives when requesting the device info. As a result it is not possible to create a volume that is larger than what heketi thinks is the size of the device even if the device actually has enough space.

Can a "device rescan" API call be created to get heketi to update it's internal information to match the actual disk size?

lpabon:
@ae-dg Wow, that would be cool, but that doesn't happen normally. Instead new disks, nodes, or clusters are normally added to Heketi to increase capacity.

If you feel you would like to add this support, PRs are welcomed :-D

hakanisaksson:
This feature is very much need for production clusters.

PR #805, opened 2017-07-11
Device resync API to update free disk space after increasing the size of disk

shiroari:
Hi. This PR adds a new REST API method `resync` that helps to update free disk space on device after increasing the size of virtual disk. See issue #756

Usage:
```
$ heketi-cli device resync id
```

### DIFF
--- a/apps/glusterfs/app_device.go
+++ b/apps/glusterfs/app_device.go
@@ -350,3 +350,97 @@ func (a *App) DeviceSetState(w http.ResponseWriter, r *http.Request) {
 		return "", nil
 	})
 }
+
+func (a *App) DeviceResync(w http.ResponseWriter, r *http.Request) {
+
+	vars := mux.Vars(r)
+	deviceId := vars["id"]
+
+	var (
+		device *DeviceEntry
+		node   *NodeEntry
+	)
+
+	// Get device info from DB
+	err := a.db.View(func(tx *bolt.Tx) error {
+		var err error
+		device, err = NewDeviceEntryFromId(tx, deviceId)
+		if err != nil {
+			return err
+		}
+		node, err = NewNodeEntryFromId(tx, device.NodeId)
+		if err != nil {
+			return err
+		}
+		return nil
+	})
+	if err == ErrNotFound {
+		http.Error(w, err.Error(), http.StatusNotFound)
+		return
+	} else if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		logger.Err(err)
+		return
+	}
+
+	logger.Info("Checking for device %v changes", deviceId)
+
+	// Check and update device in background
+	a.asyncManager.AsyncHttpRedirectFunc(w, r, func() (seeOtherUrl string, e error) {
+
+		// Get actual device info from manage host
+		info, err := a.executor.GetDeviceInfo(node.ManageHostName(), device.Info.Name, device.Info.Id)
+		if err != nil {
+			return "", err
+		}
+
+		// Note that method GetDeviceInfo returns the free disk space available for allocation.
+		// The free disk space is equal to the total disk space only if we haven't already
+		// allocated space, because every allocation decreases the free disk space returned
+		// by method GetDeviceInfo. In order to calculate a new total space we need to sum
+		// the free disk space and the space used by heketi.
+		if device.Info.Storage.Total == info.Size+device.Info.Storage.Used {
+			logger.Info("Device %v is up to date", device.Info.Id)
+			return "", nil
+		}
+
+		logger.Debug("Free space of '%v' (%v) has changed %v -> %v", device.Info.Name, device.Info.Id,
+			device.Info.Storage.Free, info.Size)
+
+		// Update device
+		err = a.db.Update(func(tx *bolt.Tx) error {
+
+			// Reload device in current transaction
+			device, err := NewDeviceEntryFromId(tx, deviceId)
+			if err != nil {
+				logger.Err(err)
+				return err
+			}
+
+			newFreeSize := info.Size
+			newTotalSize := newFreeSize + device.Info.Storage.Used
+
+			logger.Info("Updating device %v, total: %v -> %v, free: %v -> %v", device.Info.Name,
+				device.Info.Storage.Total, newTotalSize, device.Info.Storage.Free, newFreeSize)
+
+			device.Info.Storage.Total = newTotalSize
+			device.Info.Storage.Free = newFreeSize
+
+			// Save updated device
+			err = device.Save(tx)
+			if err != nil {
+				logger.Err(err)
+				return err
+			}
+
+			return nil
+		})
+		if err != nil {
+			return "", err
+		}
+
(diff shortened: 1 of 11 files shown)


MohamedAshiqrh:
I am talking about this 
```

# heketi-cli node info 4d5595370b425a02b3a5372fdcfc1c35
Node Id: 4d5595370b425a02b3a5372fdcfc1c35
State: online
Cluster Id: dea21f84a0f896d1b54cf487890174db
Zone: 2
Management Hostname: 172.19.0.3
Storage Hostname: 172.19.0.3
Devices:
Id:8f6a2dff8bd6281a3e9696f802a0b116   Name:/dev/loop7          State:online    Size (GiB):500     Used (GiB):497     Free (GiB):2       
Id:b8a7b13e415050f047fe055e1ee1ab9c   Name:/dev/loop5          State:online    Size (GiB):500     Used (GiB):498     Free (GiB):1       
Id:c79267464bd52ec842dc6a0bbf66de38   Name:/dev/loop6          State:online    Size (GiB):500     Used (GiB):499     Free (GiB):0 

# heketi-cli device resync 8f6a2dff8bd6281a3e9696f802a0b116
Device  updated

# heketi-cli node info 4d5595370b425a02b3a5372fdcfc1c35    
Node Id: 4d5595370b425a02b3a5372fdcfc1c35
State: online
Cluster Id: dea21f84a0f896d1b54cf487890174db
Zone: 2
Management Hostname: 172.19.0.3
Storage Hostname: 172.19.0.3
Devices:
Id:8f6a2dff8bd6281a3e9696f802a0b116   Name:/dev/loop7          State:online    Size (GiB):997     Used (GiB):995     Free (GiB):2       
Id:b8a7b13e415050f047fe055e1ee1ab9c   Name:/dev/loop5          State:online    Size (GiB):500     Used (GiB):498     Free (GiB):1       
Id:c79267464bd52ec842dc6a0bbf66de38   Name:/dev/loop6          State:online    Size (GiB):500     Used (GiB):499     Free (GiB):0     
```

shiroari:
Hi, @MohamedAshiqrh Thank you for review.
We are using this patch for heketi v3.1 on our clusters and I could not find any changes regarding disk size in v4, so it should be fine.

In your example why 'Used' size has been changed. Can you run it with DEBUG output?
Let's take your example and see how the disk size would be calculated. After increasing the disk size for example from 500Gb to 600Gb, the method `getVgSizeFromNode` returns 102Gb (newFree). The `resync` method changes total => 102Gb + 497Gb (newFree + oldUsed) and free => newFree (Used size never changed).

As result we have:
```
Size (GiB):599Gb     Used (GiB):497     Free (GiB):102
```
Looks pretty good I would say. 

But I agree that calculation looks a bit strange and I guess it because we expect that total size field contains total disk size but in reality (in heketi) is free disk size.

MohamedAshiqrh:
@shiroari This will work perfectly fine on sshexec and kubeexec. For Mockexec as it is returning [this](https://github.com/shiroari/heketi/blob/55c67ab9f22b5993401241477ac77443e6d061bd/executors/mockexec/mock.go#L49), Output is expected. I am not asking you to change anything. It works fine 👍. 

Thanks for the patch again.

raghavendra-talur:
@shiroari I am reviewing this. Will post my comments by tomorrow.

apps/glusterfs/app_device.go
```diff
@@ -350,3 +350,92 @@ func (a *App) DeviceSetState(w http.ResponseWriter, r *http.Request) {
...
+	// Check and update device in background
+	a.asyncManager.AsyncHttpRedirectFunc(w, r, func() (seeOtherUrl string, e error) {
+
+		// Get actual device info from manage host
+		info, err := a.executor.GetDeviceInfo(node.ManageHostName(), device.Info.Name, device.Info.Id)
+		if err != nil {
+			return "", err
+		}
+
+		if device.Info.Storage.Free == info.Size {
```
lpabon: If the free size is equal to the size of the device, then it is up to date. I'm confused on why use Free == Size, instead of Storage.Total == Size. Do you mind adding a comment on explanation?

apps/glusterfs/app_device.go
```diff
@@ -350,3 +350,92 @@ func (a *App) DeviceSetState(w http.ResponseWriter, r *http.Request) {
...
+
+			// Reload device in current transaction
+			device, err := NewDeviceEntryFromId(tx, deviceId)
+			if err != nil {
+				logger.Err(err)
+				return err
+			}
+
+			newFreeSize := info.Size
+			newTotalSize := newFreeSize + device.Info.Storage.Used
```
lpabon: What happens if the total size is smaller.. to say just a few blocks. How would heketi behave?

apps/glusterfs/app_device.go
shiroari: The confusion comes from `getVgSizeFromNode` function (source for Info.Size) that returns **free** disk size and not total disk size.

For instance we started heketi on disk 10Gb.
In this case `getVgSizeFromNode` returns **10Gb** and Total = 10Gb, Free = **10Gb**, Used = 0Gb.
If we allocate 2Gb then Total = 10Gb, Free = **8Gb**, Used = 2Gb and `getVgSizeFromNode` returns **8Gb**.

So the first point to keep in mind is that there is only one real value that we can use as baseline is Storage.Free. Maybe write that line as: 
```
device.Info.Storage.Total == info.Size + device.Info.Storage.Used
```
Will it make more sense? I admit that total size is more natural to comprehend and I was struggling with that line a lot.

apps/glusterfs/app_device.go
shiroari: I'm afraid I did not get it. As I mentioned above the free size is the only value we need to take into account. Total size is kind of fake. We reconstruct it from free disk size to show how much space we would get after releasing allocated space.

apps/glusterfs/app_device.go
lpabon: Cool, please comment this as an explanation.

apps/glusterfs/app_device.go
lpabon: I'm guessing the goal is to allow the increase of VG, which a sync would then adjust the database. What if someone shrinks it, to say 1G. What would happen? Should Heketi thrown an error saying that the device has lost data?

shiroari: The shrinking is totally fine and it is actually one of use cases. It could be useful even without any changes of the disk size. Let's consider that one copied a big file to the disk (directly by using filesystem) then heketi's disk free size is not correct anymore and resync will fix it. 
Total disk size in heketi in this case will be less then before, the free size will reflect actual free size left on the disk and the used size won't change.

apps/glusterfs/app_device.go
shiroari: Ok. Will do. Do you prefer single squashed commit or a new one on the top of current one?

shiroari: Done. I only added the comment and squashed commits.

lpabon: Squashed. Thanks for the PR!

Issue #825, opened 2017-08-03
any Trusted Storage Pool node can give the volfile.


MohamedAshiqrh:
As any Trusted Storage Pool node can give the volfile why should we just give the nodes on which the volume resides?

Instead we can give all the node in the specific cluster.

PR #826, opened 2017-08-03
TSP should be backup-volfile server

MohamedAshiqrh:
As any node in the TSP can give the volifile for the client.

closes: #825 

### DIFF
--- a/apps/glusterfs/volume_entry_create.go
+++ b/apps/glusterfs/volume_entry_create.go
@@ -41,9 +41,21 @@ func (v *VolumeEntry) createVolume(db *bolt.DB,
 
 	// Get all brick hosts
 	stringset := utils.NewStringSet()
-	for _, brick := range vr.Bricks {
-		stringset.Add(brick.Host)
-	}
+	err = db.View(func(tx *bolt.Tx) error {
+		cluster, err := NewClusterEntryFromId(tx, v.Info.Cluster)
+		if err != nil {
+			return err
+		}
+		for _, nodeId := range cluster.Info.Nodes {
+			node, err := NewNodeEntryFromId(tx, nodeId)
+			if err != nil {
+				return err
+			}
+			stringset.Add(node.StorageHostName())
+		}
+		return err
+	})
+
 	hosts := stringset.Strings()
 	v.Info.Mount.GlusterFS.Hosts = hosts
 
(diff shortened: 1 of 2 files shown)


MohamedAshiqrh:
@heketi/maintainers Can you please review?

obnoxxx:
@MohamedAshiqrh The patch looks clean to me, but if that can be done just like this, we need a unit test. :-)

obnoxxx:
@centos-ci ok to test

lpabon:
@MohamedAshiqrh 

I can see that it could work on line: https://github.com/heketi/heketi/blob/master/apps/glusterfs/volume_entry_create.go#L56 because volfiles can be taken from any host, but I'm not sure this would work on line: https://github.com/heketi/heketi/blob/master/apps/glusterfs/volume_entry_create.go#L52 ?

Maybe we can create a volfiles variable and a host one for mounting the volume? Does it matter? Is GlusterFS efficient enough to mount from one node which does not have the bricks, and be as fast as mounting from the node that does? Is mounting just getting the volfile, the the FUSE client actually reconnects to the actual node with the bricks?

lpabon:
retest this please

MohamedAshiqrh:
@lpabon Yes, It is getting the volfile and then mounting using the volfile. Volfile can be passed from any of the gluster node which belongs to the TSP. While mounting it will use the volfile to connect to (all) bricks doesn't matter which ip is mentioned to mount.

MohamedAshiqrh:
@obnoxxx @lpabon added the unit test.

obnoxxx:
@lpabon right, to make it even more clear, this is how the mount happens:  the mount proces just goes to some glusterd in the cluster. From this retrieves the volume information (volfile), in particular the brick server IPs, and then in the next step it opens connections to the brick servers. So the mount can be initiated to ANY node in the cluster.

Hence the mount can specify any ip of a gluster cluster node (tsp node).

obnoxxx:
Maybe we could make it more clear by not calling the one MointPoint and the other backup-volfile-servers, but maybe calling the backup-volfile-servers backup-mount-ips or so... (in a separate patch)

obnoxxx:
The patch LGTM.

lpabon:
LGTM

PR #831, opened 2017-08-07
Change failed to removed in cli

MohamedAshiqrh:
Signed-off-by: mliyazud <mliyazud@redhat.com>

### DIFF
--- a/client/cli/go/cmds/node.go
+++ b/client/cli/go/cmds/node.go
@@ -265,6 +265,10 @@ var nodeInfoCommand = &cobra.Command{
 		if err != nil {
 			return err
 		}
+		var entryStateRemoved api.EntryState = "removed"
+		if info.State == api.EntryStateFailed {
+			info.State = entryStateRemoved
+		}
 
 		if options.Json {
 			data, err := json.Marshal(info)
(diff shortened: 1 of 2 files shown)


MohamedAshiqrh:
@heketi/maintainers This is PR is result of discussion in https://github.com/heketi/heketi/pull/779

MohamedAshiqrh:
This PR creates a temp variable for removed and replaces the failed to removed.

MohamedAshiqrh:
@lpabon @obnoxxx PTAL

client/cli/go/cmds/node.go
```diff
@@ -265,6 +265,10 @@ var nodeInfoCommand = &cobra.Command{
 		if err != nil {
 			return err
 		}
+		const entryStateRemoved api.EntryState = "removed"
+		if info.State == api.EntryStateFailed {
+			info.State = entryStateRemoved
```
lpabon: This is good, but quick question? Why create a const? Just in case it is used in the future?

MohamedAshiqrh: Done 👍 :)

MohamedAshiqrh:
@lpabon PTAL

lpabon:
Sorry, that is not what I meant. I meant just use:

```go
info.State = "failed"
```

But it is ok, I'll just merge it and fix that later.

PR #833, opened 2017-08-12
tests: Kubernetes functional tests

lpabon:
This functiontal tests creates a three node full Kubernetes cluster
based on CoreOS. Each of the nodes has a single 32GiB drive that
is given to GlusterFS container.

The tests use the following technologies:

* https://github.com/lpabon/kubeup - Sets up the three node cluster
    * Three nodes are created: node1, node2, and node3
* registry.yaml - Sets up a Docker registry and proxies. Docker does
  not allow non-HTTPS access to external registries, but only to
  localhost. The proxies fool the Docker daemon, instead moving
  the container to one of the nodes in the cluster. When an image is
  pulled from "localhost:5000/<image>", then the local proxy gets
  the image from the real registry on node1.
* These tests are focused on running on the CI system. They may (and
  probably will be able to) run on your system, but that is a non-goal.
* New Kubernetes specific tests can be added. See
  TestKubeSmokeTest/TestDynamicProvisioning/... for more information
* [Quartermaster](https://github.com/coreos/quartermaster) is used to deploy
  GlusterFS and Heketi. It takes care of all setup and provides back status
  of the cluster.


obnoxxx:
@lpabon that looks very interesting!

I need to look more deeply, but a few comments upfront:

* Are you aware that we do already have a full 3-node vagrant-based kubernetes test environment in https://github.com/gluster/gluster-kubernetes ?
* Some caching improvements with local registry are currently being reviewed: https://github.com/gluster/gluster-kubernetes/pull/287
* What puzzles me is the statement that running the tests on the developer's machine is a non-goal. This seems to be reverse. The flow should imho be to create test suites that can run locally, and then, once they have matured, have them run in the CI. I would make it a requirement that the tests run locally.
* Finally, as a very spontaneous reaction, I would somewhat prefer to use the technologies directly related to and created by this project, e.g. the templates and mechanisms provided in gluster-kubernetes, but of course this can all be discussed... :-)

lpabon:
retest this please

lpabon:
@obnoxxx I think you are right. I'll make sure it can be run on your system (as long as it is Fedora/CentOS). Also, I would really love another test using gluster-kubernetes.

jarrpa:
@lpabon Can't you just push a docker config with `INSECURE-REGISTRY="--insecure-rgistry <registry url>"`? I do that to run an insecure registry on my laptop for a Kube cluster in VMs and it works just fine.

lpabon:
@jarrpa setting that value is ok if we were deploying kubernetes by hand, but this is a real deployment tool like kubeadm. The advantage over kubeadm is the simplicity of install using matchbox.

jarrpa:
@lpabon I set that value while using kubeadm, so your argument doesn't make sense. :)

obnoxxx:
This is outdated. Closing for now. Can be reopened if s/o wants to continue on this.

PR #834, opened 2017-08-15
check for self heal before replaceBrick

obnoxxx:
The conditions to be considered are
1. If the brick to be replaced is source for some files, don't proceed
2. We kill the brick to be replaced, if that would make the replica set
lose quorum, don't proceed.

This has a side effect that remove device won't work on volumes which
have replica count 2. Choosing data safety over migration features.


obnoxxx:
This is a rebase of PR #718  and should replace that pr.

obnoxxx:
LGTM

obnoxxx:
According to @lpabon 's LGTM barring rebase and CI test pass, which happened here now, I am merging.

lpabon:
:+1:

PR #778, opened 2017-06-13
implement timeout for kubeexec

raghavendra-talur:
Signed-off-by: Raghavendra Talur <rtalur@redhat.com>

PR #840, opened 2017-08-23
Revert "implement timeout for kubeexec"

raghavendra-talur:
Reverts heketi/heketi#778

This option does not work as expected in all the cases and leads to kubeexec timeouts in some cases.

When heketi uses kubeexec to get the list of nodes with gluster labels it fails with timeout. I increased the value to 10000 from 10 and it failed again. Reading the comments on the PR which introduced this feature in kubernetes[1] tells that it needs a corresponding feature in container run time. I have not investigated exactly what change needs to be in container runtime and if it exists, however it is for sure that with timeout set, heketi is unable to perform a kubeexec operation at all.

[1] kubernetes/kubernetes#33366

### DIFF
--- a/executors/kubeexec/kubeexec.go
+++ b/executors/kubeexec/kubeexec.go
@@ -129,7 +129,6 @@ func NewKubeExecutor(config *KubeConfig) (*KubeExecutor, error) {
 
 	// Create a Kube client configuration
 	k.kubeConfig, err = inClusterConfig()
-	k.kubeConfig.Timeout = 10
 	if err != nil {
 		return nil, logger.LogError("Unable to create configuration for Kubernetes: %v", err)
 	}


obnoxxx:
@raghavendra-talur can you be a little more specific?
* what does that mean, "does not work as expected in all cases"
* leads to kubexec timeouts - sure, that's what it's for ;-) ... is the number too low?

humblec:
@raghavendra-talur Yeah, same question from my side as well. Is it just the timeout is low or not working at all ?

raghavendra-talur:
@obnoxxx @humblec when heketi uses kubeexec to get the list of nodes with gluster labels it fails with timeout. I increased the value to 10000 from 10 and it failed again. Reading the comments on the PR which introduced this feature in kubernetes[1] tells that it needs a corresponding feature in container run time. I have not investigated exactly what change needs to be in container runtime and if it exists, however it is for sure that with timeout set, heketi is unable to perform a kubeexec operation at all.


[1] https://github.com/kubernetes/kubernetes/pull/33366

obnoxxx:
@centos-ci retest this please

humblec:
By looking at the latter statement ie `it need runtime support`, this is risky unless tested heavily. @obnoxxx @lpabon I feel we have to revert and better not pull in before we get some confidence.

obnoxxx:
Yeah, seems like it.
At least we need to mention this in the commit message.
Strangely, the travis ci test failed, but in updating glide for the build, so not related as it seems...

raghavendra-talur:
Have updated the commit message and restarted the travis ci test.

obnoxxx:
@raghavendra-talur thanks. this would update the merge commit message. could you update the actual commit msg and push over?

obnoxxx:
lgtm. we need to do it differently

panigrahis:
@obnoxxx is this fix part of the latest docker image heketi/heketi:dev?
I am facing the exact same issue and would like to know if this fix is available.

raghavendra-talur:
@panigrahis this fix should be available in latest docker image, please update and try.

PR #845, opened 2017-09-10
golang: Make this fast

purpleidea:
As per https://ttboj.wordpress.com/2017/02/26/faster-golang-builds/

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,7 +53,7 @@ package:
 	@echo $(PACKAGE)
 
 heketi: vendor glide.lock
-	go build $(LDFLAGS) -o $(APP_NAME)
+	go build -i $(LDFLAGS) -o $(APP_NAME)
 
 server: heketi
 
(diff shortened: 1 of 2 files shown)


obnoxxx:
@purpleidea thanks for your patch.
I think i had wanted to add this "-i" to the build commands but had forgotten it again...
But on the (patch-)cosmetic side:

could you please reword the commit message? "golang: Make this fast" --> "build: make the compilation faster" or similar?

Makefile
```diff
@@ -69,13 +69,13 @@ endif
 	glide install -v
 
 	echo "Building dependencies to make builds faster"
-	go install github.com/heketi/heketi
+	#go install github.com/heketi/heketi
```
obnoxxx: 1. Why comment out and not remove?
2. AFAICT, this is not about making the build fast, but about removing something that you think is harmful. Could you please create a second patch for it?

Makefile
```diff
@@ -69,13 +69,13 @@ endif
...
 
 glide.lock: glide.yaml
 	echo "Glide.yaml has changed, updating glide.lock"
 	glide update -v
 	echo "Building dependencies to make builds faster"
-	go install github.com/heketi/heketi
+	#go install github.com/heketi/heketi
```
obnoxxx: dito

obnoxxx:
I still think that this should have been 2 patches, as I said in my requests...
But I will be slightly generous on this one.... ;-)

purpleidea:
@obnoxxx Oh, sorry-- I thought my improved commit message made it clear why I thought they were the same? If you want me to send a branch for a revert + two patches, LMK. More points that way.

obnoxxx:
@purpleidea no need to say sorry. It's ok like this. Let's move on. ;-)

PR #846, opened 2017-09-10
tests: Make tests safe for personal machines

purpleidea:
This is totally unsafe to run in an environment that I use. Okay, for
travis, but not here. This should fix it, since this is already in
travis.

### DIFF
--- a/tests/functional/run.sh
+++ b/tests/functional/run.sh
@@ -7,7 +7,9 @@ TESTS="TestManyBricksVolume $TESTS"
 
 # install glide
 if ! command -v glide ; then
-    curl https://glide.sh/get | sh
+	echo glide is not installed, please install it to continue
+	echo 'get it from your package manager, or unsafely via: "curl https://glide.sh/get | sh"'
+	exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


obnoxxx:
thanks for your patch, @purpleidea !

tests/functional/run.sh
```diff
@@ -7,7 +7,9 @@ TESTS="TestManyBricksVolume $TESTS"
 
 # install glide
 if ! command -v glide ; then
-    curl https://glide.sh/get | sh
+	echo glide is not installed
+	echo 'run "curl https://glide.sh/get | sh" if you are crazy'
```
obnoxxx: please use more appropriate language ;-)

In the main Makefile we already have:

> Please install glide.
> Install it using your package manager or
> by running: curl https://glide.sh/get | sh.

this may be not perfect but I think it's better.

purpleidea:
patch updated.

purpleidea:
(now rebased to master)

obnoxxx:
LGTM, thanks

purpleidea:
w00t! first patch

PR #850, opened 2017-09-15
apps: glusterfs: Populate the ring on load from disk

purpleidea:
This now populates the rings on load from disk. Previously it only did
this if devices existed. Run time adds still don't add to the ring yet.

### DIFF
--- a/apps/glusterfs/allocator_simple.go
+++ b/apps/glusterfs/allocator_simple.go
@@ -10,8 +10,9 @@
 package glusterfs
 
 import (
-	"github.com/boltdb/bolt"
 	"sync"
+
+	"github.com/boltdb/bolt"
 )
 
 // Simple allocator contains a map to rings of clusters
--- a/apps/glusterfs/allocator_simple.go
+++ b/apps/glusterfs/allocator_simple.go
@@ -45,6 +45,11 @@ func NewSimpleAllocatorFromDb(db *bolt.DB) *SimpleAllocator {
 				return err
 			}
 
+			// Add Cluster to ring
+			if err = s.AddCluster(cluster.Info.Id); err != nil {
+				return err
+			}
+
 			for _, nodeId := range cluster.Info.Nodes {
 				node, err := NewNodeEntryFromId(tx, nodeId)
 				if err != nil {
@@ -90,15 +95,15 @@ func (s *SimpleAllocator) AddDevice(cluster *ClusterEntry,
 	node *NodeEntry,
 	device *DeviceEntry) error {
 
-	s.lock.Lock()
-	defer s.lock.Unlock()
-
 	// Create a new cluster id if one is not available
 	clusterId := cluster.Info.Id
-	if _, ok := s.rings[clusterId]; !ok {
-		s.rings[clusterId] = NewSimpleAllocatorRing()
+	// TODO: in the future, we should do this call separately
+	if err := s.AddCluster(clusterId); err != nil && err != ErrFound {
+		return err
 	}
 
+	s.lock.Lock()
+	defer s.lock.Unlock()
 	s.rings[clusterId].Add(&SimpleDevice{
 		zone:     node.Info.Zone,
 		nodeId:   node.Info.Id,
--- a/apps/glusterfs/allocator_simple.go
+++ b/apps/glusterfs/allocator_simple.go
@@ -132,6 +132,24 @@ func (s *SimpleAllocator) RemoveDevice(cluster *ClusterEntry,
 	return nil
 }
 
+// AddCluster adds an entry to the rings map. Must be called before AddDevice so
+// that the entry exists.
+func (s *SimpleAllocator) AddCluster(clusterId string) error {
+
+	s.lock.Lock()
+	defer s.lock.Unlock()
+
+	if _, ok := s.rings[clusterId]; ok {
+		logger.LogError("cluster id %s already exists", clusterId)
+		return ErrFound
+	}
+
+	// Add cluster to map
+	s.rings[clusterId] = NewSimpleAllocatorRing()
+
+	return nil
+}
+
 func (s *SimpleAllocator) RemoveCluster(clusterId string) error {
 
 	s.lock.Lock()
--- a/apps/glusterfs/app_cluster.go
+++ b/apps/glusterfs/app_cluster.go
@@ -39,8 +39,14 @@ func (a *App) ClusterCreate(w http.ResponseWriter, r *http.Request) {
 			return err
 		}
 
-		return nil
+		err = a.allocator.AddCluster(entry.Info.Id)
+		if err != nil {
+			logger.LogError("Error adding cluster to ring: %v", err)
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return err
+		}
 
+		return nil
 	})
 	if err != nil {
 		return
--- a/apps/glusterfs/allocator.go
+++ b/apps/glusterfs/allocator.go
@@ -17,6 +17,9 @@ type Allocator interface {
 	// Inform the brick allocator to not use the specified device
 	RemoveDevice(c *ClusterEntry, n *NodeEntry, d *DeviceEntry) error
 
+	// Add cluster information from allocator.
+	AddCluster(clusterId string) error
+
 	// Remove cluster information from allocator
 	RemoveCluster(clusterId string) error
 
--- a/apps/glusterfs/allocator_mock.go
+++ b/apps/glusterfs/allocator_mock.go
@@ -86,6 +86,17 @@ func (d *MockAllocator) RemoveDevice(cluster *ClusterEntry,
 	return nil
 }
 
+// AddCluster adds a cluster entry to our "clustermap".
+func (d *MockAllocator) AddCluster(clusterId string) error {
+	// Save in the object
+	d.lock.Lock()
+	defer d.lock.Unlock()
+
+	d.clustermap[clusterId] = []string{}
+
+	return nil
+}
+
 func (d *MockAllocator) RemoveCluster(clusterId string) error {
 	// Save in the object
 	d.lock.Lock()
--- a/apps/glusterfs/allocator_simple_test.go
+++ b/apps/glusterfs/allocator_simple_test.go
@@ -57,6 +57,7 @@ func TestSimpleAllocatorAddRemoveDevice(t *testing.T) {
 	device := createSampleDeviceEntry(node.Info.Id, 10000)
 
 	tests.Assert(t, len(a.rings) == 0)
+	tests.Assert(t, a.AddCluster(cluster.Info.Id) == nil)
 	err := a.AddDevice(cluster, node, device)
 	tests.Assert(t, err == nil)
 	tests.Assert(t, len(a.rings) == 1)
--- a/apps/glusterfs/errors.go
+++ b/apps/glusterfs/errors.go
@@ -15,6 +15,7 @@ import (
 
 var (
 	ErrNoSpace          = errors.New("No space")
+	ErrFound            = errors.New("Id already exists")
 	ErrNotFound         = errors.New("Id not found")
 	ErrConflict         = errors.New("The target exists, contains other items, or is in use.")
 	ErrMaxBricks        = errors.New("Maximum number of bricks reached.")
--- a/apps/glusterfs/volume_entry_test.go
+++ b/apps/glusterfs/volume_entry_test.go
@@ -46,6 +46,10 @@ func setupSampleDbWithTopology(app *App,
 		for c := 0; c < clusters; c++ {
 			cluster := createSampleClusterEntry()
 
+			if err := app.allocator.AddCluster(cluster.Info.Id); err != nil {
+				return err
+			}
+
 			for n := 0; n < nodes_per_cluster; n++ {
 				node := createSampleNodeEntry()
 				node.Info.ClusterId = cluster.Info.Id


purpleidea:
actually, don't merge this yet, got a better patch.

purpleidea:
here we go. split into multiple patches because @obnoxxx likes it that way.

should fix both causes of this bug now.

have a nice weekend everyone!

purpleidea:
Due to some non-standard usage of the golang test infrastructure, the test panics, without saying why or without saying which test was running. Of course, this is explained in https://github.com/heketi/heketi/issues/851

So I have no idea why this is broken, please explain how this output is useful:
https://travis-ci.org/heketi/heketi/jobs/276114447#L762

Otherwise, patch LGTM

obnoxxx:
You can manually run the test suite with `go test -v github.com/heketi/heketi/apps/glusterfs/`.
This shows you that it fails at `TestDeviceAddDelete`.
You can manually run the individual test with `go test -run TestDeviceAddDelete -v github.com/heketi/heketi/apps/glusterfs/`.

obnoxxx:
@centos-ci ok to test

purpleidea:
@obnoxxx How can you tell it was `TestDeviceAddDelete` and at which line?
The test case: https://travis-ci.org/heketi/heketi/jobs/276114447#L762 does not show that string anywhere... I suspect this is because of the Assert package that I suggested we remove.

purpleidea:
@obnoxxx In fact, that string doesn't appear in the CentOS ci output either, and nor does it fail for anything related to this patch AFAICT.

`/me confused face`

obnoxxx:
@purpleidea correct, i saw that by manually running `go test -v ...` where it failed (see above comment). Also the centos-ci output would not catch the exact same error since it runs the functional tests, not the unit tests.

obnoxxx:
so in order for travis output to be more useful, we could add `-v` to the invocation of go test in `./Makefile` --> good idea?

purpleidea:
If you think that will help, then sure :)

However tests using the tests package normally, that fail always
report their failures (which test) and so I think this shouldn't be
necessary. I think the heketi/tests package is the problem.

(Working on that patch)

obnoxxx:
@purpleidea go fmt failed

obnoxxx:
great. travis succeeded. will review

apps/glusterfs/allocator_simple.go
```diff
@@ -89,15 +95,15 @@ func (s *SimpleAllocator) AddDevice(cluster *ClusterEntry,
...
 
-	s.lock.Lock()
-	defer s.lock.Unlock()
-
 	// Create a new cluster id if one is not available
 	clusterId := cluster.Info.Id
-	if _, ok := s.rings[clusterId]; !ok {
-		s.rings[clusterId] = NewSimpleAllocatorRing()
+	// TODO: in the future, we should do this call separately
+	if err := s.AddCluster(clusterId); err != nil {
```
obnoxxx: Is this correct? This will bail out if the cluster already exists...

purpleidea: Unfortunately, that *is* correct. AFAICT. This is technical debt speaking. Someone made an early mistake to have a "cluster" added silently if we added a "device", but the cluster was missing. They should have been two separate pieces but they are intertwined everywhere. Eventually in a down-the-road-part-two we'll fix this up if we get there, which is why I explicitly added the TODO there, and handle the error.

You'll note the AddCluster currently doesn't ever error, but some future patch could change that, hence handling the error correctly now.

obnoxxx: Side rant: How on earth can I properly reply in this horrbile github thing? Do I always need to copy and paste the previous comment myself... :-(


> Unfortunately, that is correct. AFAICT. This is technical debt speaking. Someone made an early mistake to have a "cluster" added silently if we added a "device", but the cluster was missing. They should have been two separate pieces but they are intertwined everywhere. Eventually in a down-the-road-part-two we'll fix this up if we get there, which is why I explicitly added the TODO there, and handle the error.

I hear what you are saying about how it can be improved. But you are changing this behavior here: You error out in `AddDevice` if the cluster already exist. (Because your `AddCluster` errors out in that case.) Hence I think this is not correct. If it is, I have not understood your explanation above, and please be more specific.

> You'll note the AddCluster currently doesn't ever error, but some future patch could change that, hence handling the error correctly now.

AFAICT this is not correct: As I said in my comment AddCluster errors out if the cluster already exists. Quoting from your patch (that adds AddCluster):

```
+	if _, ok := s.rings[clusterId]; ok {
+		logger.LogError("cluster id %s already exists", clusterId)
+		return ErrFound
+	}
```

How is this not erroring out?

purpleidea: You're not looking at the correct version of the patch. I suspect you're getting confused by the GitHub UI. That diff doesn't exist in the latest patch.

obnoxxx: No, I am looking exactly at the current version of the patch. Have you pushed your latest changes? ...

purpleidea: yup, see here for a direct link: https://github.com/heketi/heketi/pull/850/files/7eae99db3e15bd77887050eaf79fa26c4e60876f#diff-36d2e1cdec2dabf576d0bc9e6deda2fcR148

obnoxxx:
Hi @purpleidea ! My previous change requests still stand:
* There is a patch called "apps: glusterfs: Add simple allocator for Add". This originally did exactly this. Now (probably after an accidential squash) it does what the commit message says plus it adds calls to AddCluster in several places and rewrites AddDevice. This should be split into multiple patches again, please.
* I don't quite like the approach that you first add the AddCluster function as it should be, and then later when you use it and rewrite AddDevice silently change it. This is misleading, imho. Can't we either fix it properly right away or at least allow both `err == nil` and `err == ErrFound` from the AddCluster call in AddDevice? (I guess the latter approach would be good for now.)

apps/glusterfs/allocator_simple.go
```diff
@@ -132,6 +132,24 @@ func (s *SimpleAllocator) RemoveDevice(cluster *ClusterEntry,
...
+
+	s.lock.Lock()
+	defer s.lock.Unlock()
+
+	if _, ok := s.rings[clusterId]; ok {
+		logger.LogError("cluster id %s already exists", clusterId)
+		return ErrFound
+	}
+
+	// Add cluster tp map
```
obnoxxx: typo: `tp`

purpleidea: fixed

purpleidea:
@obnoxxx Patch is split as requested. If you look closely, you'll note I *have* to do the weird adding of the function, and then the edit afterwards-- it's not possible without squashing (as I did previously which you didn't like) because the interface method doesn't exist yet!

Anyways, typo fixed as well, and split some more things out since I'm getting the impression you like lots of little patches.

As for your comments about the err == nil, I don't know of another way to do this correctly without ignoring errors (this is worse) so either add a patch on top (which I can review) or rephrase so I understand more clearly.

Ready for review, thanks!

obnoxxx:
> @obnoxxx Patch is split as requested.

Thanks. But I have to say that the commit messages are still somehow confused now:

* "apps:glusterfs: Add mock allocator for Add" should probably be called "apps:glusterfs: Add the AddCluster method to the mock allocator".
* Similarly for the simple allocator and the interface commits.
* These three commits could even be squashed if you wanted (adding the interface method + the implementations), but this way or another, the commit message(s) should be reworded as described above.
* The commit called "apps:glusterf: Add the AddCluster method" does not add this method (this is done by the previous patches). Instead it **makes use** of this method **and modifies** it. So the commit should be renamed, and it should not silently change the implementation of the AddCluster method.

> If you look closely, you'll note I have to do the weird adding of the function, and then the edit afterwards-- it's not possible without squashing (as I did previously which you didn't like) because the interface method doesn't exist yet!

Yes, I have looked very closely. I can see no reason for first adding the implementation as it should be, and then sneak in a change of the function in the commit that you put it to use. There are at least two cleaner ways around this:
1. Commit the function initially with the workaround and commented-out code directly. So the commit that makes use of it is clean.
2. Commit the function like you do now (i.e. like it should be), and treat the result at the callers a little differently. Instead of 
```
if err = s.AddCluster(cluster.Info.Id); err != nil {
    return err
}
```
you'd do something like:
```
if err = s.AddCluster(cluster.Info.Id); err != nil && err != ErrFound {
    return err
}
```
 Of this I greatly prefer the last.

> As for your comments about the err == nil, I don't know of another way to do this correctly without ignoring errors (this is worse) so either add a patch on top (which I can review) or rephrase so I understand more clearly.

I don't want to add patches on top if we can get this right. See the above, I hope this clarifies.

Also don't get me wrong, the actual fix is good! At least as an initial fix. But the presentation in patches is still confusing to me. We should not stop here though, and do the deeper more conceptual fix that you mentioned in the comments. That could be done as a follow-up patch though.

apps/glusterfs/app_cluster.go
```diff
@@ -46,6 +46,11 @@ func (a *App) ClusterCreate(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	if err := a.allocator.AddCluster(entry.Info.Id); err != nil {
+		// TODO: lol, error handling
```
obnoxxx: What is the point of this comment? Why here and not in other places? And why the "lol"?

obnoxxx:
@purpleidea thanks for the update!

There' still the fact that you'e modifying the AddCluster method in the patch that's first putting it to use. Why? With the follow-up patch that you added ("apps: glusterfs: Add ErrFound checks") this modification is made unnecessary. So why don't you simply omit the modification of AddCluster and squash the  last patch with the patch that puts AddCluster to use?

purpleidea:
@obnoxxx 

Firstly: when you refer to commits like: "squash the last patch" and "the patch that puts AddCluster" instead just use SHA1 id's please. It makes it much easier to read.

Secondly: I think I'm slowly understanding your objections, still think you're mistaken, but can now hopefully explain better:

In: 29dc275ddddc467c9568dd097fb13b8fa6832468 we write the function AS IT SHOULD BE, but it's not used anywhere as this would break functionality since the rest of the project got the API wrong IMO.

Next in: bb6a60a2a89b3f6c2fb3986664d4f140f0cdb736 we finally USE it, but modify it accordingly. This patch is designed to be mostly REVERTED if we ever fix the rest of the API. It serves as documentation for what we had to do to make the API work the wrong way now.

This is also why there is an "lol, error handling" comment because the author used this weird HTTP api instead of a sane interface that returns an error which is then wrapped.

Hope this makes it clearer. I've updated the patch set to make this clearer.

obnoxxx:
> Firstly: when you refer to commits like: "squash the last patch" and "the patch that puts AddCluster" instead just use SHA1 id's please. It makes it much easier to read.

Point taken. If github can just interpret them right, that's even convenient. ;-)

> Secondly: I think I'm slowly understanding your objections, still think you're mistaken, but can now hopefully explain better:
> 
> In: 29dc275 we write the function AS IT SHOULD BE, but it's not used anywhere as this would break functionality 

Well, that's not entirely true.
Using and allowing ErrFound as a vaild success return condition
at some of the callers would give us vaild use of the correct function.

> since the rest of the project got the API wrong IMO.

Could you please be more specific?
"The rest of the project got the API wrong" sounds like general bitching about.

I guess you rather mean some pieces of the logic?
The problem we were facing here is that the cluster was not consistently added before calling (in this case) AddDevice. So you chose to amend for it by calling AddCluster from within AddDevice. How is that related to a "wrong API" ?

One problem that exists is that the ring structure, which is only meant as a certain view to the DB's contents is long-lived. It should actually be created for each volume request and destroyed afterwards. Also different types of volume create request might require a different type of ring. etc. Once that is done, we get rid of most of the calls to AddCluster, AddDevice, etc.. 

> Next in: bb6a60a we finally USE it,

Btw, IMHO, it is pointless to point out in the commit message with bold face and exclamation marks that it could not be used earlier. That's just stating the obvious. You should either write a helpful commit message or just leave it at the subject line. ;-)

> but modify it accordingly. 

I keep repeating myself, but I don't like that. If you need to modify it, that should be a separate patch. But I am still convinced that the modification is not necessary. And it would be cleaner without it.

> This patch is designed to be mostly REVERTED if we ever fix the rest of the API. It serves as documentation for what we had to do to make the API work the wrong way now.

This is the approach I don't like. First of all, we would not revert it in all places. E.g. we would keep the call in ClusterCreate, I guess. Instead of modifying the function and planning to revert the patch, as I tried to express a few times, we should *keep* the AddCluster function as it is (correct), and allow the ErrFound condition in some of the callers instead (those callers that work around the lack of having called AddCluster earlier, e.g. AddDevice).

> This is also why there is an "lol, error handling" comment because the author used this weird HTTP api instead of a sane interface that returns an error which is then wrapped.

Are you aware that this is the outer layer of the API server? It's RESTful, http-based. What could it do except for using http? It can't return a proper error as return value.

Anyways, if you feel like adding a comment, it should be helpful and explain.  This one is not really helpful. ;-)

> Hope this makes it clearer.

Well, not really. I understood the points you elaborated in this comment. I feel you are mostly stating the "what" instead of really explaining the "why". Also I am yet to hear a constructive reply to the concrete proposals I have made for a slightly different approach. Maybe I have not been clear. Code speaks, I'll prepare a branch on top of yours to show what I mean.

obnoxxx:
@purpleidea  While at it:

* One thing that is still missing from this PR: There should be a regression test that is failing without and succeeding with the patch. ideally at the golang function level tests.

* If you have spotted problems in the APIS or internal logics, do you have Ideas on how to solve them? Don't you want to go and work on improvements?

Thanks - Michael

obnoxxx:
LGTM
will wait for centos-ci to finish

obnoxxx:
LGTM. will merge later - waiting for the tests...

obnoxxx:
Thanks for your patience, @purpleidea , I know this was hard on you. ;-)

purpleidea:
@obnoxxx Writing the patch wasn't hard, reading the code was ;)

w00t!

PR #852, opened 2017-09-18
build: remove outdated comments from Makefile

obnoxxx:


obnoxxx:
OK, that's weird. The centos-ci environment fails to bring up VMs... :-o
Unrelated, of course.

PR #854, opened 2017-09-19
test:functional: print errors for asserts in TestVolumeSnapshotBehavior

obnoxxx:
This makes it easier to debug a problem.


obnoxxx:
Ok. yet another problem with the centos-ci this time: error to build heketi for each one of them.
I verified the patch locally, so merging...

PR #858, opened 2017-09-19
Deps tarball

obnoxxx:
Create a dependencies tarball.


raghavendra-talur:
I still get something like heketi-deps-v5.0.0-10-g89abc968-release5.tar.gz on my release branch. I don't understand changes made with VERSION part. Rest looks good.

obnoxxx:
@raghavendra-talur correct, because the version tag `v5.0.0` is not at the branch head...
but instead 8 patches deep in. if you add a tag to the top commit, then `make version` will show exactly the tag.

obnoxxx:
Thinking of this, I guess i should disambiguate "TAG" in case there are multiple tags on the HEAD.

obnoxxx:
updated to only consider version tags (vA.B.C) and to always pick the newest tag when there are multiple tags on HEAD.

obnoxxx:
strange and unrelated failure of centos-ci run.

PR #862, opened 2017-09-21
Fix crosscompile

obnoxxx:
This fixes cross-compilation and thereby "make release"
by only conditionally using caching (go build -i instead of go build).


obnoxxx:
centos-ci failure is unrelated.

PR #863, opened 2017-09-21
apps: glusterfs: test: Use standard testing library

purpleidea:
For reasons unknown to the patch author, a special "heketi/tests"
library wrapper was used instead of the standard testing semantics. This
makes it difficult for new users to read the code, and hard to trace
things properly.

As well, this code can now "return" and end a test early without
resorting to a panic.

While the current code is now longer, this is due to the fact that this
"unit" test is more of an "integration test" and tests all sorts of
things constantly, even if many probably won't fail and which don't add
much value to the test. Simplifying that can be done in future patches.

This new style of testing (as seen in the golang stdlib) should be used
for all patches going forward.

### DIFF
--- a/apps/glusterfs/app_device_test.go
+++ b/apps/glusterfs/app_device_test.go
@@ -124,7 +124,10 @@ func TestDeviceAddDelete(t *testing.T) {
 		}
 		return nil
 	})
-	tests.Assert(t, err == nil)
+	if err != nil {
+		t.Errorf("error updating database: %+v", err)
+		return
+	}
 
 	// Create a request to a device
 	request := []byte(`{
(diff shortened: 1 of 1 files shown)


obnoxxx:
@purpleidea, thanks for the patch!

Let me go through a couple of questions/comments mostly regarding the commit message, but also the patch:

I'd like to understand what's really so bad about the `heketi/tests` package. From looking at it, it merely wraps the standard library test code and provides a more standard logging output and reduce the boilerplate code needed to exit a test early. Best I can tell, it does not panic (as you claim).  For sure it does not replace the standard test code but wraps it.

I am asking what's really so bad about `tests/heketi` because the way the code looked before, is exactly like I was used to writing test code before entering heketi development. It seems to follow best practices. Basically what `heketi/tests` has done should be part of the core `testing` package, imho. So I never questions it's use...

This patch basically adds back boilerplate code for early test exit and non-standardized log output. If I am missing subtle points, please explain!

Also, addressing your comment that this is not a real unit-test: this might not be the best example. If you look into those modules that do not start with `app_` you may see more unit-test-like code. But this is more on the unit-test side, in that it is not a full end-to-end functional or integration test but tests certain functions in the stack, in this case the higher level API server. This is done by mostly mocking the layers below this. So maybe not a unit test in the strictest sense but close enough imho.

I also don't understand how this kind of change only makes the code longer because this is (as you claim) not a unit test more of an integration test: I think it would make any test code longer, proportional to the the number of Assert() calls in the test. Or what am I missing?

Thanks - Michael

purpleidea:
> I'd like to understand what's really so bad about the heketi/tests package. From looking at it, it merely wraps the standard library test code and provides a more standard logging output and reduce the boilerplate code needed to exit a test early.

It does reduce the amount of boilerplate! However I think there is simply too much going on in these tests which is why it looks so cluttered.

> Best I can tell, it does not panic (as you claim). For sure it does not replace the standard test code but wraps it.

Indeed it *has* to panic because if you have code like:

```
if true {
    t.Errorf("we failed")
    return # ends the test
}
```

you can end right there, but with:

```
if true {
    heketi.Assert(foo == true, "fail message")
    # if this doesn't panic, then we'd keep on running after we had a failure!
    # note: sometimes we *do* want to keep running, but that's another story, and doesn't apply to any of the tests here.
}
```

> I am asking what's really so bad about tests/heketi because the way the code looked before, is exactly like I was used to writing test code before entering heketi development. It seems to follow best practices. Basically what heketi/tests has done should be part of the core testing package, imho. So I never questions it's use...

I don't think I parsed all of this correctly, but yeah, it probably makes it look like what you're used to in python or C or whatever. But this *isn't* one of those languages... If you think these Assert methods are that important, propose them upstream and see what those folks say. You'll probably get a better answer than from me, and we'd all learn more about their design reasons.

It's true that some *large* golang projects have test util libraries, but in general, it's my understanding that they should be discouraged if not absolutely necessary.

> This patch basically adds back boilerplate code for early test exit and non-standardized log output. If I am missing subtle points, please explain!

Yep :) It also makes things more readable because you very clearly see exactly which line the test fails on. For example, run the test with the old version and the new version but first change the code somewhere so it actually fails. Paste the output, which is easier for a new programmer to grok?

This could probably be improved in the lib, but that misses the point-- just use the standard conventions. The project doesn't really have any non-Red Hat contributors anyways, why make the bar higher?

> Also, addressing your comment that this is not a real unit-test: this might not be the best example.

It is the best example, because as a new contributor I had to decipher all this stuff, and it added overhead from my perspective. I would bet that anyone else would just look at this, see we're doing something non-standard, and then instantly go write code elsewhere. I know I would have if I wasn't being paid to look at this.

Lastly, if we *really* *absolutely* were convinced that we needed a test util lib, then we should have used something else instead of going the NIH route. It feels like resources weren't prioritized correctly when writing all these pieces, but then again, maybe there is some background I didn't know about.

Hope that helps.

obnoxxx:
@purpleidea after having discussed and hacked on this with you for a while and also having interviewed a few other developers of the project, I see my point confirmed that I want to stay with the current assert model for now:

* It's a tiny extension of the standard test library (~ 15 lines).
* It greatly improves readability of the code by reducing boilerplate code.
* It uses the FailNow() function of the standard test lib, just replacing asserts by returns is not correct.

It is true that the actual output format of the Assert function in the error case
might not be optimal.  ==> We can change that in the heketi/tests repo.
But at least it gives us a uniform output in case of error.

It is also true that we have not made sufficient use of the third optional
parameter in the assert function to specify a descriptive error message for
the failure. This could be improved. For future additions and touching
the code wherever we stumble across uses that irritate.

Based on the above, let's not banish the Assert function right now
concentrate on other changes.

PR #865, opened 2017-09-21
Cleanup makefiles

obnoxxx:
* Standardize on tabs vs spaces
* consolidate multiple .PHONY lines

Inspired by a comment from @purpleidea .

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,19 +17,19 @@ GOHOSTOS := $(shell go env GOHOSTOS)
 GOBUILDFLAGS :=
 ifeq ($(GOOS),$(GOHOSTOS))
 ifeq ($(GOARCH),$(GOHOSTARCH))
-  GOBUILDFLAGS :=-i
+	GOBUILDFLAGS :=-i
 endif
 endif
 GLIDEPATH := $(shell command -v glide 2> /dev/null)
 DIR=.
 
 ifeq (master,$(BRANCH))
-  VERSION = $(VER)
+	VERSION = $(VER)
 else
 ifeq ($(VER),$(TAG))
-  VERSION = $(VER)
+	VERSION = $(VER)
 else
-  VERSION = $(VER)-$(BRANCH)
+	VERSION = $(VER)-$(BRANCH)
 endif
 endif
 
@@ -150,5 +150,5 @@ darwin_amd64_dist:
 release: deps_tarball darwin_amd64_dist linux_arm64_dist linux_arm_dist linux_amd64_dist
 
 .PHONY: server client test clean name run version release \
-        darwin_amd64_dist linux_arm_dist linux_amd64_dist linux_arm64_dist \
-        heketi clean_vendor deps_tarball
+	darwin_amd64_dist linux_arm_dist linux_amd64_dist linux_arm64_dist \
+	heketi clean_vendor deps_tarball
--- a/Makefile
+++ b/Makefile
@@ -2,8 +2,6 @@
 # Based on http://chrismckenzie.io/post/deploying-with-golang/
 #
 
-.PHONY: version all run dist clean
-
 APP_NAME := heketi
 CLIENT_PKG_NAME := heketi-client
 SHA := $(shell git rev-parse --short HEAD)
@@ -151,4 +149,4 @@ release: deps_tarball darwin_amd64_dist linux_arm64_dist linux_arm_dist linux_am
 
 .PHONY: server client test clean name run version release \
 	darwin_amd64_dist linux_arm_dist linux_amd64_dist linux_arm64_dist \
-	heketi clean_vendor deps_tarball
+	heketi clean_vendor deps_tarball all dist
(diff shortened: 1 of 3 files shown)


purpleidea:
@obnoxxx Thanks for the patch... ;)

I'd NACK this without a cute little test-makefiles.sh that search for "  " (space space) and fails :P 

Well maybe not NACK, but I'd encourage one...

Does this project have a test.sh main entry point which can link to many sub-tests? I'd include it there if we have one.

purpleidea:
Here: https://github.com/heketi/heketi/pull/866

Merge this first and then add the fixes after this! :)

obnoxxx:
@purpleidea good point.
This test is not as simple. We can't bail on all spaces....
Is there a format/syntax checker or lint-like tool for makefiles?

There is currently no central tests.sh.
there is only the test target in Makefile.
We could externalize its content into a tests.sh.

purpleidea:
Also: Commits usually have the format:

lowercaseprefix: optionalsecondprefix: Capitalized message with no trailing period

obnoxxx:
@purpleidea writes:
> Also: Commits usually have the format:
> 
> lowercaseprefix: optionalsecondprefix: Capitalized message with no trailing period

1) Where have you got this "usual format" from?
2) Except for the capitalized, I have even adhered to that format, so I'm not disagreeing in general, I have been doing this more or less anyways (but without the space  between prefix and secondprefix), but just because it seemed good to me.

purpleidea:
Well-known, eg: https://chris.beams.io/posts/git-commit/
and many other sources
Actually GlusterFS (main project) and mgmt both have a CI test for this.

obnoxxx:
Thanks for the link about the commit message format. Not sure who chris.beams is, but certainly does not seem to be a universal standard, and all of these variations of a format that you find when browsing a little are to some extent taste-driven and opinionated. Seems to stem largely back to tim pope http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html . Note that none of these format descriptions seem to mention the prefixes at all, which I personally like very much. But I e.g. do not use a space between a first and second prefix (if i use one). And some patches can't find a good prefix at all. So the should be encouraged but not mandated. Also the "don't use a full stop at the end of the first message" seems only to spare characters. I wouldn't care if the message is short enough. And if we omit the full stop, I couldn't care less about capitalization. actually, if it's not a complete sentence anyways, i sometimes even prefer it to start in lower case. It's also funny that none of these descriptions seem to give a reason for the capitalization. All of the other rules are logical to me, and I intuitively try to apply them myself most of the time.

So after all of this, i mostly agree on the commit msg format. And i might start writing down our own variation of this in a doc and possibly in a patch acceptance test as well.

Thanks!

purpleidea:
@obnoxxx yw! I think the convention I am describing is common in the ecosystems I've travelled in. Surely other people have other conventions. Anyways, I'll leave the decision to you. One quick point:

doing:

`prefix:second: Message here`

vs.

`prefix: second: Message here`

The second (added space) is more readable, and more logical when extending with things like:

`prefix: second, third: Message here`

otherwise without the space it would look weird. Our CI tests all this correctly. Don't NIH a happy solution. Etcd does the same thing for example (look at their commits)

obnoxxx:
> @obnoxxx yw! I think the convention I am describing is common in the ecosystems I've travelled in. Surely other people have other conventions. Anyways, I'll leave the decision to you. One quick point:
> 
> doing:
> 
> `prefix:second: Message here`
> 
> vs.
> 
> `prefix: second: Message here`
> 
> The second (added space) is more readable, and more logical when extending with things like:
> 
> `prefix: second, third: Message here`
> 
> otherwise without the space it would look weird. Our CI tests all this correctly. Don't NIH a happy solution. Etcd does the same thing for example (look at their commits)

To me,

`prefix: second, third: Message`

is utterly inconsistent. IMHO it should be

`prefix:second:third: Message`.

Also I personally find the added space decreases readability.

The projects I have been using most seem to have applied a variation. Gluster (which you mentioned) seems to be using a pattern of:

`first/second: Message`

The Samba project does not have a strict rule, but I see

`first:second: Message`


`first/second: Message`

and sometimes

`first: second: Message`


I think for heketi I'll want to settle on s/th like 

`[first:[second:[...]]] message`

With no rule on capitalization (or punctuation) of "message", possibly with an optional space between prefixes. I'll think a little more about it.

That's not NIH but my preferred way of all the projects I've worked with.

Cheers - Michael

obnoxxx:
rebased to current master.
still might add some test to this pr

obnoxxx:
Added a test for the makefile.
Will integrate that when we have created a infra for running them.

obnoxxx:
don't merge yet, wait for PR #889 and rebase + make use of the makefile test in this pr.

obnoxxx:
Rebased on new test infra.

obnoxxx:
@raghavendra-talur this can be reviewed and (fwiw) merged now... :-)

obnoxxx:
@raghavendra-talur ping. This should be easy enough to review. :-)

obnoxxx:
Rebased over conflicting changes

PR #866, opened 2017-09-21
test: Improve test infrastructure

purpleidea:
This will simplify future patches.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -93,7 +93,8 @@ run: server
 	./$(APP_NAME)
 
 test: vendor glide.lock
-	go test $(GOFILES)
+	go test $(GOFILES)	# TODO: move this into a sub-test inside test/*.sh
+	./test.sh
 
 clean:
 	@echo Cleaning Workspace...
--- a/test.sh
+++ b/test.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+# main test runner for heketi
+
+# TODO: add tests here...


obnoxxx:
@centos-ci ok to test

obnoxxx:
1. This does not work like this (permissions on test.sh).

2. I would like to spend some more thought on the structure: Putting a `test.sh` in the top level seems arbitrary. Maybe we should put it under `tests/`. Also putting a `test.sh` call in the Makefile target *after* calling `go test ...` seems arbitrary: why not put the call to `go test ...` into a test script of its own, and have that called, as the only script for a start, from the wrapping `tests.sh`, which would just iterate over calling various test scripts (to be written...).

purpleidea:
@obnoxxx Great questions:

> This does not work like this (permissions on test.sh).

I can fixup patch if I forgot to set the bit.

> Putting a test.sh in the top level seems arbitrary

It's not. It's the main entry point where either a script could find it, or where the user manually running it will be and look for it.

> putting a test.sh call in the Makefile target after calling go test ... seems arbitrary

Indeed the `go test` is only there temporarily until it gets moved down into $ROOT/test/gotest.sh which should get called by $ROOT/test.sh

obnoxxx:
> > This does not work like this (permissions on test.sh).
> 
> I can fixup patch if I forgot to set the bit.

Yeah, that what I was asking for. :-)

> >  Putting a test.sh in the top level seems arbitrary
> 
> It's not. It's the main entry point where either a script could find it, or where the user manually running it will be and look for it.

Well, on the other hand side I'd argue, the user can run `make test`. If she wants to run tests manually, it can be expected to apply some amount of brains. But we can still fix that later on.

> > putting a test.sh call in the Makefile target after calling go test ... seems arbitrary
> 
> Indeed the go test is only there temporarily until it gets moved down into $ROOT/test/gotest.sh which should get called by $ROOT/test.sh

So we're on the same page here. Fine by me, then.

If you change the file mode (please force-push over), I can merge this.

purpleidea:
@obnoxxx Patch updated with mode change.

Now that I review my patch, I think the questions you have are all addressed by the original source comments I added :) So please look there for the answer. It clearly says under `make test`:

```
 test: vendor glide.lock
-       go test $(GOFILES)
+       go test $(GOFILES)      # TODO: move this into a sub-test inside test/*.sh
+       ./test.sh
```

So obviously you can run `make test`.

Also this explains the other point.

obnoxxx:
> @obnoxxx Patch updated with mode change.

Thanks @purpleidea .

> Now that I review my patch, I think the questions you have are all addressed by the original source comments I added :) So please look there for the answer.

Are you _really_ complaining that I asked questions and you had to answer?? ;-)

Or rather the points that you had to discuss. And no, the commit did not answer all points I wanted to discuss a priori.

> It clearly says under make test:
> ```
> test: vendor glide.lock
>        go test $(GOFILES)
> +       go test $(GOFILES)      # TODO: move this into a sub-test inside test/*.sh
> +       ./test.sh
> ```
> 
> So obviously you can run make test.

Er... no-one questioned that you can still run make test (once the perms were fixed).

> Also this explains the other point.

Which to be precise?
* One point was why do only this step and not go further in this patchset. You have clearly stated that this is only the first step. So fine by me. Will do further refactoring. 
* The other point was why to put the main test script into `./` as opposed to having it in (e.g.) `./tests/`. That was the thing about running `make test` if you want it the easy way. Also here, we could still move the test script later on. ;-)

Given all that' we're good to go with this patch.

purpleidea:
> Are you really complaining that I asked questions and you had to answer?? ;-)

Nope, more that I thought it was obvious enough. I guess not, sorry.

Anyways, thanks for the merge.

PR #872, opened 2017-10-03
Add block and file flags on upgrade

raghavendra-talur:


apps/glusterfs/cluster_entry.go
```diff
@@ -168,5 +168,51 @@ func (c *ClusterEntry) NodeDelete(id string) {
...
 func ClusterEntryUpgrade(tx *bolt.Tx) error {
+	err := addBlockFileFlagsInClusterEntry(tx)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func addBlockFileFlagsInClusterEntry(tx *bolt.Tx) error {
+
```
obnoxxx: extra blank line.

apps/glusterfs/cluster_entry.go
```diff
@@ -168,5 +168,51 @@ func (c *ClusterEntry) NodeDelete(id string) {
...
+	if err != nil && err != ErrNotFound {
+		return err
+	}
+
+	if err == ErrNotFound {
+		entry = NewDbAttributeEntry()
+		entry.Key = DB_HAS_FILE_BLOCK_FLAG
+		entry.Value = "no"
+	} else {
+		//This case is only for future, if ever we want to set this key to "no"
```
obnoxxx: space between `//` and `This`

obnoxxx: One point is that if `Value == "yes"`, this means that the block/file flags have been present at previous starts of the software on this DB, and hence have been initialized. ==> we can skip this upgrade step.

apps/glusterfs/dbattribute_entry.go
```diff
@@ -0,0 +1,74 @@
+//
+// Copyright (c) 2017 The heketi Authors
+//
+// This file is licensed to you under your choice of the GNU Lesser
+// General Public License, version 3 or any later version (LGPLv3 or
+// later), or the GNU General Public License, version 2 (GPLv2), in all
+// cases as published by the Free Software Foundation.
+//
+
```
obnoxxx: This seems to be a very general, very thin wrapper around the db functions. What's the point?

obnoxxx:
Can we have tests? I think there's a gap in the new-db case (see previous comment).

apps/glusterfs/cluster_entry.go
raghavendra-talur: done

apps/glusterfs/cluster_entry.go
raghavendra-talur: fixed first comment.

We do skip the upgrade step if value is yes. This check exists for future deprecation.

apps/glusterfs/dbattribute_entry.go
raghavendra-talur: As this is a new bucket, all the functions that operate on a bucket need to be implemented again.

raghavendra-talur:
> Do we add the DB_HAS_FILE_BLOCK_FLAG key into the DB upon creation of a new DB with this software version? It seems that we do not. Hence we would create the db and (afaict) at next open do the upgrade step and undo anything that the admin has set at first run. Please clarify.

We do handle this scenario. When new version of heketi starts, as part of NewApp we call app.Upgrade(). Here we check for the existence of the flag and when we don't find one, we add the flag into dbattribute bucket.

apps/glusterfs/dbattribute_entry.go
obnoxxx: Got it. This looks so boilerplate key/value access that I was briefly confused... ;-)

obnoxxx:
> >  Do we add the DB_HAS_FILE_BLOCK_FLAG key into the DB upon creation of a new DB with this software version? It seems that we do not. Hence we would create the db and (afaict) at next open do the upgrade step and undo anything that the admin has set at first run. Please clarify.
> 
> We do handle this scenario. When new version of heketi starts, as part of NewApp we call app.Upgrade(). Here we check for the existence of the flag and when we don't find one, we add the flag into dbattribute bucket.

Ok. The point I missed was that we always do the DB initialization through the Upgrade function, which is kind of counter-intuitive. ;-)

obnoxxx:
@centos-ci ok to test

apps/glusterfs/cluster_entry.go
```diff
@@ -168,5 +168,50 @@ func (c *ClusterEntry) NodeDelete(id string) {
 }
 
 func ClusterEntryUpgrade(tx *bolt.Tx) error {
+	err := addBlockFileFlagsInClusterEntry(tx)
```
obnoxxx: From a consistency point of view, it is slightly surprising that all this done from the `ClusterEntryUpgrade` function, while the decisive bit is the `DbAttributeEntry` here... Thinking out loud, it might be the most natural approach might be to look at the `DbAttributeEntry` in `ClusterEntryUpgrade` and if not set, do the flag-setting here, but only in a `DbAttributeEntryUpgrade` function (called at the end of `Upgrade()` set `DB_HAS_FILE_BLOCK_FLAG` to yes (unconditionally).

obnoxxx: Btw, this *could* of course be done as a clean-up afterwards if necessary.

apps/glusterfs/cluster_entry_test.go
```diff
@@ -380,3 +380,79 @@ func TestNewClusterEntryNewInfoResponse(t *testing.T) {
 	tests.Assert(t, reflect.DeepEqual(info.Nodes, c.Info.Nodes))
 	tests.Assert(t, reflect.DeepEqual(info.Volumes, c.Info.Volumes))
 }
+
+func TestUpdateClusterEntryForFlags(t *testing.T) {
+	tmpfile := tests.Tempfile()
+	defer os.Remove(tmpfile)
+
+	// Create the app
```
obnoxxx: redundant comment - remove

apps/glusterfs/cluster_entry_test.go
```diff
@@ -380,3 +380,79 @@ func TestNewClusterEntryNewInfoResponse(t *testing.T) {
...
+	app := NewTestApp(tmpfile)
+	//logger.SetLevel(utils.LEVEL_DEBUG)
```
obnoxxx: commented out code in new function - remove

apps/glusterfs/cluster_entry_test.go
```diff
@@ -380,3 +380,79 @@ func TestNewClusterEntryNewInfoResponse(t *testing.T) {
...
+
+	// Create a cluster
```
obnoxxx: redundant comment - remove

apps/glusterfs/cluster_entry_test.go
```diff
@@ -380,3 +380,79 @@ func TestNewClusterEntryNewInfoResponse(t *testing.T) {
...
+	cluster_req := &api.ClusterCreateRequest{
+		File: true,
+	}
+	c := NewClusterEntryFromRequest(cluster_req)
+	c.NodeAdd("node_abc")
+	c.NodeAdd("node_def")
+
+	// Save element in database
```
obnoxxx: redundant comment - remove

apps/glusterfs/cluster_entry.go
raghavendra-talur: Ok, my approach was:

1. what are we updating? ClusterEntry, hence in `ClusterEntryUpgrade()`
2. what are we updating with? flags, hence in `addBlockFileFlagsInClusterEntry()`
3. when should this update happen? only when `DB_HAS_FILE_BLOCK_FLAG` not set.

Yes, the above approach would become confusing if `DB_HAS_FILE_BLOCK_FLAG` is referenced in any other upgrade function; however, I don't think there is any such function.

Also, DbAttribute bucket should not have or require any upgrade function; the bucket itself is indicator of whether upgrade is required or not, isn't it?

apps/glusterfs/cluster_entry_test.go
```diff
@@ -380,3 +380,79 @@ func TestNewClusterEntryNewInfoResponse(t *testing.T) {
...
+		dbaentry, err := NewDbAttributeEntryFromKey(tx, DB_HAS_FILE_BLOCK_FLAG)
+		if err != nil {
+			return err
+		}
+		return dbaentry.Delete(tx)
+	})
+	tests.Assert(t, err == nil)
+	app.Close()
+
+	// Recreate the app
```
obnoxxx: The comment should mention that this triggers the upgrade path (that's the point).

apps/glusterfs/cluster_entry_test.go
```diff
@@ -380,3 +380,79 @@ func TestNewClusterEntryNewInfoResponse(t *testing.T) {
...
+	tmpfile := tests.Tempfile()
+	defer os.Remove(tmpfile)
+
+	// Create the app
+	app := NewTestApp(tmpfile)
+	//logger.SetLevel(utils.LEVEL_DEBUG)
+
+	// Create a cluster
+	cluster_req := &api.ClusterCreateRequest{
+		File: true,
```
obnoxxx: Should we explicitly set `Block: false` here, for clarity?

apps/glusterfs/cluster_entry_test.go
raghavendra-talur: done

apps/glusterfs/cluster_entry_test.go
raghavendra-talur: done

apps/glusterfs/cluster_entry_test.go
raghavendra-talur: done

apps/glusterfs/cluster_entry_test.go
raghavendra-talur: done

apps/glusterfs/cluster_entry_test.go
raghavendra-talur: done

apps/glusterfs/cluster_entry.go
obnoxxx: > Ok, my approach was:
> 
> 1. what are we updating? `ClusterEntry`, hence in `ClusterEntryUpgrade()`
> 1. what are we updating with? flags, hence in `addBlockFileFlagsInClusterEntry()`
> 1. when should this update happen? only when `DB_HAS_FILE_BLOCK_FLAG` not set.

Correct. But should the `ClusterEntryUpgrade` function fiddle with the `DbAttributeEntry` bucket?... I don't think so.

> Yes, the above approach would become confusing if `DB_HAS_FILE_BLOCK_FLAG` is referenced in any other upgrade function; however, I don't think there is any such function.

Right, not currently. The confusing bit is that the `ClusterEntryUpgrade` function changes the `DbAttribute` bucket. To my understanding, this bucket/entry introduces a way to make other parts of the upgrade process conditional, hence it's only natural that it should be referenced in various parts of the upgrade process.

> Also, DbAttribute bucket should not have or require any upgrade function;

Up to now all buckets have upgrade functions, mostly the `"return nil"` type. That's the consistency I was mentioning.

> the bucket itself is indicator of whether upgrade is required or not, isn't it?

No, afaict: You have created a bucket that can possibly (in the future) hold multiple entries. The existence/value of *one specific* entry in this bucket (DB_HAS_FILE_BLOCK_FLAG) is the indicator for the need to run the update of the cluster entries.

If you want to put it like this, as mentioned above, with the DbAttribute bucket, or more specifically the DB_HAS_FILE_BLOCK_FLAG key, you have created a way to make other parts of the Upgrade function conditional. So in order to be really future-proof, one could also see approach it like this (in Upgrade(), names not final yet):

1. Call `DbAttributeUpgradePhase1()`. This sets `DB_HAS_FILE_BLOCK_FLAG` to `no` if it does not exist.
2. Run all other `FooBarEntryUpgrade()` functions. `ClusterEntryUpgrade` does its new job iff `DB_HAS_FILE_BLOCK_FLAG` is set to `no`.
3. Call `DbAttributeUpgradePhase2()`. This sets `DB_HAS_FILE_BLOCK_FLAG` to `yes`.

Not sure if that makes sense.
As long as we agree on an approach that does not change the DB format any further, I'm also happy to propose an add-on patch later.

obnoxxx: One further thought: if we could detect whether the Block/File flags exist on a cluster entry (irrespective of their value), we could also do the upgrade work without the introduction of a new bucket and variable, right?

raghavendra-talur: >Correct. But should the ClusterEntryUpgrade function fiddle with the DbAttributeEntry bucket?... I don't think so.
>Right, not currently. The confusing bit is that the ClusterEntryUpgrade function changes the DbAttribute bucket. To my understanding, this bucket/entry introduces a way to make other parts of the upgrade process conditional, hence it's only natural that it should be referenced in various parts of the upgrade process.

I was aiming such that entries/keys in DbAttribute bucket influence only one of the sub-upgrade functions. If we achieve that, we can let the sub-upgrade function change the value of that one key/entry.

With respect to Phase1 Phase2 approach:
We thought about that, it has two minor problems:
1. If a sub-upgrade function performs the upgrade but a subsequent sub-upgrade function fails, Phase2 will never run and the state of flag will be "no". This might have unknown effect in next runs.
2. I wanted to differentiate between non-existence of flag and flag set to "no". If the Phase1 always sets flag to "no" then sub-upgrade functions will only have two states instead of 3.

Basically, we are talking about two options:
1. DBATTRIBUTE flags/keys that we define have one-to-one relation with sub-upgrade functions. Then what we have in PR is fine.
2. DBATTRIBUTE flags/keys have one-to-many relation with sub-upgrade functions. Then we need to have phase2 approach as we can't set flag to "yes" till all the sub-upgrade functions that reference the flag has run.

I need to think a little more before confidently saying approach 1 above with future proof enough.

raghavendra-talur:
>Thanks for adding the test. This tests that an upgrade run (enforced by unsetting the dbattribute) changes values of existing cluster file/block flags to true.

>Shouldn't we also have a test that completely removes the file/block flags from a cluster entry before triggering the upgrade path?

Tried to write this test, the problem I am facing is that ClusterEntry struct now has these two members `Block` and `File`. The marshal function will always make them false and save to db. It would never leave them `nil`.

One option is to redefine the struct in this test to the old definition and also implement the db wrapper functions for it. Seems like a lot of code for this test, do you see any other way?

obnoxxx:
> > Thanks for adding the test. This tests that an upgrade run (enforced by unsetting the dbattribute) changes values of existing cluster file/block flags to true.
> > 
> > Shouldn't we also have a test that completely removes the file/block flags from a cluster entry before triggering the upgrade path?
> 
> Tried to write this test, the problem I am facing is that ClusterEntry struct now has these two members Block and File. The marshal function will always make them false and save to db. It would never leave them nil.

Ouch, true.

> One option is to redefine the struct in this test to the old definition and also implement the db wrapper functions for it. Seems like a lot of code for this test, do you see any other way?

No, I currently don't. And probably it's not worth it since it would be kind of a self-fulfilling prophecy... It's slightly unfortunate, that we can't test the *genuine* update this way, but it seems that is the situation.

apps/glusterfs/cluster_entry.go
raghavendra-talur: >One further thought: if we could detect whether the Block/File flags exist on a cluster entry (irrespective of their value), we could also do the upgrade work without the introduction of a new bucket and variable, right?

Yes, this would have been the simplest. This is how it is for members of type string because we just compare them with "". However, with bool type, nil is false as we discovered while doing the topology load.

obnoxxx: > > Correct. But should the ClusterEntryUpgrade function fiddle with the DbAttributeEntry bucket?... I don't think so.
> > Right, not currently. The confusing bit is that the ClusterEntryUpgrade function changes the DbAttribute bucket. To my understanding, this bucket/entry introduces a way to make other parts of the upgrade process conditional, hence it's only natural that it should be referenced in various parts of the upgrade process.
> 
> I was aiming such that entries/keys in DbAttribute bucket influence only one of the sub-upgrade functions. If we achieve that, we can let the sub-upgrade function change the value of that one key/entry.
> 
> With respect to Phase1 Phase2 approach:
> We thought about that, it has two minor problems:
> 
> 1.  If a sub-upgrade function performs the upgrade but a subsequent sub-upgrade function fails, Phase2 will never run and the state of flag will be "no". This might have unknown effect in next runs.

Good point, you are right! In a more complicated future world at least, the intermediate upgrade steps would need to note their success/failure somewhere, So that it's always possible from looking at the DB to tell which parts of the upgrade were performed.

For the concrete case, we could/should make the sub-upgrade functions idempotent so that it's not a problem if they get re-run next time (until the upgrade function runs to completion).

In a more simple view, I think we need to mark the DB broken if some upgrade step fails. We need to make sure that no server code (other than the upgrade code) modifies the DB as long as it's not been completely successfully upgraded.

> 2. I wanted to differentiate between non-existence of flag and flag set to "no". If the Phase1 always sets flag to "no" then sub-upgrade functions will only have two states instead of 3.

My thought was that the differentiation would be in Phase1: it only sets the flag to "no" if it did not exist. Then the sub-upgrade function acts if it's set to "no". (I.e. value "no" is the indicator that upgrade is needed, maybe a different value would be more natural in this approach.) Then Phase2 sets it to yes.

> Basically, we are talking about two options:
> 
> 1. DBATTRIBUTE flags/keys that we define have one-to-one relation with sub-upgrade functions. Then what we have in PR is fine.

Well, all that is *very* implicit. So the code may work fine, but it may not be very future-proof.

> 2. DBATTRIBUTE flags/keys have one-to-many relation with sub-upgrade functions. Then we need to have phase2 approach as we can't set flag to "yes" till all the sub-upgrade functions that reference the flag has run.
> 
> I need to think a little more before confidently saying approach 1 above with future proof enough.

As mentioned earlier, afaict, the current code would work fine under the assumtions you described. And as long the DB format does not change, we can still change / improve the code acting on this data in the future.... I.e. there might not be a need to go to Phase1/Phase2 right now.

raghavendra-talur:
@obnoxxx I think now the PR is in right state.

obnoxxx:
LGTM.
As discussed, further points can be addressed later.

PR #874, opened 2017-10-04
Add block and file flags on upgrade

raghavendra-talur:


obnoxxx:
@centos-ci ok to test

Issue #887, opened 2017-11-02
Move the golang tests from the toplvl Makefile to a test script under tests/


obnoxxx:
This would continue the work started in PR #866 .

phlogistonjohn:
Issue was fixed but not closed. Closing. If I'm wrong feel free to slap me around a bit and reopen.

PR #889, opened 2017-11-02
Build out support for running short tests scripts in tests dir

phlogistonjohn:
Implements changes for issue #887 
Adds code to the "framework" script for running individual tests scripts under the tests dir. Moves go test invocation to a test script.
Starting off small with many opportunities for improvement. :-)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,7 +48,6 @@ LDFLAGS :=-ldflags "-X main.HEKETI_VERSION=$(VERSION) -extldflags '-z relro -z n
 PACKAGE :=$(DIR)/dist/$(APP_NAME)-$(VERSION).$(GOOS).$(GOARCH).tar.gz
 CLIENT_PACKAGE :=$(DIR)/dist/$(APP_NAME)-client-$(VERSION).$(GOOS).$(GOARCH).tar.gz
 DEPS_TARBALL :=$(DIR)/dist/$(APP_NAME)-deps-$(VERSION).tar.gz
-GOFILES=$(shell go list ./... | grep -v vendor)
 
 .DEFAULT: all
 
@@ -93,8 +92,7 @@ run: server
 	./$(APP_NAME)
 
 test: vendor glide.lock
-	go test $(GOFILES)	# TODO: move this into a sub-test inside test/*.sh
-	./test.sh
+	./test.sh $(TESTOPTIONS)
 
 clean:
 	@echo Cleaning Workspace...
--- a/test.sh
+++ b/test.sh
@@ -1,5 +1,88 @@
 #!/bin/bash
 
 # main test runner for heketi
+# Executes all executable scripts under tests dir
+# in sorted order.
 
-# TODO: add tests here...
+FAILURES=()
+
+vecho () {
+	if [[ "${verbose}" = "yes" ]] ; then
+		echo "$*"
+	fi
+}
+
+run_test() {
+	cmd="${1}"
+	vecho "-- Running: ${tname} --"
+	"${cmd}"
+	sts=$?
+	if [[ ${sts} -ne 0 ]]; then
+		vecho "failed ${cmd} [${sts}]"
+		FAILURES+=(${cmd})
+		if [[ "${exitfirst}" = "yes" ]]; then
+			exit 1
+		fi
+	fi
+}
+
+summary() {
+	if [[ ${#FAILURES[@]} -gt 0 ]]; then
+		echo "ERROR: failing tests:"
+		for i in ${!FAILURES[@]}; do
+			echo "  ${FAILURES[i]}"
+		done
+		exit 1
+	else
+		echo "all tests passed"
+		exit 0
+	fi
+}
+
+trap summary EXIT
+
+CLI="$(getopt -o xvh --long exitfirst,verbose,help -n $0 -- "$@")"
+eval set -- "${CLI}"
+while true ; do
+	case "$1" in
+		-x|--exitfirst)
+			exitfirst=yes
+			shift
+		;;
+		-v|--verbose)
+			verbose=yes
+			shift
+		;;
+		-h|--help)
+			echo "$0 [options]"
+			echo "  Options:"
+			echo "    -v|--verbose      Print verbose output"
+			echo "    -x|--exitfirst    Exit on first test failure"
+			echo "    -h|--help         Display help"
+			exit 0
+		;;
+		--)
+			shift
+			break
+		;;
+		*)
+			echo "unknown option" >&2
+			exit 2
+		;;
+	esac
+done
+
+SCRIPT_DIR="$(cd "$(dirname "${0}")" && pwd)"
+
+# environment vars exported for test scripts
+# (this way test scripts dont need cli parsing, we do it here)
+export HEKETI_TEST_EXITFIRST=${exitfirst}
+export HEKETI_TEST_SCRIPT_DIR="${SCRIPT_DIR}"
+
+cd "${SCRIPT_DIR}"
+for tname in $(ls tests | sort) ; do
+	tpath="./tests/${tname}"
+	if [[ ${tpath} =~ .*\.sh$ && -f ${tpath} && -x ${tpath} ]]; then
+		run_test "${tpath}"
+	fi
+done
--- a/tests/100-gotest.sh
+++ b/tests/100-gotest.sh
@@ -1,0 +1,4 @@
+#!/bin/sh
+
+GOFILES="$(go list ./... | grep -v vendor)"
+exec go test ${GOFILES}
--- a/tests/100-gotest.sh
+++ b/tests/100-gotest.sh
@@ -1,4 +1,19 @@
 #!/bin/sh
 
 GOFILES="$(go list ./... | grep -v vendor)"
-exec go test ${GOFILES}
+# no special options, exec to go test w/ all pkgs
+if [[ ${HEKETI_TEST_EXITFIRST} != "yes" ]]; then
+	exec go test ${GOFILES}
+fi
+
+# our options are set so we need to handle each go package one
+# at at time
+failed=0
+for gofile in ${GOFILES}; do
+	go test "${gofile}"
+	[ $? -ne 0 ] && ((failed+=1))
+	if [[ ${failed} -ne 0 && ${HEKETI_TEST_EXITFIRST} = "yes" ]]; then
+		exit ${failed}
+	fi
+done
+exit ${failed}
--- a/tests/001-testtest.sh
+++ b/tests/001-testtest.sh
@@ -1,0 +1,8 @@
+#!/bin/sh
+
+# This "test" exists merely to exercise the test "framework"
+
+if [ "$HEKETI_TEST_TEST" ]; then
+	exit "$HEKETI_TEST_TEST"
+fi
+exit 0


obnoxxx:
That looks really cool! Thanks @phlogistonjohn . I'll review.
Superficial comment up front: 
According to our earlier discussion, you might want to leave a little bit of number space before the 002-gotest.sh script , so some future tests can be run before it w/o needing to rename... e.t. 010-gotest.sh or 100-gotest.sh or ....

phlogistonjohn:
Hi there. I've updated the branch so that the file is now named 100-gotest.sh

test.sh
```diff
@@ -1,5 +1,67 @@
 #!/bin/bash
 
 # main test runner for heketi
+# Executes all executable scripts under tests dir
+# in sorted order.
```
purpleidea: Why does the ordering of the tests matter? In fact why not remove the `123-` prefix anyways?

purpleidea:
PS: @phlogistonjohn Love the use of tabs! w00t! (not sarcasm)

test.sh
```diff
@@ -1,5 +1,67 @@
...
 
-# TODO: add tests here...
+vecho () {
+	if [ "$verbose" = "yes" ] ; then
+		echo "$*"
+	fi
+}
+
+CLI=$(getopt -o xvh --long exitfirst,verbose,help -n $0 -- "$@")
```
obnoxxx: make sure to quote all string var extrapolations. (like `CLI="$(getopt ....` )

test.sh
phlogistonjohn: Thanks. I wanted to be able to run the faster, simpler checks first before moving on to items of more detail/slower. This has the added benefit that when used with the -x/--exitfirst option that simple checks, such as go vet for example, run before the test suite and if it fails the suite stops.

test.sh
phlogistonjohn: Will fix.

test.sh
purpleidea: @phlogistonjohn Agreed--- you should do that this way IMO: https://github.com/purpleidea/mgmt/blob/master/test.sh#L20

phlogistonjohn:
Feedback applied.

test.sh
obnoxxx: I see pros and cons:
* The downside is that you have to manually adapt the list scripts to be executed while @phlogistonjohn 's approach automatically executes them when they get added.
* The upside is that tests are not run automatically if they are placed but only if they are explicitly added to the runner script.
Not quite sure which I like best.

Also please take a look at the tests infra in https://github.com/gluster/gluster-kubernetes/tree/master/tests/simple which also has a traversal mechansim, and uses a subunit formatter for outputting the test results. Just to add to the confusion. ;-)

phlogistonjohn: Unsurprisingly, I like the current approach. :-)
FWIW I did look at the traversal and subunit stuff in there. At this time it seemed a bit big for what I was trying to accomplish in these initial changes. But I'm willing to down that road if you think it best.

obnoxxx: no need imho. you got a point - keep it simple for a start. we can still add more stuff later...

purpleidea: Actually the core scripts which don't change much (if ever) are listed in order statically in mgmt, whereas all the individual integration tests (which don't need to run in a particular order) are dumped into a folder.

One problem with the numbering system is if you ever need to "make room" or adjust things, moving them around will mess up the commit history, making it harder to see what happened.

phlogistonjohn: That's a fair point. I'd like @obnoxxx to weigh in on if I should do that or stick with my original approach.

purpleidea:
I'd actually like to NACK this temporarily, and instead offer that this be merged first:

https://github.com/heketi/heketi/pull/896

Perhaps we could build it out that way so that we can all have our cake and eat it too and also make it easier for others to add onto this when not using travis?

test.sh
```diff
@@ -1,5 +1,67 @@
 #!/bin/bash
 
 # main test runner for heketi
+# Executes all executable scripts under tests dir
+# in sorted order.
 
-# TODO: add tests here...
+vecho () {
```
purpleidea: So this whole thing is fun, but I think it should be in test/test-whatever.sh instead. And call that script from here in test.sh

test.sh
obnoxxx: We could work with symlinks:

```
001.sh --> this_is_the_real_test_script_name.sh
002.sh --> name_of_the_script_never_needs_to_change.sh
...
```

:-D

Seriously, I don't care too much. I even don't care too much about the order. You can see in gluster-kubernetes that I just picked up all files in the dir without particular order (apart from the name). In samba, in the ctdb-sub-component, we use numbered scripts, and that has not created too much of a disturbance.

So I won't NACK any approach. I'd vote to keep it as simple as possible.

obnoxxx: (we can always change it later on...)

obnoxxx:
@purpleidea so let me get this straight: Are you (in PR #896) proposing an alternative to the first patch of this PR (9622df0ba99d574d3415f727a)? I don't understand the comment with the cake. So how should this be merged with this effort (or rather, how would this effort need to be changed after merging #896)? ...  I don't want to lose all the good things of this PR. :-)

purpleidea:
@obnoxxx I'm recommending that after the PR #896 this be rebased and merged next.

obnoxxx:
> @obnoxxx I'm recommending that after the PR #896 this be rebased and merged next.

@purpleidea I read that, but I'm asking for details, you're also commenting that some of the stuff should go into a tests/test-whatever.sh, but that is actually test infrastructure that would apply mostly to the test runner. Hence I'm a little confused. :-)

purpleidea:
@obnoxxx This is in response to the discussion in the review. Read the comment here: https://github.com/heketi/heketi/pull/889#discussion_r149757842 which @phlogistonjohn agrees with. This enables that.

Put another way @phlogistonjohn 's patch is analogous to my https://github.com/purpleidea/mgmt/blob/master/test/test-shell.sh script. I'm also fine with you using that instead, but that's up to @phlogistonjohn

obnoxxx:
> @obnoxxx This is in response to the discussion in the review. Read the comment here: #889 (comment) which @phlogistonjohn agrees with. 

Well, at least did not disagree and tossed the decision over to me. ;-)

> Put another way @phlogistonjohn 's patch is analogous to my https://github.com/purpleidea/mgmt/blob/master/test/test-shell.sh script. 

I don't quite get that: why would one want to create test infra within a test-infra? This looks like what's going on here.

purpleidea:
@obnoxxx 

> I don't quite get that: why would one want to create test infra within a test-infra? This looks like what's going on here.

I don't think so. It's more like this: `test.sh` is a simple top-level shell script which statically lists a bunch of tests to run. This enables @phlogistonjohn to have his gofmt test run before the longer tests do, for example. (We like it this way in mgmt too.)

One of these "longer tests" is a shell script runner which lets you drop scripts into a folder. That's what I've written in mgmt and what @phlogistonjohn has written here. It's a useful sub-test because you can have it run later in the process AND when you add tests into his folder you don't need to patch the bash code here.

The main issue is that he's added it as a top-level thing, and I think it should be nested.

obnoxxx:
@purpleidea ok, I don't fully see it, but also don't care too much at this stage. If @phlogistonjohn is ok with rebasing/reworking on top of #896 , I will support it.

obnoxxx:
@purpleidea let me elaborate on the "I don't quite see it" part a bit: I thought your proposal was a *replacement* for @phlogistonjohn 's original, and to some extent still think. Having an additional layer seems for a shell script collector/runner seems a bit of an overkill to me. John's patch was a patch proposed for the basic infra. I think if we go with your proposal, we would want to work his code (for verbose, and exit early etc) into the base infra. Still somewhat confused by the counter-proposal. ;-)

phlogistonjohn:
Having looked over the other PR now I do have to agree that there appears to be enough overlap that it would seem to be an either/or situation. My original intent was for those type of checks like 'go vet', 'gofmt' would have low numbers like 10 & 20, then "real" tests at 100 and then we could also add scripts to run the python client api tests at 500 or something.
I understand @purpleidea's comment about file renames being annoying when looking at history. I originally hoped that using numbers with good sized gaps between them would lead to less rename churn but am willing to make changes to the other style if I get some buy-in to that idea (thus my earlier ping to @obnoxxx ).
I don't currently see a need for a two-tier approach because I always figured that the more complicated functional type tests would continue to be launched independently. I do agree with moving tasks that currently only appear in the travis yaml file to this "framework" - I have some preliminary changes to that effect in branches in my heketi clone on github (any jjm- prefixed branch name). I had hoped to hammer out our approach to the framework here and then follow up with a rebase and new PR for those bits.
Hope this all makes sense... this ended up a lot longer than I expected to write. :-)

Makefile
```diff
@@ -93,8 +92,7 @@ run: server
 	./$(APP_NAME)
 
 test: vendor glide.lock
-	go test $(GOFILES)	# TODO: move this into a sub-test inside test/*.sh
```
raghavendra-talur: I did not realize this when the original patch was merged but we need "go test" in topdir of repo. This is required to run the unit tests in apps/glusterfs/*

Makefile
```diff
@@ -93,8 +92,7 @@ run: server
...
-	./test.sh
+	./test.sh $(TESTOPTIONS)
```
raghavendra-talur: Is $TESTOPTIONS missing?

test.sh
raghavendra-talur: I like the model proposed by @phlogistonjohn here. Nothing stops us from having two tests with same NNN if we are comfortable with either running first.

Makefile
obnoxxx: Er, this does not seem to make a lot of sense to me: You can always change dirs from the actual test script. test scripts should go into tests/ , imho. :-)

Makefile
obnoxxx: I think it can be set from the caller, and it can be interpreted by test scripts? Not implemented yet?

test.sh
raghavendra-talur: To explain further, instead of considering the number at the start as a exact place in linear order, consider them as a run level. We can have 100 run levels and nothing stops us from having 10-15 tests in each run level, I can have 001-testthis.go and 001-testthat.go iff we don't care which is run first. @purpleidea @phlogistonjohn @obnoxxx agree?

Makefile
raghavendra-talur: Yup, as long as tests in directories other than tests/* are called, I am happy.

Makefile
phlogistonjohn: It is meant to be specified on the command line, for example: make test TESTOPTIONS=-vx . It is a style I've used in the past. Do you think adding a comment to that effect here would be helpful?

test.sh
phlogistonjohn: That's a yes from me.

phlogistonjohn:
Does anyone have any new thoughts on this? Last I looked both approaches have pluses and minuses and there wasn't a strong consensus which approach was preferred. At the moment, I'd like to just take one of the PRs and then we can start working from there... I'd rather we take one of them soon so we can build on it.

Should I close this PR to break the tie? What's the normal procedure in a situation like this?

purpleidea:
> Should I close this PR to break the tie? What's the normal procedure in a situation like this?


Mortal combat?? ;) /s

obnoxxx:
@phlogistonjohn @purpleidea you guys really need to work on patience. ;-)

I think @raghavendra-talur and I seem to be tending to start with this approach and possibly improve as we move along. Talur gave a strong +1 here. 

I wanted to look at some details of the test script as s final round of review.

purpleidea:
On Wed, Nov 15, 2017 at 10:38 AM, Michael Adam <notifications@github.com>
wrote:

> @phlogistonjohn <https://github.com/phlogistonjohn> @purpleidea
> <https://github.com/purpleidea> you guys really need to work on patience.
> ;-)
>
What are you talking about? Both of us are being patient.

tests/100-gotest.sh
```diff
@@ -1,4 +1,13 @@
 #!/bin/sh
 
 GOFILES="$(go list ./... | grep -v vendor)"
-exec go test ${GOFILES}
+
+failed=0
+for gofile in $GOFILES; do
+	go test "$gofile"
```
obnoxxx: This actually has an unpleasant side effect:
If you do a single call like previously
```
go test $(go list ./... | grep -v vendor)
```
then you get compilation once. Of you do `go test` for each directory separately, you get compilation each time, and the run time is roughly multiplied by 10 or so, which is quite nasty, because each to test run takes roughly 1m20s to 1m40 on my machine.

test.sh
```diff
@@ -1,5 +1,67 @@
...
+SCRIPT_DIR="$(cd "$(dirname "${0}")" && pwd)"
+
+# environment vars exported for test scripts
+# (this way test scripts dont need cli parsing, we do it here)
+export HEKETI_TEST_EXITFIRST=$exitfirst
+export HEKETI_TEST_SCRIPT_DIR="$SCRIPT_DIR"
+
+cd "$SCRIPT_DIR"
+failed=0
+for tname in $(ls tests | sort) ; do
```
obnoxxx: We should not take all files that are executable. It will for instance run backup files (e.g of vim) of scripts.
In gluster-kubernetes I run everything that matches test_*.sh Here you may want to use another pattern but almost certainly at minimum *.sh

test.sh
```diff
@@ -1,5 +1,67 @@
...
+	tpath="./tests/$tname"
+	if [ -f $tpath -a -x $tpath ]; then
+		vecho "-- Running: $tname --"
```
obnoxxx: The hunk in the loop (inside the if block) could nicely go into a function (run_test or similar)

tests/100-gotest.sh
phlogistonjohn: OK, that's a very good point. I did this in preparation to use the -coverprofile option to go test. If -coverprofile is used when specifying multiple packages it refuses to run reporting, "cannot use test profile flag with multiple packages". What we can do instead is a top-level conditional that runs the multiple packages as one command iff coverage is not requested. When coverage is requested we'll run go test in a loop this way. Sound reasonable?

test.sh
phlogistonjohn: OK, PR  #896 went with "function run-test()" do you want to use that convention here or do you prefer underscore? (I am somewhat more familiar naming shell functions with underscores but would be happy either way)

obnoxxx: I prefer underscores. looks more like real programming ;-)

obnoxxx: and i prefer a multi-line function over a 1-line construct for readability :-)

obnoxxx: we can also do the refactoring in a follow up commit if you feel exhausted. ;-)
But I would appreciate if the request for variable and conditials be folllowed:

> Generally as a style, please try using curly braces for variable expansion (like ${foobar} instead of $foobar), and double brackets for if: if [[ -n ${foobar} ]]; ...

phlogistonjohn: > we can also do the refactoring in a follow up commit if you feel exhausted. ;-)

Nah, not yet. :-)
Now that I've got your feedback I'll update the series.

phlogistonjohn:
This latest version incorporates the latest feedback from @obnoxxx as well as stealing the summary of failures idea from PR #896 but implemented slightly differently.

phlogistonjohn:
Hmm. Travis failed but it looks to be "not my fault" (tm).

obnoxxx:
Yeah, travis seems to have network issues.

obnoxxx:
@jarrpa was able to trigger a new build, not sure how that works...

jarrpa:
@obnoxxx @phlogistonjohn If you click on "Details" next to the Travis failure, it'll take you to the build on the Travis site. There should be a "Restart build" button on the right hand side. Not sure if you need to be logged in to see it. I'll hit it for now.

test.sh
```diff
@@ -1,5 +1,87 @@
...
+
+# environment vars exported for test scripts
+# (this way test scripts dont need cli parsing, we do it here)
+export HEKETI_TEST_EXITFIRST=${exitfirst}
+export HEKETI_TEST_SCRIPT_DIR="${SCRIPT_DIR}"
+
+cd "${SCRIPT_DIR}"
+for tname in $(ls tests | sort) ; do
+	tpath="./tests/${tname}"
+	if [[ ${tpath} =~ .*\.sh && -f ${tpath} && -x ${tpath} ]]; then
```
obnoxxx: this 
```
${tpath} =~ .*\.sh
```
does not work. Not sure how this `=~` works, but it still captures `tests/foobar.sh~`. Maybe just do:
```
for tname in $(ls tests/*.sh | sort) ; do
...
done
```
```

phlogistonjohn: Oh, I must has mis-tested it. Will re-do.

phlogistonjohn: The pattern should have been .*\.sh$ - can I make this change or do you have a strong preference for the globbing approach in your suggestion?

phlogistonjohn: (github ate my backslash - but I think my intent is still visible)

obnoxxx: > (github ate my backslash - but I think my intent is still visible)

I fixed it editing your comment (use backticks for code...)

obnoxxx: > The pattern should have been .*\.sh$ - can I make this change or do you have a strong preference for the globbing approach in your suggestion?

No preference. only that it works :-)

phlogistonjohn:
Updated pattern to only match files that _end_ in ".sh". Also I noticed I didn't have -v/--verbose mentioned in the help output so I added that too.

obnoxxx:
The script looks very clean now! Thanks.
Let's merge this now, to unblock other work.
If we find we want to change structure, like
adding more subdirectories to tests/, we can do
it in the next few days / weeks.

PR #896, opened 2017-11-09
tada

purpleidea:


test/test-gofmt.sh
```diff
@@ -0,0 +1,25 @@
...
+cd "${ROOT}"
+
+find_files() {
+	git ls-files | grep '\.go$'
+}
+
+GOFMT="gofmt -s"
+bad_files=$(find_files | xargs $GOFMT -l)
+if [[ -n "${bad_files}" ]]; then
+	fail_test "The following golang files are not properly formatted: ${bad_files}"
```
obnoxxx: Where does `fail_test` come from?

test.sh
```diff
@@ -1,5 +1,18 @@
 #!/bin/bash
 
 # main test runner for heketi
+failures=''
+function run-test()
+{
+	$@ || failures=$( [ -n "$failures" ] && echo "$failures\\n$@" || echo "$@" )
```
obnoxxx: Shouldn't this be quoted as follows?

```
failures="$( ... || echo "$@" )" 
```

test/test-gofmt.sh
obnoxxx: I think it's undefined...

obnoxxx:
Two more points:
* Isn't the first patch (8f93b9bb3d5588fff3a45c4) independent and could/should be part of a separate PR? Or is it a test for the (enhanced?) gofmt test?
* The commit message of the second commit  (ceae6841d9a01fa5dd5575e5) is a little weird: "Let's start checking these things." Not really helpful, could you make it more useful or omit the body? ;-)

Apart from this, it looks good to me. We need to figure out how to position this with/against PR #889.

test.sh
purpleidea: I don't think so. And it's been proven to work well in mgmt for a while now.

test/test-gofmt.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+# original version of this script from kubernetes project, under ALv2 license
+# modified version of this script from the purpleidea/mgmt project
+
+#. test/util.sh
```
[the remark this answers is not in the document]
purpleidea: a mistake

test/test-gofmt.sh
```diff
@@ -0,0 +1,25 @@
...
+
+echo running test-gofmt.sh
+set -o errexit
+set -o nounset
+set -o pipefail
+
+ROOT=$(dirname "${BASH_SOURCE}")/..
```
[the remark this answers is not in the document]
purpleidea: I don't think so.

test/test-gofmt.sh
```diff
@@ -0,0 +1,25 @@
...
+
+cd "${ROOT}"
+
+find_files() {
+	git ls-files | grep '\.go$'
```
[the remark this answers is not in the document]
purpleidea: I disagree. There's no other reliable way to detect what is actually in the repo and what is generated otherwise.

test/test-gofmt.sh
purpleidea: A bug... Fixing...

test.sh
obnoxxx: Well I think so. At least when I add shellcheck tests, this will be needed, I guess. Maybe it is never triggered the way we use it, but i made a habit of quoting all such variable assignment/expansions to keep checks happy. Well, we could still add this if some checker fails... ;-)

purpleidea:
@obnoxxx Changes fixed. New patch ready.

test/test-gofmt.sh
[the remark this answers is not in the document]
obnoxxx: well, we can still fix it later if needed. ;-)

obnoxxx:
@purpleidea , as you have witnessed, we decided to go with the original approach of #889 for now. 

It would be awesome if you could rebase this test to add the gofmt test to the tree. Thanks!

purpleidea:
@obnoxxx OK, I'll look into it.

PS: It would have been preferable to avoid the earlier patch review process and time spent on that if it wasn't the decided goal to merge this. IOW, please make architectural decisions about which patches we want *before* we go through code review.

No big deal here, just some constructive advice for the future.

purpleidea:
Patch updated.

obnoxxx:
> PS: It would have been preferable to avoid the earlier patch review process and time spent on that if it wasn't the decided goal to merge this. IOW, please make architectural decisions about which patches we want before we go through code review.

Well... for me this is reversing the facts a little... ;-) I think the confusion and additional effort was created by putting up this patch that was introduced more as a preparatory patch, so I started to review it, but upon closer inspection came across a little bit like a counter proposal for the other patch, the approach of which was essentially agreed upon. (I know, you explained that you did not mean it as a counter-proposal, but still..). So there was no a-priori goal to not merge this. I had to review the two patches, including this one, in order to understand the intentions or implications and come up with the decision. 

> No big deal here, just some constructive advice for the future.

Right, no big deal. I also guess that I probably spent more time reviewing, commenting and deciding than you spent reacting to my review comments. ;-)

obnoxxx:
> Patch updated.

Thanks!

Requests:
* The patch is incomplete (fail_test missing).
* Won't you add a number-prefix to the test script? (It's not strictly needed but would be consistent)
* Please do not do the rename `tests/` --> `test/` in this PR. If at all, this could be done separately. I see no real reason except for cosmetic consistency,  and more severely, it would currently break the centos-ci tests, which use the `tests/functional/` directory, so I'd rather leave it for now.

purpleidea:
> The patch is incomplete (fail_test missing).

Actually you're right! And you pointed out a great point-- because you didn't like my original nested approach, there's now nowhere to put a util file! If we put it in `test/` it will get run. Putting it in the top-level clutters things up!

Guess you either should have listened to me, or you're responsible for coming up for a clever fix for this :)

> Please do not do the rename tests/ --> test/ in this PR.

If you want to do things "right" then let's merge this. I have no idea why centos depends on some specific path or anything but calling `make test` -- if it does then that's a bug.

obnoxxx:
> > The patch is incomplete (fail_test missing).
> 
> Actually you're right! And you pointed out a great point-- because you didn't like my original nested approach, there's now nowhere to put a util file! If we put it in test/ it will get run. Putting it in the top-level clutters things up!

Sure. It just means the rebase is not that trivial, and you obviously need to adapt your patch a bit. 
Of course don't put it in the top level. I did not even want a test.sh in the top level in the first place, if you remember...

> Guess you either should have listened to me, or you're responsible for coming up for a clever fix for this :)

The solution is easy. We create a `tests/util` (or similar) dir and put the helper shell snippets in there.
I also want to structure the tests a bit more by creating multiple subdirs of the tests/ dir to hold the various groups of tests, and that the top test.sh runner script will operate on.


> > Please do not do the rename tests/ --> test/ in this PR.

> If you want to do things "right" then let's merge this. 

What is "right"? This is currently imho still just a cosmetic opinionated distraction. Nothing breaks by having it named `tests`.

> I have no idea why centos depends on some specific path or anything but calling make test -- if it does then that's a bug.

You're making assumptions. This is not related to centos. "centos-ci" is just the infrastructure that offers us jenkins space to run our tests... The bug is in how we (someone in the past) coded the scripts there to run our functional tests. And of course this could be improved so that the actual directory is hidden behind a make target or script invocation. But this would need to be done first before renaming directories.

obnoxxx:
> Guess you either should have listened to me, or you're responsible for coming up for a clever fix for this :)

Btw, if you don't want to adapt your test, just let me know, and I'll happily rework it the way described above. ;-)

obnoxxx:
ping @purpleidea . any updates?

purpleidea:
@obnoxxx Sure, want it rebased I guess?

obnoxxx:
@purpleidea and fixed so that it works, there were several requests and suggestions above.

purpleidea:
@obnoxxx You're right. I've tried to do too much in each single PR. I've simplified this PR with just a single commit. We can merge, then I'll open a PR for the other stuff later.

obnoxxx:
@purpleidea welll.. I asked you to delay the (entirely cosmetic) renaming of `tests/` to `test/` because there are external dependencies to that name. (Not saying this is good, and started to prepare fixing this with 0f10fb7756c93ae5a3704c490d7b3276f60987ff, but it'll need a bit, and this should not block us from other progress...)

What I asked you to do in this PR (since it was actually part of the original PR), is to rebase/adapt the go fmt script. :-)

purpleidea:
@obnoxxx I understand-- but it's silly to merge new code, to then only rename it, and you said you wanted it split. So I did the first dependency first (this commit) -- no rush, and then I'll send the next one.

If I missed something new, LMK. I did miss two tiny things. Patch updated.

obnoxxx:
@purpleidea hmmm, the tests/ directory already has a history of roughly 100 commits over 2.5 years. So I don't see how it's much more silly to rename it after a few more patches than now. And as I said, there are external changes needed to cope with the rename. Hence I don't see the requirement to let all other changes to tests/ wait until you renamed it.

Hence repeating my request: please refrain from renaming tests/ now, and do the other changes. We can again talk about the rename once the external dependencies have been removed. It does not seem important enough for me to justify blocking other changes.

purpleidea:
@obnoxxx You made a big point of trying to things right and clean and so I'm doing that. The old history doesn't get deleted in a move. I can wait on this discussion and leave this PR open, since I'm working on other things in parallel anyways. Let me know when you're out-of-tree dependencies are un-blocked.

obnoxxx:
> @obnoxxx You made a big point of trying to things right and clean and so I'm doing that.

> The old history doesn't get deleted in a move.

Right, and hence I'm not sure why you are so adamant about doing the rename *before* doing other useful changes to `tests/`.

> I can wait on this discussion and leave this PR open, since I'm working on other things in parallel anyways. Let me know when you're out-of-tree dependencies are un-blocked.

No problem. then we should just open another PR that does the "go fmt" script. As you pointed out, it's not a problem if the directory gets renamed afterwards.

obnoxxx:
@purpleidea meanwhile a go fmt test script has been added (PR #970), so that is not needed any more, and the centos-ci tests have been changed to use the test-functional make target instead of the literal "tests" directory (https://github.com/gluster/glusterfs-patch-acceptance-tests/pull/114, with three follow up PRs to fix up stuff...), so you could now re-do the patch to rename tests to test. :-)

Maybe we should this old PR and just create a new one. This history is not very helpful any more.

purpleidea:
On Thu, Feb 1, 2018 at 6:10 AM, Michael Adam <notifications@github.com> wrote:
>
> @purpleidea meanwhile a go fmt test script has been added (PR #970), so that is not needed any more, and the centos-ci tests have been changed to use the test-functional make target instead of the literal "tests" directory (gluster/glusterfs-patch-acceptance-tests#114, with three follow up PRs to fix up stuff...), so you could now re-do the patch to rename tests to test. :-)
>
> Maybe we should this old PR and just create a new one. This history is not very helpful any more.
>

Just to clarify: you *do* want the tests -> test, patch now?

Thanks,
James

obnoxxx:
@purpleidea 
> Just to clarify: you *do* want the tests -> test, patch now?

I would not request it actively, but since you advice this is the standard, 
if you want to re-do it, I would take it. :-)

purpleidea:
On Thu, Feb 1, 2018 at 6:19 AM, Michael Adam <notifications@github.com> wrote:
>
> @purpleidea
>
> Just to clarify: you do want the tests -> test, patch now?
>
> I would not request it actively, but since you advice this is the standard,
> if you want to re-do it, I would take it. :-)

I think "standard" is perhaps a bit strong, and I now realize it's a
dumb thing to discuss (my bad) but some time back I got convinced it
made more sense, and some personal annoyance with other patches led me
to possibly dig my heels in deeper about this. In any case :/

obnoxxx:
@purpleidea don't worry too much about the past. :-) If there are standards or conventions that are worth following or don't harm, then we can apply them. I checked a number of other repose, and most use "test/". Not sure it changes anything functionally  really, but if you still want to do it. No reason any more to hold it back, according to me.  @raghavendra-talur @phlogistonjohn - any opinions?

obnoxxx:
closing due to inactivity. if we want to take any of this on again, we'll start a clean new PR.

PR #898, opened 2017-11-13
Improve test coverage in utils

phlogistonjohn:
Adds test coverage for jsonutils.
Removes unused and untested stringstack.


obnoxxx:
@centos-ci ok to test

phlogistonjohn:
Applied changes to check err value after http.NewRequest is called

pkg/utils/jsonutils_test.go
```diff
@@ -0,0 +1,111 @@
...
+		"failed check", dest.Bar, "==", 88)
+}
+
+func TestGetJsonFromRequestBadJson(t *testing.T) {
+	dest := &testDest{}
+	req, err := http.NewRequest(
+		http.MethodPut,
+		"/foo/bar",
+		bytes.NewBufferString(`{"foo": `))
+	tests.Assert(t, err == nil, "err is nil", err)
```
obnoxxx: Er, now I read it again, isn't the logic reversed: The Assert will trigger when the err == nil condition is *not* met, hence, the error message should not read "err is nil" but rather reflect the fact that an error occurred. like
```
tests.Assert(t, err == nil, "Error calling http.NewRequest:", err)
```
I think several messages below have this same issue...

pkg/utils/jsonutils_test.go
```diff
@@ -0,0 +1,111 @@
...
+	resp := &http.Response{
+		Status:     "200 OK",
+		StatusCode: 200,
+		Body:       dummyCloser{bytes.NewBufferString(`{"foo": "port", "bar": 22}`)},
+	}
+	dest := &testDest{}
+	err := GetJsonFromResponse(resp, dest)
+	tests.Assert(t, err == nil, "err is nil", err)
+	tests.Assert(t, dest.Foo == "port", "failed check:", dest.Foo, "==", "port")
+	tests.Assert(t, dest.Bar == 22, "failed check", dest.Bar, "==", 22)
```
obnoxxx: Isn't this kind of message somwhat awkward to read? How about
```
tests.Assert(t, dest.Bar == 22, "Expected Bar == 22, got ", dest.Bar)
```

phlogistonjohn: Yes. These are better.

I have some ideas about reducing the boilerplate in the Assert calls, but I wanted to prototype them first, or look around to see if there are other Go libraries that do something similar to what I was thinking. Regardless of my grand plans, I will incorporate these suggestions.

phlogistonjohn:
Applied changes to improve assert details.

phlogistonjohn:
Hmm. centos-ci failed but I don't see how my changes caused the failure. I think it must have been caused by something else.

obnoxxx:
no idea. seen it before. some issue in the machine maybe...

obnoxxx:
@centos-ci retest please

obnoxxx:
@centos-ci re-test this please

phlogistonjohn:
@obnoxxx - ping. Any thoughts on this updated version?

obnoxxx:
Will merge once the year is fixed, @phlogistonjohn .

phlogistonjohn:
Oops. Copy-n-paste. Fixed now.

obnoxxx:
Ignoring centos-ci failure since only the year in the comment changed. LGTM!

phlogistonjohn:
Awesome! Thank you.

PR #900, opened 2017-11-16
eliminate the use of the stringset type

phlogistonjohn:
This will (hopefully) be my most debatable change so far. :-)

The aim is to replace and remove the stringset type from the code and corresponding
tests. Use a map and the new SortedMapKeys function as a replacement.

The stringset type has a couple of issues IMO. First, it was a custom type
that was only being used in one place. Second, it used an append-and-sort
method to add items to the set. I have seen in the Go docs a
recommendation to use maps (with dummy keys for this purpose). We
retain the end result of a sorted list of unique hosts by using a
single function call to handle the sorting of the keys.

FWIW it is also a net decrease in the number of lines.


pkg/utils/sortedstrings_test.go
```diff
@@ -81,3 +81,41 @@ func TestSortedStringsDelete(t *testing.T) {
...
+}
+
+func TestSortedMapKeys2(t *testing.T) {
+	abc := map[string]bool{
+		"11 misfits":       false,
+		" this is a test ": true,
+		"flupmpf":          true,
+		"":                 false,
+	}
+	k := SortedMapKeys(abc)
```
purpleidea: NACK on this test. I'd like to suggest a more correct way.

1) I think we need less of this tests.Assert custom thing, and more of "the standard way". Here's an example:

Instead of coding each specific test, thus believing we need many asserts to test, instead code the expected results in a big map or array and test that.

Eg:

```
	values := map[string]string{
		""          : "",
		"sOmEsTrInG": "SOMESTRING",
		"uppercase" : "UPPERCASE",
		// add more individual cases here...
	}

	for v, exp := range values { // run all the tests
		s := strings.ToUpper(v)
		if s != exp {
			t.Errorf("string `%s` did not match expected: `%s`", s, exp)
		}

		// NOTE: you could alternatively store some len value, and the values
		// shown in the map could be a big struct with expected things to test
		if len(s) != len(exp) {
			t.Errorf("string length `%d` did not match expected: `%d`", len(s), len(exp))
		}
	}
```

purpleidea:
PS: You'll note that AFAIK the golang stdlib does implement it's own test lib, and it's golang!
Here's an example of how they do it... https://golang.org/src/strings/strings_test.go#L184

I'm sad my point was rejected in https://github.com/heketi/heketi/pull/863 but maybe these more specific example will cause you all to reconsider.

pkg/utils/sortedstrings_test.go
obnoxxx: It is a little hard to understand what the above comment and suggestion really has to do with the tests that John wrote, because your example does/tests something else, and the real argument is obfuscated by the rant about `tests.Assert`, which is in fact entirely unrelated. See how much nicer your test example would read if instead of the three line boilerplate code blocks you'd have 1 line of `tests.assert(...)` instead. ;-)

And the point you're actually making applies just the same irrespective of whether you use
```
tests.Assert(t, condition, message)
```
or
```
if !condition {
    t.Errorf(message)
    // optionally return here...
}
```
:-)

purpleidea: @obnoxxx Red Hat pays me for my experience and my advice. So far you seem to not be interested in benefiting from either! You're entirely too opinionated about how this stuff should be done, for someone with far less golang experience. I'm fed up of wasting time discussing constantly instead of merging productive code.

pkg/utils/stringset.go
```diff
@@ -1,44 +0,0 @@
-//
```
raghavendra-talur: I am not sure of deleting files under pkg. Theoretically it is possible that other projects are using this.

phlogistonjohn:
@raghavendra-talur - I have a couple of thoughts about your review that I hope are worth bringing up. It took me a while to find references that state the pkg dir convention is meant for publicly accessible interfaces. Most golang resources that refer to a pkg dir refer to one that lives as a sibling to src. I found one page that stated pkg in the repo was for public interfaces. 

Next, it seems to me that heketi has made no promises about having stable interfaces outside the API and API client libs, so I feel that things in there would be use at your own risk. It also feels like if pkg is meant for stable api stuff there's a bunch of things already in there that probably should have been private implementation details.

Last, I tried searching the web for other users of Heketi's stringset and could not find any. I found some other codebases that created their own stringset files though.

If you still feel that it's not worth the risk, and that we need to keep the type, please say so. In that case I'd like to convert the type to use a map internally as that is the more language standard approach and in theory would perform better.

raghavendra-talur:
> @raghavendra-talur - I have a couple of thoughts about your review that I hope are worth bringing up. It took me a while to find references that state the pkg dir convention is meant for publicly accessible interfaces. Most golang resources that refer to a pkg dir refer to one that lives as a sibling to src. I found one page that stated pkg in the repo was for public interfaces.
> 
> Next, it seems to me that heketi has made no promises about having stable interfaces outside the API and API client libs, so I feel that things in there would be use at your own risk. It also feels like if pkg is meant for stable api stuff there's a bunch of things already in there that probably should have been private implementation details.
> 
> Last, I tried searching the web for other users of Heketi's stringset and could not find any. I found some other codebases that created their own stringset files though.
> 
> If you still feel that it's not worth the risk, and that we need to keep the type, please say so. In that case I'd like to convert the type to use a map internally as that is the more language standard approach and in theory would perform better.

It appears my understanding of pkg dir in heketi is also completely wrong. You are right. I don't have any objection to removal of the files.

pkg/utils/stringset.go
raghavendra-talur: Actually, we don't need the stringset at all. The requirement of stringset went away with https://github.com/heketi/heketi/commit/d1e62d6e57d97290310a37d738dce400e33ac4cb . This code can now simplified to be 
```
// Get all nodes in the trusted storage pool(this is gluster name for cluster)
        hosts := make([]string, 0)
	if err := db.View(func(tx *bolt.Tx) error {
		cluster, err := NewClusterEntryFromId(tx, v.Info.Cluster)
		if err != nil {
			return err
		}
		for _, nodeId := range cluster.Info.Nodes {
			node, err := NewNodeEntryFromId(tx, nodeId)
			if err != nil {
				return err
			}
                        hosts = append(hosts, node.StorageHostName())
		}
		return err
	}); err != nil {
		return err
	}
	v.Info.Mount.GlusterFS.Hosts = hosts
```

raghavendra-talur:
@phlogistonjohn See my latest comment on the commit. We don't need stringset or any replacement for it. I am sorry I took this long to realize that. Let us delete the stringset package. Also, let's not add any other util function for now.

raghavendra-talur:
@obnoxxx @phlogistonjohn To explain why we don't need stringset:

Earlier, the list of mount hosts was created by adding all unique brick hosts. As multiple bricks of a volume could reside on the same node, it was required that we find unique hosts. Commit https://github.com/heketi/heketi/commit/d1e62d6e57d97290310a37d738dce400e33ac4cb adds all nodes in a cluster to the mount hosts list. This list is already guaranteed to not have any duplicate host names. Hence, we don't need stringset.

phlogistonjohn:
That's good news and will simplifies things even further. Patch updated. It passed unit tests and functional tests locally. I ran the functional tests because I wasn't 100% sure about not needing the sort, but it worked fine without. :-)

raghavendra-talur:
@obnox if you agree with my analysis on why stringset isn't need, please merge the patch.

obnoxxx:
@centos-ci ok to test

PR #904, opened 2017-11-17
add coverage tooling to test scripts

phlogistonjohn:
These changes:
* clean up some naming
* add coverage support to the test runner and gotests script which tries to mimic the behavior of .travis-coverage script in generating a coverage summary in addition to being able to generate coverage html files
* clean up generated files
* add .gitignore rules for generated files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,6 +48,8 @@ log
 .vscode
 debug
 
+coverage
+packagecover.out
 .coverage
 .tox
 *.log
--- a/Makefile
+++ b/Makefile
@@ -97,7 +97,7 @@ test: vendor glide.lock
 clean:
 	@echo Cleaning Workspace...
 	rm -rf $(APP_NAME)
-	rm -rf dist
+	rm -rf dist coverage packagecover.out
 	@$(MAKE) -C client/cli/go clean
 
 clean_vendor:
--- a/test.sh
+++ b/test.sh
@@ -39,12 +39,41 @@ summary() {
 	fi
 }
 
-trap summary EXIT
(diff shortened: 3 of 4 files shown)


phlogistonjohn:
> * test.sh help output needs to be updated for -c|--coverage description
> * need description of what the various modes actually do

Yup. I don't know how I missed that

> * running ./test.sh -c stdout|summary does not produce any additional output or files for me - is that expected?

Summary should only produce a packagecover.out file. stdout should print a bunch of stuff to the console. I will look into it.

> * in a last patch to the series we should change .travis.yml to call this test mode an remove the .travis-coverage script

I had planned to do that in a followup PR, but I can certainly add it to this series if you think that is more appropriate.

phlogistonjohn:
There was a regression in this series from when I last rebased it. That is why you weren't getting any output! Fixing it now.

phlogistonjohn:
Current version now takes care of the first 3 bullet items. 

FWIW what I planned to do to the travis-coverage script was to remove all the logic that generated the coverage data but keep it as a stub script to submit the data to coveralls. At this time I'm assuming that it works becasue I don't know where the COVERALLS_TOKEN value is coming from. If you would like to take a different approach to that, perhaps renaming or reworking it into a single embedded command in the travis.yml file let me know.

I would like to keep that stuff separate from the test script and coverage generation at the very least.

obnoxxx:
LGTM, we can do the .travis changes in a follow-up PR.

obnoxxx:
Does not affect functional tests at all, so going without centos-ci run.

PR #906, opened 2017-11-20
show all hostnames of node in topology info

raghavendra-talur:
This shows the hostnames, both manage and storage as an array because
they are arrays. This changes cli format a little bit but we never
guaranteed cli output backward compatibility.

### DIFF
--- a/client/cli/go/cmds/topology.go
+++ b/client/cli/go/cmds/topology.go
@@ -14,6 +14,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"strings"
 
 	client "github.com/heketi/heketi/client/api/go-client"
 	"github.com/heketi/heketi/pkg/glusterfs/api"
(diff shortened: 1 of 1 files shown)


phlogistonjohn:
If I may make a suggestion I think the output would look clearer with a strings.Join(x, ", ") where x is either of the arrays in question. In addition we might want to consider pluralizing the output to "Management Hostnames" and "Storage Hostnames" since the output can now be multiple.

obnoxxx:
What would the difference be in output format for plain `x` vs `strings.Join(x, ", ")` ?

phlogistonjohn:
Based on my quick test, the array will be displayed as: [foo bar baz]
if directly stringified. If strings.Join is used in the way I mentioned it would look like: foo, bar, baz
which looks a lot cleaner to me.

raghavendra-talur:
@obnoxxx @phlogistonjohn I agree with string suggestion. 
@phlogistonjohn For plural names: you would like it to be
1. Storage Hostnames
2. Storage Hostname(s)

I don't mind leaving it as it is too.

phlogistonjohn:
I have a slight preference for the former. It's cli tool and I don't think anyone will be shocked by a single item "hostnames". The advantage of having the s is that users are hinted that there can be more than one item here. It's easier to grep for than the 2nd option with the parens.

obnoxxx:
@raghavendra-talur I'm fine with either spelling, but the plain "Hostnames" is more "greppable" as @phlogistonjohn pointed out, so I'd say go with that.

obnoxxx:
@centos-ci ok to test

raghavendra-talur:
@obnoxxx @phlogistonjohn changes done.

phlogistonjohn:
Looks good to me.
centos-ci failure seems unrelated to this change.

obnoxxx:
LGTM

PR #910, opened 2017-11-21
Adding serviceaccount for openshift deploy

ghamed:
Hi,

This is for fixing the issue with deploying heketi on openshift when using dedicated storage. The following issue is also raised https://github.com/heketi/heketi/issues/909 .

Regards,

### DIFF
--- a/client/cli/go/cmds/heketi_storage.go
+++ b/client/cli/go/cmds/heketi_storage.go
@@ -35,7 +35,7 @@ const (
 	HeketiStorageEndpointName = "heketi-storage-endpoints"
 	HeketiStorageSecretName   = "heketi-storage-secret"
 	HeketiStorageVolTagName   = "heketi-storage"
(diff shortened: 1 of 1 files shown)


raghavendra-talur:
@ghamed reviewing this. I will get back to you soon.

jarrpa:
Howdy! Thanks for the PR. :)

Unfortunately, this is not quite adequate. The name `heketi-storage-service-account` is dynamically generated and will not always be fixed. The right fix would be to detect and use the ServiceAccount from the deploy-heketi pod.

obnoxxx:
Hi, what's the situation with this PR? The last comment by @jarrpa seems to indicate that this change is not the correct fix. Do we want to update this PR for a different change or can we close it? @ghamed? Thanks!

ghamed:
Hi,

Sorry for the late reply, I agree with @jarrpa and this needs more thought to be honest.

PR #911, opened 2017-11-22
doc: add an initial contributing document

phlogistonjohn:
Add the start of a contributing doc with the goal of laying
out the basics of what a new contributor needs to know about
the project. It is heavily based on the wiki page
"Development Guide" but modified for what I hope is clarity
and organization. Some of my opinions may have leaked in too.
Having just gone through the process of getting my first
contributions into the project I hope the value judgements
I made in editing the doc are worthwhile. :-)

Initially, I planned on having a third section with
basic recommendations (dos & donts) but I couldn't think
of more than two items and they already overlapped
with previous sections (writing clear commit messages
and adding tests).

However if we want to add a section like this to the doc
I'd be happy to hear the team's thoughts and collect these
into a recommendations section.

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -11,6 +11,7 @@ volumes.
 * [Heketi API](./api/api.md)
 * [Adminstration](./admin/readme.md)
 * [Troubleshooting Guide](./troubleshooting.md)
+* [Contributing to Heketi](./contributing.md)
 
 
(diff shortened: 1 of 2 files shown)


phlogistonjohn:
It may be helpful to review the doc in its rendered form: https://github.com/phlogistonjohn/heketi/blob/jjm-doc-contrib/doc/contributing.md

PR #912, opened 2017-11-23
Add make target for functional tests

obnoxxx:
This will allow us to remove the special knowledge about how to run the functional tests from the centos-ci integration at https://github.com/gluster/glusterfs-patch-acceptance-tests


raghavendra-talur:
@obnoxxx I think you need to add test-functional to PHONY. rest of the changes look good to me.

obnoxxx:
@raghavendra-talur good point, thanks! updated.

PR #916, opened 2017-11-28
switch travis toolchain to call test suite

phlogistonjohn:
The following changes:
* Reduce the travis coverage script to only reporting coverage to coveralls service
* take the last "special" travis task and converts it into a test script (go vet)
* Convert travis.yml to use the makefile/test suite

Two things I thought about doing but did not:
* Rename/move the travis coverage script to something more discover-able
* Handle the "go fmt" step as we are already discussing that stuff in another PR.

Let me know if we feel these are things we ought to do and if we should update this PR or just do a follow-on one.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,10 +20,9 @@ matrix:
     env: COVERAGE="true"
 script:
 - go fmt $(go list ./... | grep -v vendor) | wc -l | grep 0
-- go vet $(go list ./... | grep -v vendor)
 - make
-- if [[ -z "$COVERAGE" ]]; then make test ; fi
-- if [[ -n "$COVERAGE" ]]; then bash .travis-coverage; fi
+- if [[ -z "$COVERAGE" ]]; then make test TESTOPTIONS=-v ; fi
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -19,11 +19,9 @@ matrix:
   - go: 1.8.2
     env: COVERAGE="true"
 script:
-- go fmt $(go list ./... | grep -v vendor) | wc -l | grep 0
-- go vet $(go list ./... | grep -v vendor)
```
obnoxxx: I think the `go fmt` is not yet put into a script, right?
It's in a PR #896 which would need update and rebase.

phlogistonjohn: Yup. That's what I was trying to get at with my final bullet in my description above.

obnoxxx: Yeah, but instead of removing the go fmt line in this PR/commit, we should leave it in the travis file and remove it when the go fmt script is added, right?

phlogistonjohn: Sure. Will update.

obnoxxx:
LGTM

PR #917, opened 2017-11-28
middleware: increase code coverage of jwt error condition

phlogistonjohn:
Cover a previously uncovered error condition.

### DIFF
--- a/middleware/jwt_test.go
+++ b/middleware/jwt_test.go
@@ -10,18 +10,21 @@
 package middleware
 
 import (
+	"crypto/rand"
+	"crypto/rsa"
 	"crypto/sha256"
 	"encoding/hex"
-	jwt "github.com/dgrijalva/jwt-go"
-	"github.com/gorilla/context"
-	"github.com/heketi/heketi/pkg/utils"
-	"github.com/heketi/tests"
-	"github.com/urfave/negroni"
 	"net/http"
 	"net/http/httptest"
 	"strings"
 	"testing"
 	"time"
+
+	jwt "github.com/dgrijalva/jwt-go"
+	"github.com/gorilla/context"
+	"github.com/heketi/heketi/pkg/utils"
+	"github.com/heketi/tests"
+	"github.com/urfave/negroni"
 )
 
 func TestNewJwtAuth(t *testing.T) {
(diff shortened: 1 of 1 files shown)


raghavendra-talur:
@centos-ci ok to test

middleware/jwt_test.go
```diff
@@ -526,3 +529,68 @@ func TestJwtInvalidKeys(t *testing.T) {
...
+	tests.Assert(t, err == nil, "http.NewRequest failed:", err)
+
+	// Miss 'bearer' string
+	req.Header.Set("Authorization", "bearer "+tokenString)
+	r, err := http.DefaultClient.Do(req)
+	tests.Assert(t, err == nil, "http.DefaultClient failed:", err)
+	tests.Assert(t, r.StatusCode != 0)
+	s, err := utils.GetStringFromResponse(r)
+	tests.Assert(t, strings.Contains(s, "Unexpected signing method"),
+		`expected s to contain "Unexpected signing method", got:`, s)
```
raghavendra-talur: @phlogistonjohn Is the backquote here right? I don't know what it does.

phlogistonjohn:
The backticks create a [raw string literal](https://golang.org/ref/spec#String_literals) which is generally for strings where the language doesn't interpret the contents and is used in this case as a cheap way to avoid escaping the double quotes around the expected string.

Let me know if you or the team has a strong preference against this use because I have a lot of upcoming changes that use it frequently. :-)

middleware/jwt_test.go
```diff
@@ -526,3 +529,68 @@ func TestJwtInvalidKeys(t *testing.T) {
...
+	pk, err := rsa.GenerateKey(rand.Reader, 256*2)
+	tests.Assert(t, err == nil, "rsa.GenerateKey failed:", err)
+	tokenString, err := token.SignedString(pk)
+	tests.Assert(t, err == nil, "token.SignedString failed:", err)
+
+	// Setup header
+	req, err := http.NewRequest("GET", ts.URL, nil)
+	tests.Assert(t, err == nil, "http.NewRequest failed:", err)
+
+	// Miss 'bearer' string
```
obnoxxx: I think this is a copy'n'paste of a copy'n'paste error, right? We *do* add the bearer string...

I.e. if my understanding is correct,  this comment should be removed. :-)

phlogistonjohn:
Updated with feedback incorporated.

obnoxxx:
LGTM. the centos-ci failure is unrelated.

PR #919, opened 2017-11-29
Replace api spec in markdown with openapi yaml spec

phlogistonjohn:
The following changes are a bit more of an RFC than a "please merge this now." :-)

The first change simply replaces all of the hand written bullet-list style api doc with a swagger.yml file implmenting OpenAPI spec (version 2.0). I chose 2.0 because that what most of the tooling currently supports. We can switch to 3 once the tooling in there.

I tried to make sure I covered everything that was in the markdown file plus a little extra stuff I know that the system does. However, its possible that I missed something so please let me know if I missed something.

The final change is a bit hacky but is meant to demonstrate what swagger/openapi can do vs. what we had before. Unfortunately, I spent a lot of time wrangling with how github renders content and now have a seperate copy of eveything on my own gh-pages branch. At the very least you can view things live at: https://phlogistonjohn.github.io/heketi/doc/api/latest.html

PLEASE DO NOT MERGE this as-is. If we like where I've been wandering I should fix up the urls to point at master and not my own branch. :-)

Anyway since I've spent a few days fooling with this I figured I would put it out there at the very least!


obnoxxx:
added do-not-merge because it's mentioned in the PR text

phlogistonjohn:
Closing PRs older than 2018. No expectation this will be continued. Feel free to reopen if things change.

PR #920, opened 2017-11-30
doc: point the project readme at the docs & not the wiki

phlogistonjohn:
Now that we have documentation in the doc/ dir we should
point at it rather than the (defunct) wiki content.

The downloads link to the install page didn't make any sense either
so it now points at the github releases page.


phlogistonjohn:
This may not be the best place for this thought, but what the hey:
There's a badge on the readme for a gitter.im chat thingy that seems very rarely used. The prose of the pages point users at the mailing list and freenode IRC.

I'd like to remove the gitter badge so that there's no confusion that project chat activity happens on IRC. Any thoughts? Is there any love for that application?

obnoxxx:
> This may not be the best place for this thought, but what the hey:
> There's a badge on the readme for a gitter.im chat thingy that seems very rarely used.

That's true!

> The prose of the pages point users at the mailing list and freenode IRC.

Right, even though also those are not super high traffic... Still more than the gitter.

> I'd like to remove the gitter badge so that there's no confusion that project chat activity happens on IRC. Any thoughts? Is there any love for that application?

I am not against this. You could file a PR and we can discuss it further there.

PR #925, opened 2017-12-01
extras: fix fromsource Dockerfile for current fedora (fix #923)

### LINKED ISSUE #923: Heketi dev build on docker has been failing
https://hub.docker.com/r/heketi/heketi/builds/

I was looking into something else when I noticed that the automated build of the dev docker tag has been failing for a better part of November. The logs there do not appear to show any obvious error so I'm not yet sure what is breaking the build. It does happen locally for me as well, when running docker build in extras/docker/fromsource/.

Example:
https://hub.docker.com/r/heketi/heketi/builds/br74dfat6nqnqbfap8r2bxi/

If no one immediately knows the cause I'll try to spend some time looking into it myself soon.

phlogistonjohn:
It appears that in f27 make has become a dependency of dnf.
The RUN line in the dockerfile was trying to remove make causing
dnf to exit nonzero and failing the build.

I broke up some of the multi-action lines so that its clearer
what steps are being taken. I also kept the instruction to
install make but combined it with the existing dnf install
line. Finally, I removed the -q from some of the dnf calls
as it was also supressing error output making the failure
very mysterious.

### DIFF
--- a/extras/docker/fromsource/Dockerfile
+++ b/extras/docker/fromsource/Dockerfile
@@ -13,21 +13,21 @@ ENV HEKETI_BRANCH="master"
 
 # install dependencies, build and cleanup
 RUN mkdir $BUILD_HOME $GOPATH && \
-    dnf -q -y install glide golang git && \
-    dnf -q -y install make && \
-    dnf -q -y clean all && \
+    dnf -y install glide golang git make && \
+    dnf -y clean all && \
     mkdir -p $GOPATH/src/github.com/heketi && \
     cd $GOPATH/src/github.com/heketi && \
     git clone -b $HEKETI_BRANCH https://github.com/heketi/heketi.git && \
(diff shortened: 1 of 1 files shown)


raghavendra-talur:
If I may ask for a change, please remove all instances of `-q`. I don't think having a sparse output is useful if it takes so much time for someone to debug an error case. Rest of the changes look good.

phlogistonjohn:
Great point. Patch updated.

raghavendra-talur:
closes #923

obnoxxx:
> If I may ask for a change, please remove all instances of -q. I don't think having a sparse output is useful if it takes so much time for someone to debug an error case.

Great point, but just a reminder that this should have been done in a separate patch. (Possibly in the same PR.) One patch to fix the issue, one patch to improve the code to make it easier to debug in the future. You know my sermon. ;-)

PR #927, opened 2017-12-04
add test for shellscript syntax and quality

phlogistonjohn:
These changes:
* add a checker script
* make a bunch of largely mechanical changes suggested by shellcheck
* tag some files as not to be checked (for now)
* tag some lines to ignore a particular shellcheck rule

I decided not to try and fix some of the more complex files in this PR as there's already a lot to review here. I fully intend on removing the skip markers that are also marked fixme.

### DIFF
--- a/tests/020-test-shellscripts.sh
+++ b/tests/020-test-shellscripts.sh
@@ -1,0 +1,51 @@
+#!/bin/bash
+
+# Check for shell syntax & style.
+
+test_syntax() {
+	bash -n "${1}"
+}
+
+test_shellcheck() {
+	if [[ "${SHELLCHECK}" ]]; then
+		# only look for the "flag" on comment lines
+		# without this we can match our own code :-)
+		if grep -q '^#.*HEKETI-SKIP-SHELLCHECK' "${1}"; then
+			return 0
+		fi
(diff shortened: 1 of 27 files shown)


tests/020-test-shellscripts.sh
```diff
@@ -0,0 +1,46 @@
...
+SCRIPT_DIR="$(cd "$(dirname "${0}")" && pwd)"
+
+BASE_DIR="$(cd "${SCRIPT_DIR}/.." && pwd)"
+
+if [[ -z "${SHELLCHECK}" ]]; then
+	echo "warning: could not find shellcheck ... will skip checks" >&2
+fi
+
+cd "${BASE_DIR}"
+SCRIPTS=$(find -name '*.sh' | grep -v ./vendor | sort)
```
obnoxxx: we need to also grep -v ./.git here.

phlogistonjohn:
Updated patch to exlude .git (using find itself, let me know if this approach is ok or not).
Also found and fixed two issues: one with the comment tag check causing the script to ignore itself and  an issue with flagging all files after a failed check as failed.

obnoxxx:
LGTM, thanks.

PR #936, opened 2017-12-12
executor:block: explicitly use GiB in gluster-block volume create

obnoxxx:
G is interpreted by gluster-block as GiB.
So be more explicit about it.


obnoxxx:
Related: https://github.com/kubernetes-incubator/external-storage/pull/496
Related: https://github.com/gluster/gluster-block/issues/44
Related: PR #935 

@humblec @pkalever FYI

PR #946, opened 2017-12-20
app:glusterfs:volume_entry: save bricks to db after creation

obnoxxx:
The bricks have been given a path in the CreateBricks
function, but this is not yet stored in the DB. Store it.
This will reduce the chance of seeing invalid brick entries
without a path.

Note this is a temporary fix, since we're about to change
the way this code works.


raghavendra-talur:
@centos-ci retest this please

PR #952, opened 2018-01-05
Update README.md

Yannovitch:
Fix Typo : glusterfs-daemonset instead of gluster-daemonset


Yannovitch:
Fix Typo : glusterfs-daemonset instead of gluster-daemonset

phlogistonjohn:
Good catch. Change looks OK to me.

PR #955, opened 2018-01-12
executors: predetermine the brick paths

phlogistonjohn:
Traditionally, Heketi has been letting the executor decide where
in the filesystem bricks were being mounted. In an effort to remove
state from the executor we move the logic for determining where these paths are to their own file and move state out of the executor and into the BrickRequest.

### DIFF
--- a/apps/glusterfs/brick_entry.go
+++ b/apps/glusterfs/brick_entry.go
@@ -152,6 +152,7 @@ func (b *BrickEntry) Create(db *bolt.DB, executor executors.Executor) error {
 	req.TpSize = b.TpSize
 	req.VgId = b.Info.DeviceId
 	req.PoolMetadataSize = b.PoolMetadataSize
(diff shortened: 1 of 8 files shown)


obnoxxx:
ok to test

obnoxxx:
@phlogistonjohn thanks for the patch! Generally it looks good to me.

One cosmetic comment (referring to a previous offline discussion) is that I originally meant the patch that introduces the separate utility functions module to be a refactoring in the sense that it also removes the old implementations and hence replaces the callers in the executors. Then the next patch would just move the brick path calculation outside of the executor... I don't think I will insist, but I think this would have been cleaner. For instance, the last patch now not only moves the determination of the brick path outside of the executor, it also replaces various other calls of the original functions by the new functions and removes the old functions themselves. That latter part does not technically belong to this patch.

I'll wait for the results of the test run first and then see what we'll do about it.

PR #964, opened 2018-01-17
executors: create a common msgexec package

phlogistonjohn:
Split sshexec into ssh-specific and msg based executor components, so that only the common components are shared between the kube and ssh executors.

This is a step towards making the command exec logic better separated from the transport. 

I hope to use the cleaned up msgexec package to add a bunch more unit tests in the future, as well.

### DIFF
--- a/executors/cmdexec/block_volume.go
+++ b/executors/cmdexec/block_volume.go
@@ -7,7 +7,7 @@
 // cases as published by the Free Software Foundation.
 //
 
-package sshexec
+package cmdexec
 
 import (
 	"encoding/json"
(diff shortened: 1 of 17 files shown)


obnoxxx:
@phlogistonjohn Thanks for the patch!

Before looking any further - there's a go fmt test failure...

phlogistonjohn:
Arg! Need to make that check part of the regular test suite. ;-)

obnoxxx:
@centos-ci ok to test

obnoxxx:
retest this please

phlogistonjohn:
As discussed with @obnoxxx offline we're going with the name "CmdExecutor"/"cmdexec" rather than "MsgExecutor"/"msgexec". Everything else should pretty much be the same as the previous series.

obnoxxx:
retest this please

obnoxxx:
(the old libvirt problem)

obnoxxx:
retest this please

PR #965, opened 2018-01-17
apps: change app to dynamically construct allocator

phlogistonjohn:
Before this series the app constructing the allocator once and persisting it across the lifetime of the heketi process. This series transitions the app to dynamically creating the allocator on request. A few hooks are also added to support temporarily persisting a custom allocator for tests.


obnoxxx:
@centos-ci ok to test

PR #966, opened 2018-01-18
cli: add a --durability= option for the database storage

nixpanic:
For testing and development environments where only a single storage pod
is available, the "heketidbstorage" volume can now be created without
any redundancy. The new --durability=... option defaults to "replicate",
but also accepts "none".

Signed-off-by: Niels de Vos <ndevos@redhat.com>


nixpanic:
Updated indention/alignment of an option, Travis reported a `go fmt` error.

obnoxxx:
ok to test

client/cli/go/cmds/heketi_storage.go
```diff
@@ -113,9 +118,17 @@ func createHeketiStorageVolume(c *client.Client) (*api.VolumeInfoResponse, error
...
 	req := &api.VolumeCreateRequest{}
 	req.Size = HeketiStorageVolumeSize
-	req.Durability.Type = api.DurabilityReplicate
-	req.Durability.Replicate.Replica = 3
 	req.Name = db.HeketiStorageVolumeName
+	req.Durability.Type = dt
+
+	switch dt {
+	case api.DurabilityReplicate:
+		req.Durability.Replicate.Replica = 3
```
obnoxxx: Do we want to hardcode replica 3 here? or if replica is specified also offer the possibility to specify the count?...

nixpanic: This is what there was before. If we want to change it to an other type (arbiter?), that would be a different patch/pr.

nixpanic: Ah, yes, I see what you mean. That makes sense.

Do you want that in this patch, or a follow-up one?

obnoxxx: This one would be good. Simple amend and force-push over, please.

client/cli/go/cmds/heketi_storage.go
```diff
@@ -55,6 +57,15 @@ func init() {
...
 		"container image to run this job")
+	setupHeketiStorageCommand.Flags().StringVar(&heketiStorageDurability,
+		"durability",
+		"replicate",
+		"Durability of the database storage")
+	setupHeketiStorageCommand.Flags().IntVar(&heketiStorageReplicaCount,
+		"replica",
+		3,
+		"\n\tReplica value for durability type 'replicate'."+
+			"\n\tDefault is 3")
```
obnoxxx: Here the formatting (line break / indentation) is a little inconsistent with the --durability option.
I suggest being a little more verbose and more or less copying the description of
--durability from the "volume create" command (omitting the disperse type).

client/cli/go/cmds/heketi_storage.go
```diff
@@ -113,9 +124,17 @@ func createHeketiStorageVolume(c *client.Client) (*api.VolumeInfoResponse, error
...
 	req.Name = db.HeketiStorageVolumeName
+	req.Durability.Type = dt
+
+	switch dt {
+	case api.DurabilityReplicate:
+		req.Durability.Replicate.Replica = replicaCount
+	case api.DurabilityDistributeOnly:
+		// no further options needed
+	default:
+		return nil, fmt.Errorf("Durability %s id not supported for heketi database storage", dt)
```
obnoxxx: Typo: id -> is

client/cli/go/cmds/heketi_storage.go
nixpanic: Done!

client/cli/go/cmds/heketi_storage.go
nixpanic: Done!

obnoxxx:
The code looks good to me now.
But unclear what acutally fails in the centos-ci.
It does *not* seem to be the cpu virtualization issue with libvirt/kvm.

obnoxxx:
failure unrelated according to @phlogistonjohn

PR #968, opened 2018-01-18
Add support to set clusterflags through API and cli

obnoxxx:
It is useful/needed to be able to change the cluster File and Block flags after cluster creation.

### DIFF
--- a/client/api/go-client/cluster.go
+++ b/client/api/go-client/cluster.go
@@ -62,6 +62,42 @@ func (c *Client) ClusterCreate(request *api.ClusterCreateRequest) (*api.ClusterI
 	return &cluster, nil
 }
 
+func (c *Client) ClusterSetFlags(id string, request *api.ClusterSetFlagsRequest) error {
+
+	buffer, err := json.Marshal(request)
+	if err != nil {
+		return err
+	}
+
+	// Create a request
+	req, err := http.NewRequest("POST", c.host+"/clusters/"+id+"/flags",
+		bytes.NewBuffer(buffer))
+	if err != nil {
+		return err
+	}
+	req.Header.Set("Content-Type", "application/json")
+
+	// Set token
+	err = c.setToken(req)
+	if err != nil {
+		return err
+	}
+
+	// Send request
+	r, err := c.do(req)
+	if err != nil {
+		return err
+	}
+
+	if r.StatusCode != http.StatusOK {
+		return utils.GetErrorFromResponse(r)
+	}
+
+	r.Body.Close()
+
+	return nil
+}
+
 func (c *Client) ClusterInfo(id string) (*api.ClusterInfoResponse, error) {
 
 	// Create request
@@ -78,6 +114,7 @@ func (c *Client) ClusterInfo(id string) (*api.ClusterInfoResponse, error) {
 
 	// Get info
 	r, err := c.do(req)
+	defer r.Body.Close()
 	if err != nil {
 		return nil, err
 	}
@@ -88,7 +125,6 @@ func (c *Client) ClusterInfo(id string) (*api.ClusterInfoResponse, error) {
 	// Read JSON response
 	var cluster api.ClusterInfoResponse
 	err = utils.GetJsonFromResponse(r, &cluster)
-	r.Body.Close()
 	if err != nil {
 		return nil, err
 	}
--- a/client/api/python/test/unit/test_client.py
+++ b/client/api/python/test/unit/test_client.py
@@ -42,6 +42,19 @@ class test_heketi(unittest.TestCase):
         info = c.cluster_info(cluster['id'])
         self.assertEqual(True, info == cluster)
 
+        # change cluster flags
+        cluster_setflags_req = {}
+        cluster_setflags_req['block'] = False
+        cluster_setflags_req['file'] = True
+        ok = c.cluster_setflags(cluster['id'], cluster_setflags_req)
+        self.assertTrue(ok)
+
+        # verify the cluster flags have changed
+        info = c.cluster_info(cluster['id'])
+        self.assertEqual(info['id'], cluster['id'])
+        self.assertFalse(info['block'])
+        self.assertTrue(info['file'])
+
         # Get a list of clusters
         list = c.cluster_list()
         self.assertEqual(True, len(list['clusters']) == 1)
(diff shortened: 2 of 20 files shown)


client/cli/go/cmds/cluster.go
```diff
@@ -43,10 +47,22 @@ func init() {
 			"\n\tdefault. Use '--file=false' to disable creation of file"+
 			"\n\tvolumes on this cluster.")
 
+	clusterSetFlagsCommand.Flags().StringVar(&cl_block_str, "block", "",
+		"\n\tOptional: Control the possibility of creating block volumes"+
```
phlogistonjohn: Perhaps "Control the possibility of" is more succinctly written as "Allow" or "Permit"

client/cli/go/cmds/cluster.go
```diff
@@ -96,6 +112,60 @@ var clusterCreateCommand = &cobra.Command{
...
+
+		heketi := client.NewClient(options.Url, options.User, options.Key)
+
+		info, err := heketi.ClusterInfo(clusterId)
+		if err != nil {
+			return err
+		}
+
+		req := &api.ClusterSetFlagsRequest{}
+
```
phlogistonjohn: Do we care if they run the command with neither --block=<x> nor --file-<x> ?
Right now, it would be a no-op, but a no-op that "writes" to the server. Maybe printing an error (or a warning) would be better to provide feedback to the user?

client/cli/go/cmds/cluster.go
obnoxxx: Actually, thinking about it, the text is not *that* wrong:

The `--block` option lets you specify whether or not it should be possible to create block volumes on the cluster. It lets us allow or forbid (i.e. control) that possibility. That's where I was coming from.

I still agree that it's not worded too well. Thinking how to frame it better...

client/cli/go/cmds/cluster.go
obnoxxx: done. heketi-cli cluster setflags now expects at least one of --file or --block.

obnoxxx:
Branch updated.
* based on top of new PR #971 for manpage cleanup
* added manpage section for setflags
* added setflags operation to python client and added python-test
* heketi-cli cluster setflags now requires at least one of --file or --block to be specified

TODO:
* Find out whether introduction of ClusterFlags is an api violation that we can't do (look at external consumers kubernetes in-tree and external provisioners), and ether remove the patch or complete it to cover the functional tests....

obnoxxx:
rebased over merged PR #966 which changed the manpage

obnoxxx:
Fixed up the use of Block/File flags as literal struct member in the functional test cases.

@phlogistonjohn for the question about whether to do this change at all: This does not seem to be used outside. After all, we don't do cluster operations in our provisioners -- they only do volume operations.

obnoxxx:
Updates:
* rebased on top of #971 which is rebased on latest master.
* Added tests for the api server and api client.
* Updated the API documentation for the new call.
* updated the API doc for the block/file flags in the create cluster call.

Now this should be complete.
According to discussion with @phlogistonjohn, we should be good to do the api normalization with the ClusterFlags struct.

client/api/python/test/unit/test_client.py
```diff
@@ -42,6 +42,19 @@ def test_cluster(self):
...
+        cluster_setflags_req['block'] = False
+        cluster_setflags_req['file'] = True
+        ok = c.cluster_setflags(cluster['id'], cluster_setflags_req)
+        self.assertEqual(True, ok)
+
+        # verify the cluster flags have changed
+        info = c.cluster_info(cluster['id'])
+        self.assertEqual(info['id'], cluster['id'])
+        self.assertEqual(False, info['block'])
+        self.assertEqual(True, info['file'])
```
phlogistonjohn: Small nit. Python unittest has assertTrue and assertFalse methods that are more appropriate for testing booleans. However, since the existing test code is already doing this all over the place (and some even more uglier stuff) I don't think its worth complaining too much about.

doc/api/api.md
```diff
@@ -224,6 +224,26 @@ Heketi is able to manage multiple GlusterFS clusters, each composed of a set of
 }
 ```
 
+### Set Cluster Flags
```
phlogistonjohn: This looks great. One small thing is that there's no corresponding link at the top of the document to this section.

obnoxxx: fixed, thanks

client/api/python/test/unit/test_client.py
obnoxxx: fixed, thanks

obnoxxx:
rebased on current master after #971 has been merged.

client/api/go-client/cluster.go
```diff
@@ -62,6 +62,42 @@ func (c *Client) ClusterCreate(request *api.ClusterCreateRequest) (*api.ClusterI
...
+	if err != nil {
+		return err
+	}
+
+	// OBNOX: do we need to do anything here?
+	//        go async or so like in node set or device set?
+	if r.StatusCode != http.StatusOK {
+		return utils.GetErrorFromResponse(r)
+	}
+
```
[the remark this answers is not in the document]
obnoxxx: Does that mean that `ClusterDelete()` also needs `r.Body.Close()`? And so does `ClusterList()`?
Initially I thought that it would only be needed for those calls that do read a json from the body, etc, but it does not seem to be related.

What happens if we don't close the body?

I also fail to understand the relationship with the link you posted, please explain...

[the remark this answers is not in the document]
obnoxxx: Also is it correct to close the body only after the block returning in case of failed status check?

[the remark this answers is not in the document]
obnoxxx: done

obnoxxx:
retest this please

(libvirt/kvm cpu flags error - svm)

phlogistonjohn:
@obnoxxx unless @raghavendra-talur  objects I think this is probably in a good state to merge. Please rebase onto master when you have a moment.

obnoxxx:
@phlogistonjohn agreed, thanks.

I rebased. The only other change is that I changed the `r.Body.Close()` into a `defer r.Body.Close()`. Please see also my analysis and patches in #984.

PR #969, opened 2018-01-18
apps: Split eligble clusters logic out from volume entry create

phlogistonjohn:
Changes split out cluster filtering logic, clean up and compact the filtering logic, and add check fo bad clusters.

### DIFF
--- a/apps/glusterfs/volume_entry.go
+++ b/apps/glusterfs/volume_entry.go
@@ -605,63 +605,35 @@ func eligibleClusters(db *bolt.DB, v *VolumeEntry,
 	// If the request does *not* carry the Block flag, consider
 	// only those clusters that do not carry the Block flag.
 	//
-	var candidateClusters []string
-	for _, clusterId := range possibleClusters {
-		err := db.View(func(tx *bolt.Tx) error {
+	candidateClusters := []string{}
+	err := db.View(func(tx *bolt.Tx) error {
+	CLUSTERS:
+		for _, clusterId := range possibleClusters {
 			c, err := NewClusterEntryFromId(tx, clusterId)
 			if err != nil {
 				return err
 			}
-			if v.Info.Block {
-				if c.Info.Block {
-					candidateClusters = append(candidateClusters, clusterId)
-				}
-			} else {
-				if c.Info.File {
-					candidateClusters = append(candidateClusters, clusterId)
-				}
+			switch {
+			case v.Info.Block && c.Info.Block:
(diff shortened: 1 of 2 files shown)


apps/glusterfs/volume_entry.go
```diff
@@ -615,6 +615,12 @@ func eligibleClusters(db *bolt.DB, v *VolumeEntry,
 			switch {
 			case v.Info.Block && c.Info.Block:
 			case !v.Info.Block && c.Info.File:
+			case !(c.Info.Block || c.Info.File):
+				// bad cluster config!
+				err = fmt.Errorf("Cluster %v lacks both block and file flags",
+					clusterId)
+				logger.LogError(err.Error())
+				return err
```
obnoxxx: I don't think we want this: The administrator might deliberately have mark this cluster to not create any volumes (maybe temporarily). So I suggest to drop this commit.

phlogistonjohn:
> I don't want us to treat the condition of c.Info.File == false && c.Info.Block == False as an error and bail out on it: It is a valid configuration that the admin might have deliberately chosen to disable volumes to be created on that cluster (maybe temporarily). So I propose to drop this last commit.

Do you mind if I just switch it to a info log line? I think it is worth having that extra knowledge about why the cluster was skipped. This issue bit me hard when I was working on the pr to update doc/admin/install-kubernetes.md

obnoxxx:
> > I don't want us to treat the condition of c.Info.File == false && c.Info.Block == False as an error and bail out on it: It is a valid configuration that the admin might have deliberately chosen to disable volumes to be created on that cluster (maybe temporarily). So I propose to drop this last commit.
> 
> Do you mind if I just switch it to a info log line? I think it is worth having that extra knowledge about why the cluster was skipped. This issue bit me hard when I was working on the pr to update doc/admin/install-kubernetes.md

Sure, that's a good idea.

apps/glusterfs/volume_entry.go
```diff
@@ -615,6 +615,10 @@ func eligibleClusters(db *bolt.DB, v *VolumeEntry,
 			switch {
 			case v.Info.Block && c.Info.Block:
 			case !v.Info.Block && c.Info.File:
+			case !(c.Info.Block || c.Info.File):
+				// possibly bad cluster config
+				logger.Info("Cluster %v lacks both block and file flags",
+					clusterId)
```
obnoxxx: I think you should return nil here, right?

obnoxxx:
retest this please

PR #973, opened 2018-01-22
utils: testing and test support in uuid module

phlogistonjohn:
* Separate the uuid generation/formatting from its source of randomness
* Allow other code to more easily override things with a custom (predictible) source for uuid generation
* Tests!

### DIFF
--- a/apps/glusterfs/node_entry_test.go
+++ b/apps/glusterfs/node_entry_test.go
@@ -22,11 +22,12 @@ import (
 )
 
 func createSampleNodeEntry() *NodeEntry {
+	g := utils.GenUUID()
 	req := &api.NodeAddRequest{
 		ClusterId: "123",
 		Hostnames: api.HostAddresses{
-			Manage:  []string{"manage" + utils.GenUUID()[:8]},
-			Storage: []string{"storage" + utils.GenUUID()[:8]},
+			Manage:  []string{"manage" + g[:4] + g[24:]},
+			Storage: []string{"storage" + g[:4] + g[24:]},
 		},
 		Zone: 99,
 	}
--- a/pkg/utils/uuid_test.go
+++ b/pkg/utils/uuid_test.go
@@ -31,6 +31,54 @@ func TestFakeUUID(t *testing.T) {
 	tests.Assert(t, uuid == "68656b65746968656b65746968656b65")
 }
 
+func TestNonRandomUUID(t *testing.T) {
+	s := IdSource{&NonRandom{}}
+
+	uuid := s.ReadUUID()
+	tests.Assert(t, len(uuid) == 32, "expected len(uuid) == 32, got:", len(uuid))
+	tests.Assert(t, uuid == "00000000000000000000000000000000", "got:", uuid)
+	uuid = s.ReadUUID()
+	tests.Assert(t, len(uuid) == 32, "expected len(uuid) == 32, got:", len(uuid))
+	tests.Assert(t, uuid == "00000000000000000000000000000001", "got:", uuid)
+	uuid = s.ReadUUID()
+	tests.Assert(t, len(uuid) == 32, "expected len(uuid) == 32, got:", len(uuid))
+	tests.Assert(t, uuid == "00000000000000000000000000000002", "got:", uuid)
+
+	for i := 0; i < 106; i++ {
+		s.ReadUUID()
+	}
+	uuid = s.ReadUUID()
+	tests.Assert(t, len(uuid) == 32, "expected len(uuid) == 32, got:", len(uuid))
+	tests.Assert(t, uuid == "0000000000000000000000000000006d", "got:", uuid)
+}
+
+func TestReplaceRandomness(t *testing.T) {
+	before := Randomness
+
+	n := &NonRandom{}
+	Randomness = n
+	defer func() {
+		Randomness = before
+	}()
+
+	// now we're using a non-random source. we should have predictable values
+	uuid := GenUUID()
+	tests.Assert(t, len(uuid) == 32, "expected len(uuid) == 32, got:", len(uuid))
+	tests.Assert(t, uuid == "00000000000000000000000000000000", "got:", uuid)
+
+	uuid = GenUUID()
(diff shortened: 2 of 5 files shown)


obnoxxx:
ok to test

obnoxxx:
@phlogistonjohn very interesting and clever! Need some more time to digest this.

obnoxxx:
weird vagrant bug in ci again.

retest this please

phlogistonjohn:
I've added a few changes that tweak various tests so they don't fail if you globally switch the Randomness source to the fake one. Before doing this work I got a bunch of test failures unrelated to what I was actually debugging.

pkg/utils/uuid_test.go
```diff
@@ -0,0 +1,37 @@
+//
+// Copyright (c) 2015 The heketi Authors
```
obnoxxx: 2018

pkg/utils/uuid_test.go
```diff
@@ -0,0 +1,37 @@
...
+// of the output as we are relying on an actual random source here
+func TestGenUUID(t *testing.T) {
+	uuid := GenUUID()
+	tests.Assert(t, len(uuid) == 32, "bad length", len(uuid), 32)
+}
+
+// test actual output by specifying our own source of "randomness"
+func TestFakeUUID(t *testing.T) {
+	r := strings.NewReader("heketiheketiheketi")
+	uuid := IdSource{r}.ReadUUID()
```
obnoxxx: Instead of using the `IdSource{r}.ReadUUID()`, wouldn't it be even nicer to have a function to initialize a new UUID generator `g = NewUuidGenerator(source)` and then later do `g.GenUUID()` ? Just a thought ...

pkg/utils/uuid_test.go
```diff
@@ -31,6 +31,54 @@ func TestFakeUUID(t *testing.T) {
...
+	tests.Assert(t, len(uuid) == 32, "expected len(uuid) == 32, got:", len(uuid))
+	tests.Assert(t, uuid == "00000000000000000000000000000001", "got:", uuid)
+
+	// restore the random source back to what it was before
+	Randomness = before
+
+	// uuid should NOT be the next non-random number
+	uuid = GenUUID()
+	tests.Assert(t, len(uuid) == 32, "expected len(uuid) == 32, got:", len(uuid))
+	tests.Assert(t, uuid != "00000000000000000000000000000001", "got:", uuid)
```
obnoxxx: I think the next would be `000....02` not `0000...01`

pkg/utils/uuid_test.go
```diff
@@ -31,6 +31,54 @@ func TestFakeUUID(t *testing.T) {
...
+}
+
 // NOTE: the Original GenUUID function aborts the applicaion
 // when conditions are not met. This was carried over into the
 // version with selectable random sources so we dont actually
```
obnoxxx: This comment is incomplete. And maybe misplaced?

apps/glusterfs/node_entry_test.go
```diff
@@ -22,11 +22,12 @@ import (
...
 
 func createSampleNodeEntry() *NodeEntry {
+	g := utils.GenUUID()
 	req := &api.NodeAddRequest{
 		ClusterId: "123",
 		Hostnames: api.HostAddresses{
-			Manage:  []string{"manage" + utils.GenUUID()[:8]},
-			Storage: []string{"storage" + utils.GenUUID()[:8]},
+			Manage:  []string{"manage" + g[:4] + g[24:]},
+			Storage: []string{"storage" + g[:4] + g[24:]},
```
obnoxxx: This also changes the code to use the same uuid (pieces) instead of generating a new one, for manage and storage. Intended?

pkg/utils/uuid_test.go
```diff
@@ -31,6 +31,54 @@ func TestFakeUUID(t *testing.T) {
...
+func TestReplaceRandomness(t *testing.T) {
+	before := Randomness
+
+	n := &NonRandom{}
+	Randomness = n
+	defer func() {
+		Randomness = before
+	}()
+
+	// now were using a non-random source. we should have predictible values
```
obnoxxx: predictible --> predictable

obnoxxx:
Looks good to me except for the detail points.
And the question about initialization / resetting the random source via an API vs directly accessing the struct member.

pkg/utils/uuid_test.go
phlogistonjohn: Wow. Good catch. That was a copy and paste error. I've now proved that is fails the test when I comment out the line that restores the previous version of Randomness.

pkg/utils/uuid_test.go
phlogistonjohn: Perhaps I'm missing something but I'm not sure what the difference would be?

apps/glusterfs/node_entry_test.go
phlogistonjohn: I think it should be fine because its just generating fake "hosts" for the db entires for the test. They only need to be across nodes not within the the node. 

That said while I was commuting home yesterday I was thinking that these changes were a bit ugly. What do you think about adding a "short id" generating function that is meant specifically for cases like this within the test code? Something like:

```golang
// ShortId returns a truncated id that is unique-ish but short and convenient for use
// within test code. Specify the length of this id as the first argument. If this length is
// not valid the function will panic.
func ShortId(s int) string {
  // ... stuff ...
}
```

phlogistonjohn:
Updated to fix various issues from review. I replied inline to two of your points that I want to discuss some more.

apps/glusterfs/node_entry_test.go
nixpanic: As a  side note, I would love to be able to use shortened UUIDs on the heketi-cli and have them match the actual longer ones. Just like git commands handle the hashes... But, that is probably a db feature, and not related to the UUID generation :-)

obnoxxx:
@phlogistonjohn I agree that creating a short-id function would be a good thing.
We can do that in a separate PR.

PR #975, opened 2018-01-23
Some further cleanups after introduction of CmdExec

obnoxxx:
Remove some outdated references to sshexec and refactor the common config from sshexec into cmdexec.

### DIFF
--- a/executors/cmdexec/cmdexec.go
+++ b/executors/cmdexec/cmdexec.go
@@ -16,7 +16,7 @@ import (
 )
 
 var (
-	logger = utils.NewLogger("[sshexec]", utils.LEVEL_DEBUG)
+	logger = utils.NewLogger("[cmdexec]", utils.LEVEL_DEBUG)
 )
 
 type RemoteCommandTransport interface {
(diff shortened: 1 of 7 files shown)


obnoxxx:
This is based on PR #974 .

obnoxxx:
@raghavendra-talur, what do you think about @phlogistonjohn 's point?

> I intentionally didn't change the logger prefix in my series because I wanted to discuss that change and then forgot. My primary concern was if anyone had built tooling or tests around scraping the logs. For both kubernetes and ssh executors it always reported as 'sshexec'
> 
> If its not a major concern I'm happy with the rest of these changes - but I figured the first was worth thinking about.

Do we care about keeping the logging tag as "sshexec" even if it's inconsistent, or do we want to change it for consisteny?

raghavendra-talur:
@obnoxXx I agree with the changes done. It was confusing to see sshexec in kubernetes environments. Some tooling update might be required at user's end but it would be worth it.

phlogistonjohn:
Good enough for me. We'll take it.

PR #978, opened 2018-01-23
utils: add test file for bodystring

phlogistonjohn:
Add some unit tests for functions in bodystring.go. This gets us up to around test 90% coverage on all of utils.

(From the basement of old commits)

### DIFF
--- a/pkg/utils/bodystring_test.go
+++ b/pkg/utils/bodystring_test.go
@@ -1,0 +1,119 @@
+//
+// Copyright (c) 2018 The heketi Authors
+//
+// This file is licensed to you under your choice of the GNU Lesser
+// General Public License, version 3 or any later version (LGPLv3 or
+// later), or the GNU General Public License, version 2 (GPLv2), in all
+// cases as published by the Free Software Foundation.
+//
+
+package utils
+
+import (
+	"bytes"
+	"errors"
+	"net/http"
+	"testing"
+
+	"github.com/heketi/tests"
+)
+
+type testGetStringFromResponseBody struct {
+	Data string
+}
+
+func (b testGetStringFromResponseBody) Close() error { return nil }
+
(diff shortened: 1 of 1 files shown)


pkg/utils/bodystring_test.go
```diff
@@ -0,0 +1,118 @@
...
+func TestGetStringFromResponseOK(t *testing.T) {
+	bodytext := "Hello, Heketi!"
+	resp := &http.Response{
+		Status:        "200 OK",
+		StatusCode:    200,
+		Body:          testGetStringFromResponseBody{bodytext},
+		ContentLength: int64(len(bodytext)),
+	}
+	s, err := GetStringFromResponse(resp)
+	tests.Assert(t, err == nil, "err is nil", err)
```
obnoxxx: Shouldn't the message in the Assert, after the condition, naturally state what went wrong not what was expected, or that this was expected.. e.g. `tests.Assert(t, err == nil, "Expected err == nil, got : ", err)` ?

pkg/utils/bodystring_test.go
```diff
@@ -0,0 +1,118 @@
...
+	tests.Assert(t, s == "Hello, Heketi!",
+		"failed check:", s, "==", "Hello, Heketi!")
```
obnoxxx: Similar here, better like ` ..., "Expected 'Hello, Heketi!', got '", s,'"` ?
The question really is how this message reads when the assert fails...

phlogistonjohn:
You're right. I pulled this patch out of my backlog when I was reminded of its existence during the uuid PR but I forgot to update the assert statements to the "new style". I will update those and resubmit soon.

obnoxxx:
ping @phlogistonjohn? This update should be easy. Let's get this out of our way. ;-)

phlogistonjohn:
Updated assert messages.

PR #983, opened 2018-01-24
utils: support generating "short ids"

phlogistonjohn:
This is a followup to the discussion in PR #973 .

I am adding functionality to uuid.go to generate the short ids both for the true randomness source and the Nonrandom source. Unit tests are added for both conditions.

Tests that were using truncated uuids to generate test data are then adapted to use the new ShortID function.

### DIFF
--- a/apps/glusterfs/node_entry_test.go
+++ b/apps/glusterfs/node_entry_test.go
@@ -22,12 +22,11 @@ import (
 )
 
 func createSampleNodeEntry() *NodeEntry {
-	g := utils.GenUUID()
 	req := &api.NodeAddRequest{
 		ClusterId: "123",
 		Hostnames: api.HostAddresses{
-			Manage:  []string{"manage" + g[:4] + g[24:]},
-			Storage: []string{"storage" + g[:4] + g[24:]},
+			Manage:  []string{"manage" + utils.ShortID(8)},
+			Storage: []string{"storage" + utils.ShortID(8)},
 		},
 		Zone: 99,
 	}
--- a/client/api/go-client/client_test.go
+++ b/client/api/go-client/client_test.go
@@ -132,13 +132,12 @@ func TestTopology(t *testing.T) {
 		// Create a device request
 		sg := utils.NewStatusGroup()
 		for i := 0; i < 50; i++ {
-			b := utils.GenUUID()
 			sg.Add(1)
 			go func() {
 				defer sg.Done()
 
 				deviceReq := &api.DeviceAddRequest{}
-				deviceReq.Name = "/sd" + b[:4] + b[24:]
+				deviceReq.Name = "/sd" + utils.ShortID(8)
 				deviceReq.NodeId = node.Id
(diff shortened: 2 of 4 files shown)


obnoxxx:
ok to test

phlogistonjohn:
Rebased. @obnoxxx truncating the id would have to be done differently for the hashes vs. the "fake" linearly allocated ids. Usually these tools, like git, that truncate the hashes do so by taking the first N characters, which wouldn't work for the linear ids. So I just thought it was easier to pass a buffer of size X and let the lower level populate it with whatever that layer deemed appropriate.

obnoxxx:
> Rebased. @obnoxxx truncating the id would have to be done differently for the hashes vs. the "fake" linearly allocated ids. Usually these tools, like git, that truncate the hashes do so by taking the first N characters, which wouldn't work for the linear ids. So I just thought it was easier to pass a buffer of size X and let the lower level populate it with whatever that layer deemed appropriate.

Ok, got it! But: Could you please remind me why we are using these short(ened) IDs in those few test cases (instead of standard long ones)?

phlogistonjohn:
I suppose we could always use full length IDs there. I think the gist was that for generating test data, "storage57040af1" looks a lot nicer than and is a bit easier on the eyes than "storagea224439fca97d97e357048d2d9551cb1"

On the downside it is a bit more code to maintain. I think that would be a bigger "bang for the buck" if we used the short IDs more in the test code.

obnoxxx:
> I suppose we could always use full length IDs there. I think the gist was that for generating test data, "storage57040af1" looks a lot nicer than and is a bit easier on the eyes than "storagea224439fca97d97e357048d2d9551cb1"

Yeah, but this is used in automated test code. Those eyes couldn't care less, could they?

> On the downside it is a bit more code to maintain. I think that would be a bigger "bang for the buck" if we used the short IDs more in the test code.

Right. The code is nice, but is it worth adding and maintaining it just for 4 uses in test code that would equally be happy with full-length ids... 

That was also my original confusion: I thought that we were trying to get a shortened view of the original uuids for the human consumer.

So why don't we just change those 4 places  to use full uuids?

obnoxxx:
summary of a discussion I just had with @phlogistonjohn  is that we'll put the short ID module on hold and change the 4 test places to use full IDs. One valid argument for the short IDs in the test code would be ease of debugging if needed. But currently does not seem worth the maintenance overhead. If we ever decide we need this after all, we can resurrect.

phlogistonjohn:
@obnoxxx Procedural question: should I just close out this PR and make a new one with the changes we discussed or just overwrite the changes on this branch?

obnoxxx:
> @obnoxxx Procedural question: should I just close out this PR and make a new one with the changes we discussed or just overwrite the changes on this branch?

@phlogistonjohn, I'd lean towards creating a new PR.

phlogistonjohn:
Obsoleted by pr #1005

PR #984, opened 2018-01-25
client:api: add missing r.Body.Close() calls

obnoxxx:
As per discussion with @raghavendra-talur earlier.

### DIFF
--- a/client/api/go-client/backup.go
+++ b/client/api/go-client/backup.go
@@ -37,12 +37,12 @@ func (c *Client) BackupDb(w io.Writer) error {
 	if err != nil {
 		return err
 	}
+	defer r.Body.Close()
 	if r.StatusCode != http.StatusOK {
 		return utils.GetErrorFromResponse(r)
 	}
 
 	// Read data from response
-	defer r.Body.Close()
 	_, err = io.Copy(w, r.Body)
 
 	return err
--- a/client/api/go-client/block_volume.go
+++ b/client/api/go-client/block_volume.go
@@ -47,6 +47,7 @@ func (c *Client) BlockVolumeCreate(request *api.BlockVolumeCreateRequest) (
 	if err != nil {
 		return nil, err
 	}
+	defer r.Body.Close()
 	if r.StatusCode != http.StatusAccepted {
 		return nil, utils.GetErrorFromResponse(r)
 	}
@@ -61,7 +62,6 @@ func (c *Client) BlockVolumeCreate(request *api.BlockVolumeCreateRequest) (
 
 	var blockvolume api.BlockVolumeInfoResponse
 	err = utils.GetJsonFromResponse(r, &blockvolume)
-	r.Body.Close()
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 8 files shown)


phlogistonjohn:
From a very quick glance it looks like something that would be paired with a 'defer' statement. Any reason why it isn't?

obnoxxx:
> From a very quick glance it looks like something that would be paired with a 'defer' statement. Any reason why it isn't?

Talur explained to me that it should only be done in the case of ok status. This is an inconsistent api of the http things: they only open the body if the proceed further. so the defer could only come after the if status != ok ... block. We could of course do that instead...

phlogistonjohn:
Now I'm confused. I'm looking at both bodystring.go and jsonutil.go (both under pkg/utils) and both GetJsonFromResponse and GetErrorFromResponse end up calling r.Body.Close(). So I'm not sure what path through the code fails to call r.Body.Close().

obnoxxx:
Oh, ok. I did not know that, I just did what @raghavendra-talur asked me to do. :-)

So what you're saying is that we should instead of adding r.Body.Close() calls rather remove all of them ?

phlogistonjohn:
> So what you're saying is that we should instead of adding r.Body.Close() calls rather remove all of them ?

Not yet. At the moment they seem superfluous but I don't think we've taken the time to check all the relevant code paths. If we don't add close calls I am assuming we leak fds. If having extra unnecessary redundant calls to Close is harmless than I'd say leave them for now. If extra close calls are harmless then I'm more in favor of adding defer statements so that we know all other code paths in these functions are automatically handled.

I'd like to hear what @raghavendra-talur has to say since he made the suggestion to add them.

obnoxxx:
Ok, @phlogistonjohn , @raghavendra-talur , after discussing with @phlogistonjohn and going through the code again carefully, here are the findings:

* There was an error in `GetStringFromResponse()` that led to `GetStringFromResponse()` `GetErrorFromResponse()` not calling `r.Body.Close()` in an error path.
* Looking at all the code in the API, some code paths don't call any of these functions. Hence we do need calls to `r.Body.Close()` outside of the above functions. 

Hence I
* fixed the error in `GetStringFromResponse()` in a first commit
* Based on @phlogistonjohn's suggestion, changed the API client code to call `r.Body.Close()` directly after the instantiation with `client.Do()` in all cases in a second patch.

Of course this puts the question if we need the close call in the GetFooFromBar functions at all...

It would be so nice if this language had destructors, has it? ... ;-)

obnoxxx:
Correct, missing "don't" added to commit message. thanks! @phlogistonjohn

obnoxxx:
Rebased.

obnoxxx:
> from what I understand, we can defer statements after the err is checked to be nil. Something like this:
> ```
> r, err := c.do(req)
>  	if err != nil {
>  		return nil, err
>  	}
> defer r.Body.Close()
> ```
> This needs change in all the places though.

True *iff* do can never return an initialized r (with Body) in case of err != nil...

Question is, too, whether it would harm to call r.Body.Close in other cases.

But according to https://golang.org/pkg/net/http/#Client.Do, we have

> If the returned error is nil, the Response will contain a non-nil Body which the user is expected to close

Hence the the proposed change looks correct. I will update accordingly.

obnoxxx:
Updated + rebased

PR #988, opened 2018-01-29
support database abstraction and encapsulation

phlogistonjohn:
As part of the effort to untangle db write operations and command execution I found it difficult to incrementally refactor code that might be called in two or more code-paths. I might want to be able to call a function within an existing DB transaction or standalone. These change allow us to:

a) provide clean interface based function arguments, abstracting away the exact database object in use

b) provide wrapper objects that let us build our own hooks on top of boltdb behavior without needing changes to bolt and encapsulating existing transactions in mock db interface objects


obnoxxx:
Great stuff! Reviewing...

PR #989, opened 2018-01-29
Rework allocator code to get rid of mock allocator and don't modify the allocator outside the loading function.

obnoxxx:
The mock allocator was only there still in the test code, because the test code was looking into the allocator, using its implementation.

This patchset introduces nwe methods HasDevice and HasNode to the allocator interface and changes the test code to use those. It then changes the test code to use the standard allocator instead of the mock one manually.

The mock allocator is unused now and can be removed.


obnoxxx:
Adding "do not merge" label because there's still this one TODO commit.

raghavendra-talur:
@obnoxxx Having looked at the existing code and the work required to add Node information to MockAllocator, I don't think we need to have MockAllocator.

The ring allocator is deterministic, hence can be used for all mocking purposes.

obnoxxx:
Updated to remove the mock allocator.

obnoxxx:
Also todo is now gone.

obnoxxx:
Added some more cleanups. Now AddCluster and AddDevice are only called in NewSimpleAllocatorFromDb.

PR #991, opened 2018-01-30
predetermine brick path on brick db entries

phlogistonjohn:
As a followup to PR #955, this series of changes prepares the executor to accept the brick path, rather than the brick mount-point (path). It then moves the origination of the path to when the brick entry is first created and uses that path for all subsequent operations (with an extra check left in the code intentionally since the lack of it was biting me hard during testing).

### DIFF
--- a/apps/glusterfs/brick_entry.go
+++ b/apps/glusterfs/brick_entry.go
@@ -153,7 +153,7 @@ func (b *BrickEntry) Create(db wdb.RODB, executor executors.Executor) error {
 	req.TpSize = b.TpSize
 	req.VgId = b.Info.DeviceId
 	req.PoolMetadataSize = b.PoolMetadataSize
-	req.Path = utils.BrickMountPoint(req.VgId, req.Name)
+	req.Path = utils.BrickPath(req.VgId, req.Name)
 
 	// Create brick on node
 	logger.Info("Creating brick %v", b.Info.Id)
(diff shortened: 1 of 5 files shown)


pkg/utils/paths.go
```diff
@@ -36,6 +37,27 @@ func BrickIdToThinPoolName(brickId string) string {
...
 
+// BrickPath returns the "full" path to a brick.
+func BrickPath(vgId, brickId string) string {
+	return path.Join(
+		BrickMountPoint(vgId, brickId),
+		"brick")
+}
+
+// BrickMountFromPath returns the mount point of the brick given
+// the brick's full path. This is a convenicne method that assumes
```
raghavendra-talur: spell error convenience

apps/glusterfs/brick_entry.go
```diff
@@ -196,7 +196,6 @@ func (b *BrickEntry) Destroy(db wdb.RODB, executor executors.Executor) error {
 	req.Size = b.Info.Size
 	req.TpSize = b.TpSize
 	req.VgId = b.Info.DeviceId
-	req.Path = utils.BrickMountPoint(req.VgId, req.Name)
```
raghavendra-talur: I have a suggestion. Instead, make the destroy and destroycheck functions *only* refer to what is in Brick.Info. 
This is vital for my next patch, which changes the pattern of brickPath. In setups with existing bricks, only the new bricks would have VolumeName in the brickPath.

raghavendra-talur: I take this back. Everything looks good here.

PR #992, opened 2018-01-30
cli:python: use natural, idiomatic asserts in test_client.py

obnoxxx:


obnoxxx:
@phlogistonjohn fyi. it irritated me when looking into this test script...

obnoxxx:
Thanks! :-)

PR #993, opened 2018-01-30
tests: allow reuse of VMs

raghavendra-talur:


jarrpa:
I think we have similar features in gluster-kubernetes already. On the automation side, if the user has the sahara vagrant plugin installed then the VMs can just roll back between tests. On the non-automation side, if the user runs anything other than the main run.sh script the environment is not torn down and can be manipulated or rolled back as desired.

phlogistonjohn:
Oh I really needed this already. :-) Looks good on a quick skim, I will try to look over it with more closely soon.

tests/functional/TestSmokeTest/tests/heketi_test.go
```diff
@@ -284,7 +308,7 @@ func HeketiCreateVolumeWithGid(t *testing.T) {
 		"sudo useradd writer2 -G writegroup -p'$6$WBG5yf03$3DvyE41cicXEZDW.HDeJg3S4oEoELqKWoS/n6l28vorNxhIlcBe2SLQFDhqq6.Pq'",
 		fmt.Sprintf("sudo mount -t glusterfs %v /mnt", volInfo.Mount.GlusterFS.MountPoint),
 	}
-	_, err = vagrantexec.ConnectAndExec("192.168.10.100:22", cmd, 10, true)
+	_, err = vagrantexec.ConnectAndExec(storage0+":"+portNum, cmd, 10, true)
```
phlogistonjohn: Would it be nicer to have a predefined storage0ssh variable that already includes the port in order to avoid having to repeat the string concatenation for the port ? (ditto for storage1,2,3,etc)

tests/functional/TestSmokeTest/README.md
```diff
@@ -7,7 +7,44 @@ This functional test can be used on a system with at least 8GB of RAM.
...
+$ ./run.sh
+```
+
+Output will be shown by the logs on the heketi server.
+
+### To retain VMs after test
+If you wish to retain the setup after the tests, you can use the following command instead:
+
+```bash
+$ HEKETI_TEST_NO_CLEANUP="yes" ./run.sh
```
phlogistonjohn: Generally I think it is more preferable to have variables that are true/false on their face and don't include an implicit negation. To be more specific HEKETI_TEST_CLEANUP=(yes/no)  instead of HEKETI_TEST_NO_CLEANUP=(no/yes). Understanding that in the HEKETI_TEST_CLEANUP case Yes would be the default. This goes for the other variables that also have a _NO_ in the name.

This might make the code a tad more different than before but I think it is much easier to reason about and remember.

tests/functional/lib.sh
```diff
@@ -58,19 +58,28 @@ force_cleanup_libvirt_disks() {
 }
 
 teardown() {
-    teardown_vagrant
-    force_cleanup_libvirt_disks
+    if [[ x"$HEKETI_TEST_NO_VAGRANT" != "xyes" ]]
```
phlogistonjohn: I don't think you need to do the traditional x prefix thing when using the bash double-bracket [[ test. I think the bash specific builtin is smart enough to do the comparision like a "normal" programming language.

raghavendra-talur: done

tests/functional/TestSmokeTest/tests/heketi_test.go
raghavendra-talur: Good suggestion!

tests/functional/TestSmokeTest/README.md
raghavendra-talur: Makes sense. Done.

raghavendra-talur:
@phlogistonjohn  Thanks for the review ! Agree with all the suggestions.

phlogistonjohn:
@raghavendra-talur when you have a moment, please rebase on master and ping me or this PR and I'll merge it.

tests/functional/TestSmokeTest/tests/heketi_test.go
```diff
@@ -65,6 +67,40 @@ var (
 func setupCluster(t *testing.T, numNodes int, numDisks int) {
 	tests.Assert(t, heketi != nil)
 
+	// Get ssh port first, we need it to create
+	// storageXssh variables
+	env = os.Getenv("HEKETI_TEST_STORAGEPORT")
```
phlogistonjohn: Looks like an issue here!

phlogistonjohn:
@raghavendra-talur Whoops. Actually there's an issue with the env variable that the centos-ci run found. Please fix it and then once we've got a successful run of centos-ci then I'll be happy to merge it.

phlogistonjohn:
retest this please

obnoxxx:
retest this please

obnoxxx:
retest this please

PR #994, opened 2018-01-30
sshexec: restrict sudo to required commands

### LINKED ISSUE #958: Executing commands with sudo
We have searched how heketi can execute commands with sudo and found an implementation error. In the Issue #299 the sudo was removed from the commands which is great, but the [ssh.go](https://github.com/heketi/heketi/blob/255ff7a68de06d3016379817c213a9875c66107f/pkg/utils/ssh/ssh.go) **ConnectAndExec** function has still the issue that it runs a bash with the sudo command **(it becomes root!!!)** 

```
// Execute command in a shell
command = "/bin/bash -c '" + command + "'"

// Check if we need to use sudo for the entire command
if useSudo {
command = "sudo " + command
}
```
so as you can see every command is executed with root priviliges in bash. Maybe the fault happened here #427 

The command should simply be executed with /bin/bash -c + "sudo" + command if sudo is enabled. So that you can unlock these commands with visudo under /etc/sudoers.d/heketi

```
Cmnd_Alias HEKETI_CMDS = /bin/mkdir, /bin/rmdir, /bin/mount, /bin/umount, /sbin/mkfs.xfs, /bin/sed, /sbin/lvs, /sbin/pvcreate, /sbin/pvremove, /sbin/vgcreate,/sbin/vgremove, /sbin/vgdisplay, /sbin/lvcreate, /sbin/lvremove, /usr/sbin/gluster

heketi  ALL=(ALL:ALL)NOPASSWD: HEKETI_CMDS
```

and everything will be fine!
Thanks a lot for the great work you did...

raghavendra-talur:
ssh executor starts bash with sudo if useSudo is set to true. Instead,
if only the command is executed with sudo permissions it will allow
admins to restrict sudo access using visudo.

Closes #958

### DIFF
--- a/executors/cmdexec/brick.go
+++ b/executors/cmdexec/brick.go
@@ -66,7 +66,7 @@ func (s *CmdExecutor) BrickCreate(host string,
 		fmt.Sprintf("mkfs.xfs -i size=512 -n size=8192 %v", devnode),
 
 		// Fstab
-		fmt.Sprintf("echo \"%v %v xfs rw,inode64,noatime,nouuid 1 2\" | tee -a %v > /dev/null ",
+		fmt.Sprintf("awk \"BEGIN {print \\\"%v %v xfs rw,inode64,noatime,nouuid 1 2\\\" >> \\\"%v\\\"}\"",
 			devnode,
 			mountPath,
 			s.Fstab),
--- a/executors/cmdexec/volume.go
+++ b/executors/cmdexec/volume.go
@@ -195,7 +195,9 @@ func (s *CmdExecutor) createAddBrickCommands(volume *executors.VolumeRequest,
 	}
 
 	// Add the last add-brick command to the command list
-	commands = append(commands, cmd)
+	if cmd != "" {
+		commands = append(commands, cmd)
+	}
 
 	return commands
 }
--- a/executors/cmdexec/brick_test.go
+++ b/executors/cmdexec/brick_test.go
@@ -63,10 +63,10 @@ func TestSshExecBrickCreate(t *testing.T) {
 
 			case 3:
 				tests.Assert(t,
-					cmd == "echo \"/dev/mapper/vg_xvgid-brick_id "+
+					cmd == "awk \"BEGIN {print \\\"/dev/mapper/vg_xvgid-brick_id "+
 						"/var/lib/heketi/mounts/vg_xvgid/brick_id "+
-						"xfs rw,inode64,noatime,nouuid 1 2\" | "+
-						"tee -a /my/fstab > /dev/null", cmd)
+						"xfs rw,inode64,noatime,nouuid 1 2\\\" "+
+						">> \\\"/my/fstab\\\"}\"", cmd)
 
 			case 4:
 				tests.Assert(t,
@@ -136,10 +136,10 @@ func TestSshExecBrickCreateWithGid(t *testing.T) {
 
 			case 3:
 				tests.Assert(t,
-					cmd == "echo \"/dev/mapper/vg_xvgid-brick_id "+
+					cmd == "awk \"BEGIN {print \\\"/dev/mapper/vg_xvgid-brick_id "+
 						"/var/lib/heketi/mounts/vg_xvgid/brick_id "+
-						"xfs rw,inode64,noatime,nouuid 1 2\" | "+
-						"tee -a /my/fstab > /dev/null", cmd)
+						"xfs rw,inode64,noatime,nouuid 1 2\\\" "+
+						">> \\\"/my/fstab\\\"}\"", cmd)
 
 			case 4:
 				tests.Assert(t,
@@ -220,10 +220,10 @@ func TestSshExecBrickCreateSudo(t *testing.T) {
 
 			case 3:
 				tests.Assert(t,
-					cmd == "echo \"/dev/mapper/vg_xvgid-brick_id "+
+					cmd == "awk \"BEGIN {print \\\"/dev/mapper/vg_xvgid-brick_id "+
 						"/var/lib/heketi/mounts/vg_xvgid/brick_id "+
-						"xfs rw,inode64,noatime,nouuid 1 2\" | "+
-						"tee -a /my/fstab > /dev/null", cmd)
+						"xfs rw,inode64,noatime,nouuid 1 2\\\" "+
+						">> \\\"/my/fstab\\\"}\"", cmd)
 
 			case 4:
 				tests.Assert(t,
--- a/pkg/utils/ssh/ssh.go
+++ b/pkg/utils/ssh/ssh.go
@@ -142,13 +142,11 @@ func (s *SshExec) ConnectAndExec(host string, commands []string, timeoutMinutes 
 		session.Stdout = &b
 		session.Stderr = &berr
 
-		// Execute command in a shell
-		command = "/bin/bash -c '" + command + "'"
-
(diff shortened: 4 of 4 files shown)


obnoxxx:
@phlogistonjohn writes:
> Unfortunately, I don't think we can take this until we make sure we eliminate any commands that contain pipelines in the executor code. I sent one of those such commands in IRC yesterday. We can go through the list and see how many we have and perhaps clean them up first -- then we might be able to take this.

Help me a bit, please: How does the piping relate to this change specifically? Is it because sudo and pipes does not work so well?...

phlogistonjohn:
> Help me a bit, please: How does the piping relate to this change specifically? Is it because sudo and pipes does not work so well?...

Exactly. If we construct a command like: sudo bash -c "foo x | bar y" then the sudo applies to foo & bar (as well as bash). However, bash -c "sudo foo x | bar y" will only run foo with privs. Since we have a few commands like this inside the executor code we'd need to clean those up first before taking this change.

phlogistonjohn:
Instead of making the command construction more complicated, how about using something like (only quickly tested): sed -i '$a\my test line'  /tmp/xxx
to append stuff to files?

(note that I haven't checked for other pipelines yet, those would also have to get replaced by a single command)

In the longer term I think we should try to move away from using "shell commands" to argv commands (strings vs arrays). IMO its both cleaner and safer but I haven't spent any time yet checking how well our underlying libraries would support it.

raghavendra-talur:
> Instead of making the command construction more complicated, how about using something like (only quickly tested): sed -i '$a\my test line' /tmp/xxx
> to append stuff to files?

I should have given more explanation. I tried using sed and it works for all the cases except for empty file. 
This stackoverflow question explains why https://stackoverflow.com/questions/38697974/how-to-append-line-to-empty-file-using-sed-but-not-echo. When using fstab which is in /var/lib/heketi we do start with a empty file. 

Once I gave up on sed, I started looking in existing projects on how they do it. Found this sudocommand trick in vagrant code which heavily uses echo and tee. 

Design wise it felt a little better too as sudo bool is a part of cmdconfig and not sshconfig, hence, that is where it should be interpreted. 

Implementation wise, I agree that adding an extra argument to every command built looks a little too much. I ignored it considering the above reasons. 

> 
> (note that I haven't checked for other pipelines yet, those would also have to get replaced by a single command)

As of now, this is the only command which had a pipe or a redirector. 

> 
> In the longer term I think we should try to move away from using "shell commands" to argv commands (strings vs arrays). IMO its both cleaner and safer but I haven't spent any time yet checking how well our underlying libraries would support it.

Definitely.  One good thing is that apart from this echo + tee command everything else is a non shell operation. I wish there was a command to edit fstab.

phlogistonjohn:
>  I wish there was a command to edit fstab.

Indeed. There are drawbacks to requiring an "agent" on the client side but it would avoid some problems like these.

I'm still looking for a practical short-term replacement for the pipe to tee. How about an awk command, like so? Again, not heavily tested but it does work for the empty file case for me:
```sh
$ rm -rf /tmp/yyy
$ awk 'BEGIN {print "foobar" >> "/tmp/yyy" }' 
$ awk 'BEGIN {print "foobar 2" >> "/tmp/yyy" }' 
$ cat /tmp/yyy 
foobar
foobar 2
```

obnoxxx:
To avoid using redirections and pipes with sudo, we  *could* also copy the fstab file elsewhere (using sudo), give the non-root user write access, do the modifications, and copy it back over the original location (using sudo)

raghavendra-talur:
@phlogistonjohn using awk and only change is to move "sudo" inside bash command.

executors/cmdexec/brick.go
```diff
@@ -66,7 +66,7 @@ func (s *CmdExecutor) BrickCreate(host string,
 		fmt.Sprintf("mkfs.xfs -i size=512 -n size=8192 %v", devnode),
 
 		// Fstab
-		fmt.Sprintf("echo \"%v %v xfs rw,inode64,noatime,nouuid 1 2\" | tee -a %v > /dev/null ",
+		fmt.Sprintf("awk \"BEGIN {print \\\"%v %v xfs rw,inode64,noatime,nouuid 1 2\\\" >> \\\"%v\\\"}\"",
```
obnoxxx: Not sure why awk is really used here.

Why not use echo with a regular `>> %v` instead of that weird `| tee -a %v > /dev/null` ? E.g.:

```
- fmt.Sprintf("echo \"%v %v xfs rw,inode64,noatime,nouuid 1 2\" | tee -a %v > /dev/null ",
+ fmt.Sprintf("echo \"%v %v xfs rw,inode64,noatime,nouuid 1 2\" >> %v",
```
What's the difference of this compared to the awk variant, except for the awk variant being much more cryptic, imho?

obnoxxx: Oh, I got it! The `>>` in the awk command is not a redirect, but it's basically the open mode (append) in the awk code... That makes sense to me.

obnoxxx:
I think all of @phlogistonjohn 's requests are addressed. Let him confirm.

raghavendra-talur:
For future reference:

```
# expected failure case, no sudo
rtalur@rastarpad [/tmp]$ echo "hi" > rtalurtee  
-bash: rtalurtee: Permission denied    

# sudo only before echo does not help
rtalur@rastarpad [/tmp]$ sudo echo "hi" > rtalurtee  
-bash: rtalurtee: Permission denied    

# expected failure case, no sudo
rtalur@rastarpad [/tmp]$ echo "hi" | tee -a rtalurtee >/dev/null  
tee: rtalurtee: Permission denied      

# sudo only before echo does not help
rtalur@rastarpad [/tmp]$ sudo echo "hi" | tee -a rtalurtee >/dev/null  
tee: rtalurtee: Permission denied      

# creating a privileged shell works but risky
rtalur@rastarpad [/tmp]$ sudo /bin/bash -c 'echo "hi" > rtalurtee'  

# provide sudo privilege only to tee and consume output of echo, least risky
rtalur@rastarpad [/tmp]$ echo "hi" | sudo tee -a rtalurtee >/dev/null
```

The last example could have been used to create commands but every command preparation function should be aware of useSudo setting and led to a lot of change.

Hence, we took the awk route.

executors/cmdexec/brick.go
phlogistonjohn: Yeah. It uses a similar syntax to shell for "open this file in append mode".  The only thing that worries me here is that there's a lot of backslashes in there, but there's not a blocking concern, it's just annoying to parse mentally.

executors/cmdexec/volume.go
```diff
@@ -195,7 +195,9 @@ func (s *CmdExecutor) createAddBrickCommands(volume *executors.VolumeRequest,
 	}
 
 	// Add the last add-brick command to the command list
-	commands = append(commands, cmd)
+	if cmd != "" {
+		commands = append(commands, cmd)
+	}
```
phlogistonjohn: We probably should have been doing something like this all along. :-)

executors/cmdexec/brick.go
raghavendra-talur: I learnt after many attempts that single quotes can't be escaped using backslash inside other single quotes. However, double quotes can be. This seems to be the only way to do it.

phlogistonjohn: Ok. Good to know. Thanks!

PR #999, opened 2018-02-02
apps: add initial support for hiding certain items from apis

phlogistonjohn:
Add the initial support for hiding certain items from api requests. Specifically, it adds the ability to hide volumes from the volume list, volume info, and cluster info requests. Nothing is actually hidden yet, this is just the framework.

### DIFF
--- a/apps/glusterfs/app_block_volume.go
+++ b/apps/glusterfs/app_block_volume.go
@@ -98,7 +98,7 @@ func (a *App) BlockVolumeList(w http.ResponseWriter, r *http.Request) {
 	err := a.db.View(func(tx *bolt.Tx) error {
 		var err error
 
-		list.BlockVolumes, err = BlockVolumeList(tx)
+		list.BlockVolumes, err = ListCompleteBlockVolumes(tx)
 		if err != nil {
 			return err
 		}
@@ -129,9 +129,9 @@ func (a *App) BlockVolumeInfo(w http.ResponseWriter, r *http.Request) {
 	var info *api.BlockVolumeInfoResponse
 	err := a.db.View(func(tx *bolt.Tx) error {
 		entry, err := NewBlockVolumeEntryFromId(tx, id)
-		if err == ErrNotFound {
+		if err == ErrNotFound || !entry.Visible() {
 			http.Error(w, "Id not found", http.StatusNotFound)
-			return err
+			return ErrNotFound
 		} else if err != nil {
 			http.Error(w, err.Error(), http.StatusInternalServerError)
 			return err
--- a/apps/glusterfs/app_cluster.go
+++ b/apps/glusterfs/app_cluster.go
@@ -150,6 +150,10 @@ func (a *App) ClusterInfo(w http.ResponseWriter, r *http.Request) {
 		if err != nil {
 			return err
 		}
+		err = UpdateClusterInfoComplete(tx, info)
+		if err != nil {
+			return err
+		}
 
 		return nil
 	})
--- a/apps/glusterfs/app_volume.go
+++ b/apps/glusterfs/app_volume.go
@@ -172,7 +172,7 @@ func (a *App) VolumeList(w http.ResponseWriter, r *http.Request) {
 	err := a.db.View(func(tx *bolt.Tx) error {
 		var err error
 
-		list.Volumes, err = VolumeList(tx)
+		list.Volumes, err = ListCompleteVolumes(tx)
 		if err != nil {
 			return err
 		}
@@ -201,9 +201,10 @@ func (a *App) VolumeInfo(w http.ResponseWriter, r *http.Request) {
 	var info *api.VolumeInfoResponse
 	err := a.db.View(func(tx *bolt.Tx) error {
 		entry, err := NewVolumeEntryFromId(tx, id)
-		if err == ErrNotFound {
+		if err == ErrNotFound || !entry.Visible() {
+			// treat an invisible entry like it doesn't exist
 			http.Error(w, "Id not found", http.StatusNotFound)
-			return err
+			return ErrNotFound
 		} else if err != nil {
 			http.Error(w, err.Error(), http.StatusInternalServerError)
 			return err
--- a/apps/glusterfs/listing.go
+++ b/apps/glusterfs/listing.go
@@ -1,0 +1,35 @@
+//
+// Copyright (c) 2018 The heketi Authors
+//
+// This file is licensed to you under your choice of the GNU Lesser
+// General Public License, version 3 or any later version (LGPLv3 or
+// later), or the GNU General Public License, version 2 (GPLv2), in all
+// cases as published by the Free Software Foundation.
+//
+
+package glusterfs
+
+import (
+	"github.com/boltdb/bolt"
+
+	"github.com/heketi/heketi/pkg/glusterfs/api"
+)
+
+// ListCompleteVolumes returns a list of volume ID strings for volumes
+// that are not pending.
+func ListCompleteVolumes(tx *bolt.Tx) ([]string, error) {
+	return VolumeList(tx)
+}
+
+// ListCompleteBlockVolumes returns a list of block volume ID strings for bricks
+// that are not pending.
+func ListCompleteBlockVolumes(tx *bolt.Tx) ([]string, error) {
+	return BlockVolumeList(tx)
+}
+
+// UpdateClusterInfoComplete updates the given ClusterInfoResponse object so
+// that it only contains references to complete volumes, etc.
+func UpdateClusterInfoComplete(tx *bolt.Tx, ci *api.ClusterInfoResponse) error {
+	// currently this is a no-op because we have nothing to filter out
+	return nil
+}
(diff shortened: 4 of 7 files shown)


raghavendra-talur:
I agree with most of the code as is. Just a question to make my understanding better : 

If I understand correctly, calling VolumeList() gives all volumes without filtering and using the lookupHandler version provides a filtered list. Why would you choose to implement of LookupHandler and not just add another VolumeListComplete() method on volume?

phlogistonjohn:
@raghavendra-talur : The next iteration of this code will need to filter out "pending" items. There are two ways to do this, looking at metadata inside the volume entry or looking at metadata on the pending operation entry. Doing the first means reading and parsing every volume entry, the latter means reading and parsing only the pending operations. I feel that in the average case there will be much fewer pending operations it should perform better, but to speed things up more I want to cache those pending operations once we parse them. So it seemed more appropriate to cache those entries in a common "lookup" object rather than in the Volume entry.

Secondarily, I feel like the volume entry objects are already kind of "overstuffed" with a lot of methods, whereas this new object is more single purpose.

apps/glusterfs/lookup.go
```diff
@@ -0,0 +1,45 @@
...
+
+// CompleteBricks returns a list of brick ID strings for bricks
+// that are completely created.
+func (lh *LookupHandler) CompleteBricks() ([]string, error) {
+	return BrickList(lh.tx)
+}
+
+// CompleteClusterInfo updates the given ClusterInfoResponse object so
+// that it only contains references to complete volumes, etc.
+func (lh *LookupHandler) CompleteClusterInfo(ci *api.ClusterInfoResponse) error {
```
obnoxxx: This api is slightly asymmetric compared to the other ones.
The others just mimic the VolumeList and BrickList functions,
and return lists of objects, but this one does not mimic the clusterinfo
or so but instead potentially modifies the provided info structure.

Also the nomenclature is not quite consistent:

* CompleteVolumes returns the list of complete volumes
* CompleteBrick returns the list of complete bricks
* CompleteClusterInfo does not return the complete cluster info, but instead modifies the provided cluster info to only list the complete objects... 
 
Not entirely sure what to propose here, but it somehow feels a little odd.

apps/glusterfs/app_cluster.go
```diff
@@ -150,6 +150,7 @@ func (a *App) ClusterInfo(w http.ResponseWriter, r *http.Request) {
 		if err != nil {
 			return err
 		}
+		Lookup(tx).CompleteClusterInfo(info)
```
obnoxxx: Even if it's currently just returning nil, we should still treat a possible future error condition here.

apps/glusterfs/lookup_test.go
```diff
@@ -0,0 +1,67 @@
...
+
+		// this next bit just tickles test coverage for now
+		cids, err := ClusterList(tx)
+		tests.Assert(t, err == nil, "expected err == nil, got:", err)
+		tests.Assert(t, len(cids) == 1, "expected len(cids) == 1, got:", len(cids))
+		ce, err := NewClusterEntryFromId(tx, cids[0])
+		tests.Assert(t, err == nil, "expected err == nil, got:", err)
+		info, err := ce.NewClusterInfoResponse(tx)
+		tests.Assert(t, err == nil, "expected err == nil, got:", err)
+		err = l.CompleteClusterInfo(info)
```
obnoxxx: need to check error condition

obnoxxx:
Summary of agreed changes as a result of an offline discussion:
* For now, change this to a module that exposes 2 functions (for a start):
  * `ListVolumesComplete()`
  * `UpdateClusterInfoComplete()` (or so)
* I.e. remove the currently unneded BrickList variant (we can add it later if we want to)
* hide the Lookup api inside the module (we can later chose to expose it and /or change it to a callback model if we want to)
* add the return code checking of the Update call

Later we will add `ListVolumesPending()` and corresponding functions for BlockVolumes.

phlogistonjohn:
I've reworked the patches to provide a much simpler external interface: ListComplete(Block)Volumes & UpdateClusterInfoComplete. I tweaked the word order in the first two function names to be more natural when reading it. I also added the stubs for block volumes because I had done much of that work already.

obnoxxx:
@phlogistonjohn this needs rebase and I can't update :-)

phlogistonjohn:
Rebase'd

apps/glusterfs/app_block_volume.go
```diff
@@ -141,6 +141,9 @@ func (a *App) BlockVolumeInfo(w http.ResponseWriter, r *http.Request) {
 		if err != nil {
 			http.Error(w, err.Error(), http.StatusInternalServerError)
 			return err
+		} else if !entry.Visible() {
+			http.Error(w, "Id not found", http.StatusNotFound)
+			return nil
```
raghavendra-talur: i think we need to return ErrNotFound

apps/glusterfs/app_volume.go
```diff
@@ -207,6 +207,10 @@ func (a *App) VolumeInfo(w http.ResponseWriter, r *http.Request) {
 		} else if err != nil {
 			http.Error(w, err.Error(), http.StatusInternalServerError)
 			return err
+		} else if !entry.Visible() {
+			// treat an invisible entry like it doesn't exist
+			http.Error(w, "Id not found", http.StatusNotFound)
+			return nil
```
raghavendra-talur: i think we need to return ErrNotFound

apps/glusterfs/listing.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	"github.com/boltdb/bolt"
+
+	"github.com/heketi/heketi/pkg/glusterfs/api"
+)
+
+// ListCompleteVolumes returns a list of volume ID strings for volumes
+// that are not pending.
+func ListCompleteVolumes(tx *bolt.Tx) ([]string, error) {
```
raghavendra-talur: Just a suggestion, how about ListCompletedVolumes ?

apps/glusterfs/listing.go
```diff
@@ -0,0 +1,35 @@
...
+	return VolumeList(tx)
+}
+
+// ListCompleteBlockVolumes returns a list of block volume ID strings for bricks
+// that are not pending.
+func ListCompleteBlockVolumes(tx *bolt.Tx) ([]string, error) {
```
raghavendra-talur: same here, how about ListCompletedBlockVolumes ?

apps/glusterfs/listing.go
```diff
@@ -0,0 +1,35 @@
...
+	return BlockVolumeList(tx)
+}
+
+// CompleteClusterInfoComplete updates the given ClusterInfoResponse object so
```
raghavendra-talur: I guess the comment needs to be updated.

apps/glusterfs/app_block_volume.go
phlogistonjohn: I will double check it.

apps/glusterfs/listing.go
phlogistonjohn: That's fine too. I'll let @obnoxxx chime in with which version is preferred as a tiebreaker.

apps/glusterfs/listing.go
phlogistonjohn: Yeah, that just got messed up. That's neither the original name nor the current name. :-)

obnoxxx: UpdateClusterInfoComplete it should be

apps/glusterfs/app_block_volume.go
obnoxxx: 1.) it should treat the !visible condition just like not finding the entry. hence we need to return ErrNotFound
2.) In this case (as opposed to the volumeinfo function), the location is wrong. needs to in the block line 132 ff

apps/glusterfs/app_volume.go
obnoxxx: and merge with err == ErrNotFound condition?!

apps/glusterfs/listing.go
obnoxxx: imho , let's stick with complete (1 char shorter...) ;-)

apps/glusterfs/listing.go
obnoxxx: same, let's stick with ListCompleteBockVolumes, imho.

obnoxxx:
@phlogistonjohn updated ... scary, that actually worked!

obnoxxx:
@phlogistonjohn btw, looking at and working on the code, the callback approach came to me again: we could also pass a filter function to `NewVolumeEntryFromId` and friends, or just create a filtering wrapper, like `NewCompleteVolumeEntryFromId` or so, that would include the Visible check, or so ... taking the need from the callers to do the check... 

well, something for the future.

obnoxxx:
For now, I think we've addressed all review comments

obnoxxx:
LGTM. requests have been addressed.

phlogistonjohn:
The updated changes LGTM

PR #1003, opened 2018-02-05
tests: add some unit tests

phlogistonjohn:
These tests check for error conditions that I created when I was doing preliminary untangle work.

PR #1004, opened 2018-02-06
travis: try to let the travis run pass again

obnoxxx:
Condense the coverage vs no-coverage test runs into
one if-else statement.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,5 @@ matrix:
     env: COVERAGE="true"
 script:
 - make
-- if [[ -z "$COVERAGE" ]]; then make test TESTOPTIONS=-v ; fi
-- if [[ -n "$COVERAGE" ]]; then make test TESTOPTIONS=-vcstdout && bash .travis-coverage; fi
+- if [[ -z "$COVERAGE" ]]; then make test TESTOPTIONS=-v ; else make test TESTOPTIONS=-vcstdout && bash .travis-coverage; fi
 - make release


obnoxxx:
an attempt...

obnoxxx:
retest this please

obnoxxx:
@phlogistonjohn ok, now I see that what I thought needs fixing does not need fixing at all. Did I just misread the test output? We should really stop the test immediately when something fails, and clearly print out what failed...

obnoxxx:
(This patch might still make the output less misleading...)

phlogistonjohn:
> @phlogistonjohn ok, now I see that what I thought needs fixing does not need fixing at all. Did I just misread the test output? 

Possibly. It is kind of buried amongst other stuff (see below)

> We should really stop the test immediately when something fails, 

I somewhat agree and when I run tests manually I am almost always using the -x switch which I added to the top level runner to stop the tests on first failure. However, when the tests are run asynchronously (like thru travis) it makes sense to run the entire suite of tests and then print out all the different failures in case you hit multiple issues.

> and clearly print out what failed...

I strongly agree. The problem here was that the failure in pr #1003 is printed in between two travis commands (at the end of) make test and before make release. The make release command generates a lot of output and so you have to scroll back quite a bit to see the make test failures.

There's a couple things we could do. We could change the ordering of the commands so that make release comes before make test. We could also change the output of the test runner script so that it is bigger/bolder/more obvious (more colorful?). 


All that said I do like your change. Unless you or someone else NAKs it I'll probably merge it later.

PR #1021, opened 2018-02-13
cli/heketi_storage: show usage only when asked

raghavendra-talur:


obnoxxx:
@raghavendra-talur why the do-not-merge label?

raghavendra-talur:
@obnoxxx I was not sure if this was required, hence the label. I have removed the label, good to merge.

raghavendra-talur:
>  I wonder if in the future we should add our own wrapper function for AddCommand that does some of this common boilerplate for us.

@phlogistonjohn yes, that would be useful.

PR #1022, opened 2018-02-15
First sweep of untangle patches.

obnoxxx:
Preparations for introducing pending operations.

### DIFF
--- a/apps/glusterfs/volume_entry.go
+++ b/apps/glusterfs/volume_entry.go
@@ -324,6 +324,16 @@ func (v *VolumeEntry) createOneShot(db wdb.DB,
 		return ErrNoSpace
 	}
 
+	err = v.updateMountInfo(db)
+	if err != nil {
+		return err
+	}
+
+	err = v.saveCreateVolume(db, allocator)
+	if err != nil {
+		return err
+	}
+
 	// Make sure to clean up bricks on error
 	defer func() {
 		if e != nil {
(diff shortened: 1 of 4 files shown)


obnoxxx:
cleaned-up part of untangle work for volume.

obnoxxx:
@phlogistonjohn wrote:
> LGTM. But of course I wrote most of it. ;-)

small correction: all of it (since I removed the other patches from the top for now) ;-)

apps/glusterfs/volume_entry.go
```diff
@@ -292,88 +261,126 @@ func (v *VolumeEntry) Create(db wdb.DB,
...
+	// Create the bricks on the nodes
+	e = CreateBricks(db, executor, brick_entries)
+	if e != nil {
+		return
 	}
 
-	// Destroy volume on failure
-	defer func() {
-		if e != nil {
-			v.Destroy(db, executor)
```
raghavendra-talur: This is the v.Destroy I was referring to in the other comment.

apps/glusterfs/volume_entry_create.go
```diff
@@ -37,6 +37,11 @@ func (v *VolumeEntry) createVolume(db wdb.RODB,
 	if _, err := executor.VolumeCreate(host, vr); err != nil {
 		return err
 	}
```
raghavendra-talur: This is possibly where we call v.Destroy.

PR #1027, opened 2018-02-16
apps: add pending operations support

phlogistonjohn:
Depends on pr #1026 atm.

This is a cleaned up version of what's in my untangle3a branch with regards to defining the data structures and db entries for pending operations.

I have tests for the patch "add basic support for db import/export of pending op entries" but those tests depend on implementation of volume create operation code in the other branch so I left it out for now.

### DIFF
--- a/apps/glusterfs/pendingop.go
+++ b/apps/glusterfs/pendingop.go
@@ -1,0 +1,87 @@
+//
+// Copyright (c) 2018 The heketi Authors
+//
+// This file is licensed to you under your choice of the GNU Lesser
+// General Public License, version 3 or any later version (LGPLv3 or
+// later), or the GNU General Public License, version 2 (GPLv2), in all
+// cases as published by the Free Software Foundation.
+//
+
+package glusterfs
+
+import (
+	"fmt"
+)
+
+// The pendingop.go file defines the basic structures needed to track
+// life-cycle of database entries w/in Heketi. There are generally two
+// levels of objects which we track: the pending operation a higher-level
+// operation that should roughly correspond to an action in the Heketi API,
+// and lower level change tracking where individual object creates, deletes,
+// and modifications are logged. Note that objects such as bricks and
+// volumes are still managed in their own buckets. The pending operation
+// metadata tracks the IDs of these objects and these objects have references
+// back to their associated pending operations (via IDs).
+
+// PendingOperationType identifies what kind of high-level operation a
+// PendingOperation will be.
+type PendingOperationType int
+
+const (
+	OperationUnknown PendingOperationType = iota
+	OperationCreateVolume
+	OperationDeleteVolume
+	OperationExpandVolume
+	OperationCreateBlockVolume
+	OperationDeleteBlockVolume
+)
+
+// PendingChangeType identifies what kind of lower-level new item or change
+// is being made to the system as part of a higher-level pending operation.
+type PendingChangeType int
+
+const (
+	OpUnknown PendingChangeType = iota
+	OpAddBrick
+	OpAddVolume
+	OpDeleteBrick
+	OpDeleteVolume
+	OpExpandVolume
+	OpAddBlockVolume
+	OpDeleteBlockVolume
+)
+
+// PendingOperationAction tracks individual changes to entries within the
+// heketi db. It consists of a required change type and (heketi uuid) id,
+// as well as an optional delta object for extra metadata.
+type PendingOperationAction struct {
+	Change PendingChangeType
+	Id     string
+	Delta  interface{}
+}
+
+// PendingItem encapsulates the common pending item ID field.
+type PendingItem struct {
+	Id string
+}
+
+// PendingOperation tracks higher-level changes to the heketi system, such
+// as volume creation or deletion.
+type PendingOperation struct {
+	PendingItem
+	Timestamp int64
+	Type      PendingOperationType
+	Actions   []PendingOperationAction
+}
+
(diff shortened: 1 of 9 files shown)


apps/glusterfs/pendingop_entry.go
```diff
@@ -0,0 +1,265 @@
...
+	"github.com/boltdb/bolt"
+	wdb "github.com/heketi/heketi/pkg/db"
+	"github.com/heketi/heketi/pkg/utils"
+	"github.com/lpabon/godbc"
+)
+
+const (
+	NEW_ID                    = ""
+	BOLTDB_BUCKET_PENDING_OPS = "PENDING_OPERATIONS"
+	DB_HAS_PENDING_OPS        = "PENDING_OPERATIONS"
```
raghavendra-talur: 1. I suggest DB_HAS_PENDING_OPS_BUCKET to make it less alarming. Having pending ops in closed Db should alarm.
2. Would like if const was same as the string for dbattributes.

So suggestion to change from 
DB_HAS_PENDING_OPS        = "PENDING_OPERATIONS"
to
DB_HAS_PENDING_OPS_BUCKET        = "DB_HAS_PENDING_OPS_BUCKET"

apps/glusterfs/pendingop_entry.go
```diff
@@ -0,0 +1,265 @@
...
+	godbc.Require(id != "")
+
+	entry := &PendingOperationEntry{}
+	err := EntryLoad(tx, entry, id)
+	if err != nil {
+		return nil, err
+	}
+
+	if entry.Actions == nil {
+		entry.Actions = []PendingOperationAction{}
```
raghavendra-talur: is this required?

apps/glusterfs/pendingop_entry.go
phlogistonjohn: Updated.

apps/glusterfs/pendingop_entry.go
phlogistonjohn: Not sure really. This pattern is fairly frequent in other *_entry.go files so I figured better safe than sorry in that regard and copied that pattern.

apps/glusterfs/pendingop.go
```diff
@@ -0,0 +1,87 @@
...
+	OpDeleteBlockVolume
+)
+
+// PendingOperationAction tracks individual changes to entries within the
+// heketi db. It consists of a required change type and (heketi uuid) id,
+// as well as an optional delta object for extra metadata.
+type PendingOperationAction struct {
+	Change PendingChangeType
+	Id     string
+	Delta  interface{}
```
raghavendra-talur: wondering if we need parent operation ID here. Just for debugging, logging and Db export cases.

apps/glusterfs/pendingop_entry.go
raghavendra-talur: makes sense..

apps/glusterfs/pendingop.go
phlogistonjohn: I get you, but I don't think its needed because these are only ever found "nestled" within the PendingOperationEntry. If we started embedding or using them outside the pending operation I might agree, but I'd skip it for now.

apps/glusterfs/pendingop_entry.go
```diff
@@ -0,0 +1,265 @@
...
+	p.recordChange(OpAddVolume, v.Info.Id)
+	p.Type = OperationCreateVolume
+	// link back from "temporary" object to op
+	v.Pending.Id = p.Id
+}
+
+// FinalizeVolume removes tracking metadata from the volume entry.
+// This means that the volume is no longer pending.
+func (p *PendingOperationEntry) FinalizeVolume(v *VolumeEntry) {
+	v.Pending.Id = ""
```
raghavendra-talur: Spent some time thinking if it would help to have a DONE_ID like you have NEW_ID. Looks like there is not much use, as the objects start with a UUID. Hence, there are only two states "" denotes done, valid ID denotes in_progress. Right?

apps/glusterfs/pendingop.go
raghavendra-talur: Agreed 👍

apps/glusterfs/pendingop_entry.go
phlogistonjohn: Yes, in the case of NEW_ID I was trying to make explicit that passing an empty string into the New... function would make it allocate a new "guid".  In the case of setting x.Pending.Id = "" that's more standard (IMO) golang practice of empty string is the string type's zero value.

raghavendra-talur: +1 for the idea of integrating with rest op id, with your current model it can be achieved in upcoming releases.

apps/glusterfs/pendingop_entry.go
```diff
@@ -0,0 +1,265 @@
...
+)
+
+const (
+	NEW_ID                    = ""
+	BOLTDB_BUCKET_PENDING_OPS = "PENDING_OPERATIONS"
+	DB_HAS_PENDING_OPS_BUCKET = "DB_HAS_PENDING_OPS_BUCKET"
+)
+
+var (
+	// support unit test dep. injection for custom timestemps
```
obnoxxx: timest*a*mps

phlogistonjohn:
Typo fixed. I have a lot more tests incoming but they depend on operations and other various things and are hard to shuffle around atm. Hopefully that's not too much of an issue.

PR #1030, opened 2018-02-17
delete empty path bricks

raghavendra-talur:


apps/glusterfs/device_entry.go
```diff
@@ -463,3 +464,33 @@ func (d *DeviceEntry) Remove(db wdb.DB,
...
+		fmt.Fprintf(os.Stdout, "before calling brick entry %v", id)
+		brick, err := NewBrickEntryFromId(tx, id)
+		if err != nil {
+			return err
+		}
+		fmt.Fprintf(os.Stdout, "before going into delete %v", id)
+		if brick.Info.Path == "" {
+			d.StorageFree(brick.TotalSize())
+			d.BrickDelete(brick.Info.Id)
+			err = d.Save(tx)
```
raghavendra-talur: should not be saving here. 
Should cache and save at the end of loop or on err

phlogistonjohn: Sounds good. Since you are planning on doing that it might be cool to also have a dry-run mode that only prints what its going to do. I don't think that would be too much extra effort.

main.go
```diff
@@ -118,22 +119,47 @@ var exportdbCmd = &cobra.Command{
 	},
 }
 
+var deleteBricksWithEmptyPath = &cobra.Command{
+	Use:     "delete-bricks-with-empty-path",
+	Short:   "removes brick entries from db that have empty path",
+	Long:    "removes brick entries from db that have empty path",
+	Example: "heketi db delete-brick-with-empty-path --dbfile=/db/file/path/",
```
phlogistonjohn: Typo: the word is -brick- here but is -bricks- in the Use line above.

apps/glusterfs/dbcommon.go
```diff
@@ -69,3 +71,87 @@ func initializeBuckets(tx *bolt.Tx) error {
 
 	return nil
 }
+
+func DeleteBricksWithEmptyPath(db *bolt.DB, all bool, clusterIDs []string, nodeIDs []string, deviceIDs []string, debug bool) error {
```
phlogistonjohn: I'm not in love with this function signature. I won't object to it in the short term but I would have gone for an approach where the all flag populates a simple struct with the ID list and the id level (eg. clear all bricks matching id X, vs. clear all bricks under cluster Y, and made them mutually exclusive. That said, this seems OK for now.

apps/glusterfs/dbcommon_test.go
```diff
@@ -0,0 +1,216 @@
+//
```
phlogistonjohn: Why dbcommon_test.go ? Why not something like delete_bricks_empty_path_test.go ?

apps/glusterfs/dbcommon_test.go
```diff
@@ -0,0 +1,216 @@
...
+			if len(d.Bricks) > 0 {
+				return nil
+			}
+		}
+		t.Fatalf("should have at least one device with bricks")
+		return nil
+	})
+
+	// At this point, we have 15 legal bricks on each device and we have
+	// made note of a device(node and cluster in corollary) where to will
```
phlogistonjohn: Typo: "where to will"

apps/glusterfs/dbcommon_test.go
```diff
@@ -0,0 +1,216 @@
...
+
+	err = app.db.View(func(tx *bolt.Tx) error {
+		d, err = NewDeviceEntryFromId(tx, d.Info.Id)
+		return err
+	})
+	tests.Assert(t, err == nil, "expected err == nil, got:", err)
+	tests.Assert(t, len(d.Bricks) == 15,
+		"expected len(d.Bricks) == 15, got:", len(d.Bricks))
+
+	// CASE2: delete empty path bricks from this device
```
phlogistonjohn: This feels like it should be a 2nd test case function. You could move the setup stuff into a common function like "setupClusterForEmptyPathBricks" or something.

apps/glusterfs/dbcommon_test.go
```diff
@@ -0,0 +1,216 @@
...
+
+	err = app.db.View(func(tx *bolt.Tx) error {
+		d, err = NewDeviceEntryFromId(tx, d.Info.Id)
+		return err
+	})
+	tests.Assert(t, err == nil, "expected err == nil, got:", err)
+	tests.Assert(t, len(d.Bricks) == 15,
+		"expected len(d.Bricks) == 15, got:", len(d.Bricks))
+
+	// CASE3: delete empty path bricks from a node
```
phlogistonjohn: And a 3rd test case function. (I'll stop repeating myself now)

apps/glusterfs/dbcommon.go
raghavendra-talur: Agreed. Needs to be refactored.

apps/glusterfs/dbcommon_test.go
raghavendra-talur: it is recommended that the test file has same name as the corresponding file containing function. I will move functions from app_db.go and dbcommon.go which belong to db tooling to another file in next PR.

PR #1037, opened 2018-02-19
Untangle Pt. II - Device remove operation

phlogistonjohn:
This series wraps the action of removing a device (aka marking it failed) in an Operation. There are a couple of points worth mentioning for when this is being reviewed:

- The Operation abstraction did not fit great around the brick replace part of device remove. Thus, for now, the operation actually needs to use the allocator in more than just the Build function and takes it in the New... function.
- If the device has no bricks and can be updated immediately, it will set everything in the Build transaction, skip creating the pending operation and Exec and Finalize act as no-ops.
- Working on nodes was deferred for now. As each device w/in a node is removed an operation will be created but there is no overall operation for nodes. Similarly, the operation is left as an implementation detail of the Remove function rather than used directly in the rest endpoints.

### DIFF
--- a/apps/glusterfs/operations.go
+++ b/apps/glusterfs/operations.go
@@ -671,6 +671,137 @@ func (vdel *BlockVolumeDeleteOperation) Finalize() error {
 	})
 }
 
+// DeviceRemoveOperation is a phony-ish operation that exists
+// primarily to a) know that set state was being performed
+// and b) to serve as a starting point for a more proper
+// operation in the future.
+type DeviceRemoveOperation struct {
+	OperationManager
+	DeviceId  string
+	allocator Allocator
+}
+
+// Note: passing this allocator here a big hack, but its a temporary
+// workaround for how brick relocation currently works
+func NewDeviceRemoveOperation(
+	deviceId string, allocator Allocator, db wdb.DB) *DeviceRemoveOperation {
+
+	return &DeviceRemoveOperation{
+		OperationManager: OperationManager{
+			db: db,
+			op: NewPendingOperationEntry(NEW_ID),
+		},
+		DeviceId:  deviceId,
+		allocator: allocator,
+	}
+}
+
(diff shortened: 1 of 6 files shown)


apps/glusterfs/operations.go
```diff
@@ -671,6 +671,137 @@ func (vdel *BlockVolumeDeleteOperation) Finalize() error {
...
+	id, err := dro.deviceId()
+	if err != nil {
+		return err
+	}
+	if id == "" {
+		return nil
+	}
+
+	var d *DeviceEntry
+	if e := dro.db.Update(func(tx *bolt.Tx) error {
```
obnoxxx: Why `Update()` instead of `View()` here?

apps/glusterfs/operations.go
```diff
@@ -671,6 +671,137 @@ func (vdel *BlockVolumeDeleteOperation) Finalize() error {
...
+// and b) to serve as a starting point for a more proper
+// operation in the future.
+type DeviceRemoveOperation struct {
+	OperationManager
+	DeviceId  string
+	allocator Allocator
+}
+
+// Note: passing this allocator here a big hack, but its a temporary
+// workaround for how brick relocation currently works
```
obnoxxx: To be even more clear, if I got it right, the hack is not only passing the allocator in to the constructor, but storing it in the operation and using it later in the exec part instead of the build part:

Usually, the allocator is passed into the `Build` method, and the `Exec` part has nothing to do with the allocator. In the case of the `DeviceRemove`, the actual code is not (yet?) untangled enough to allow for this separation. Hence the woraround.

apps/glusterfs/operations.go
phlogistonjohn: Good catch, There's no reason this should be an Update.

apps/glusterfs/operations.go
phlogistonjohn: Correct.

PR #1039, opened 2018-02-19
Untangle Pt. II - Final cleanups

phlogistonjohn:
Depends on pr #1038 

This PR adds a few final test cases and makes delete volume and delete block volume operations clean up the pending operation metadata on failure. The overall operation still fails but no lingering pending operation will be left in the DB.


raghavendra-talur:
retest this please

PR #1043, opened 2018-02-20
Allow delete of device only when it is in failed state

raghavendra-talur:


apps/glusterfs/app_device_test.go
```diff
@@ -262,6 +262,73 @@ func TestDeviceAddDelete(t *testing.T) {
 	})
 	tests.Assert(t, err == nil)
 
+	// Set offline
+	request = []byte(`{
+				"state" : "offline"
+				}`)
```
phlogistonjohn: Not sure what's up with all the whitespace here.

obnoxxx:
i think python tests need to be adapted?

PR #1045, opened 2018-02-20
allow moving device from failed to offline state

raghavendra-talur:


raghavendra-talur:
>I assume this change is to allow the system to go online -> offline -> failed -> offline -> online?
>What did we have to do before, delete the device and the re-add it?

Yes, delete and re-add was the only way.

PR #1048, opened 2018-02-20
Refactor allocate bricks

obnoxxx:
This is a prep work for upcoming change to the allocator.
It makes allocateBricks() a little easier to digest.
And mainly it changes the code to use a single transaction around all loops.
This will later allow us to change GetNodes to dynamically load the ring from the db via the transaction.


obnoxxx:
@raghavendra-talur @phlogistonjohn 

FYI: the last patch "SQ(?): apps:glusterfs ONE transaction" is not strictly needed to prep for the allocator work, but I think it's consistent. If you agree, I will squash it with the previous commit, only kept it for you to review and decide/advise.

PR #1052, opened 2018-02-20
Rework allocator initialization

obnoxxx:
Let the allocator only load the ring from the DB in GetNodes().
This makes sure the allocator is always up to date when using it.


obnoxxx:
Currently, this includes the patches of PR #1048 (which is required) and of PR #1051 (which would be optional).

obnoxxx:
Do not merge for now (there's an issue I'm debugging)

PR #1056, opened 2018-02-21
apps: add a general operation retry mechanism

phlogistonjohn:
Build a retry mechanism into RunOperation and AsyncHttpOperation,
such that if the operation fails with an error of type
OperationRetryError and the Operation provides a greater-than-zero
MaxRetries value the operation will be re-run until the
max retries are exceeded or all operation phases succeed.

The operation Rollback method _must_ set all inputs to the operation
back to a state where Build can be safely executed again.

### DIFF
--- a/apps/glusterfs/operations.go
+++ b/apps/glusterfs/operations.go
@@ -19,6 +19,15 @@ import (
 	"github.com/boltdb/bolt"
 )
 
+type OperationRetryError struct {
+	OriginalError error
+}
+
+func (ore OperationRetryError) Error() string {
+	return fmt.Sprintf("Operation Should Be Retried; Error: %v",
+		ore.OriginalError.Error())
+}
+
 // The operations.go file is meant to provide a common approach to planning,
 // executing, and completing changes to the storage clusters under heketi
 // management as well as accurately reflecting these changes in the heketi db.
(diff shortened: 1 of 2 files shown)


apps/glusterfs/operations.go
```diff
@@ -980,6 +1010,10 @@ func RunOperation(o Operation,
 		return err
 	}
 	if err := o.Exec(executor); err != nil {
+		if _, ok := err.(OperationRetryError); ok && o.MaxRetries() > 0 {
```
obnoxxx: Somehow, the way it's done feels a little bit backward and redundant. Can't we make it more natural by wrapping it the other way to remove code duplication, like so (simplified):

```
func RunOperation(o, e) (err) {
    max := o.MaxRetries()
    for i := 0; i < max + 1; i++ {
        if e := o.Build(); e != nil {
            return e
        }

        err = o.Exec()
        if err == nil {
            return o.Finalize()
        }

        if re := o.Rollback(); re != nil {
            return re
        }

        if _, ok := err.(OperationRetryError); !ok {
            break
        }
    }

    // too many retries or not OperationRetryError
    return err
}
```

Of course for the async form this would need to be adapted...

phlogistonjohn: This is a very good point and I agree that the current loop is a bit funny. However, it might be better to take this PR and clean up these loops in a followup. That way we get some testing around the overall feature for volume create retries and we validate the overall concept before cleaning up this part.

obnoxxx: Agreed!

Issue #1035, opened 2018-02-19
Provide metrics for consumption by Prometheus


allen13:
I am currently using this project in my Openshift cluster to monitor CNS/Heketi: https://github.com/CSCfi/heketi-metrics-exporter

While it works great, it would be best to directly integrate it into the heketi server. Since Openshift is moving from hawkular to prometheus soon I assume it will eventually happen.

Is this something on the roadmap? Would a pull request be accepted?

Thanks!

phlogistonjohn:
Hi, that looks pretty cool. I personally don't have experience with Prometheus other than hearing it's name bandied about but I certainly think its worth consideration.

I gave the repo you linked to a look over and it looks interesting. I'm sure a PR would be seriously considered and discussed if one were filed.

phlogistonjohn:
Feature was merged to heketi. New metrics or other changes to the metrics endpoint should be done as individual new issues/prs.

PR #1057, opened 2018-02-21
Add prometheus metrics collection endpoint

allen13:
First attempt at addressing this issue: https://github.com/heketi/heketi/issues/1035


phlogistonjohn:
Hi! This is really cool!

I've given this a quick look over and without going too deeply into code line by line here are a few high-level thoughts I have about the change.

- I think the URL should should refer to Prometheus (or if there's a common name for the data exchange protocol, that name). In other words /metrics/prometheus. This way we have the ability to additional metrics types in the future if we need to without conflicting over the name metrics.
- If we're going to be part of the App in glusterfs it should set the URLs up like the other rest endpoints do (see SetRoutes in app.go).
- I'm not sold on adding all these new functions (that implement the prometheus collector interface) directly to glusterfs.App. I already feel that app is too big. I would consider writing a new struct that takes a db argument and implements the the collector interface. In addition I'd like to see it walk the db entries in a single transaction.
- Because this is a new feature we're going to need a few (basic) tests to start with. I don't expect anything fancy, we don't need to verify the exact behavior of Prometheus, just that we can hit this code and get some data out of it. This could be done in a 2nd patch.

I hope this feedback does not seem intimidating. I'm really enthusiastic about your contribution! Feel free to discuss any and all of my comments above. If you do agree to make changes feel free to make them incrementally as well.

allen13:
Thanks for the great feedback! I’m definitely looking forward addressing all of the issues you pointed out. I plan work on this incrementally over the next week.

allen13:
Issue: Extracting metrics collection from the app object

After refactoring I definitely liked moving the metrics code outside of the app, especially since I could use mocks for the app interface for testing. I wasn't sure about passing in the DB directly since the app basically owned it. By creating the app_topology.go file I felt it was a comfortable middle ground. Adding TopologyInfo to the interface seemed like the best route to go. 

I'm definitely open to more guidance on this latest commit: https://github.com/heketi/heketi/pull/1057/commits/48f3c189dee6a22ac5ba0eb9bf5d5c1b19de17f6

phlogistonjohn:
I'll try to dig into the actual changes soon, hopefully later today or tomorrow morning.

> By default prometheus expects the /metrics url which can be auto discovered in a 
> Kubernetes/Openshift environment. Additionally, the prometheus metrics format has become a sort 
> of default for pull metrics. Prometheus's direct competitor InfluxDB even uses it.

That's interesting. Seems a bit odd that the client wants to "own" that part of the server. Do you have any links handy where I could read up more on this behavior? Regardless, this is a good enough reason to leave it as metrics for now.

> I actually started off using SetRoutes but then I realized that /metrics endpoint would be put behind authentication, which would break metrics collection when auth was turned on.

Understood. This does remind me that heketi's auth support is fairly sub-par. I would not be shocked to find out that prometheus doesn't support JWT as a client. Does it support any auth types?

> After refactoring I definitely liked moving the metrics code outside of the app, especially since I could use mocks for the app interface for testing. 

:-)

> I wasn't sure about passing in the DB directly since the app basically owned it. 

Personally, I'd be OK with that. It's not perfect. But because the app/glusterfs package is kind of a bunch of related things mushed together its not easy just to use the piece you want at a clean boundary.

> By creating the app_topology.go file I felt it was a comfortable middle ground. Adding TopologyInfo to the interface seemed like the best route to go.

OK. I'll try to spend some time with this soon, and provide more feedback. Thanks!

allen13:
Here is the documentation for Prometheus scrape configs. https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config

Here is the kubernetes-pod scrape config: https://github.com/prometheus/prometheus/blob/master/documentation/examples/prometheus-kubernetes.yml#L239

The /metrics can be overriden by using this annotation: prometheus.io/path 
It would still be nice to keep the default but it can be worked around.

Additionally, if you notice the scrap_config docs there is support for token authentication. However, this would require the kubernetes scrape configs to be modified entirely as there is no annotation that controls that setting currently.

allen13:
Created a new pull request with a single cleaned up commit complete with all tests: https://github.com/heketi/heketi/pull/1068

PR #1060, opened 2018-02-22
tests: add a check that functional test go code compiles

phlogistonjohn:
Add a test to verify that the Go code used for the functional
tests can be compiled. This should help find some issues with the
functional test code before the functional tests are run.


phlogistonjohn:
Weird, I could have sworn I ran the test suite before pushing. Oh, well. Fixed.

It probably could but I felt this was a simpler (one liner) approach that I knew would find any tests "buried" in subdirs in the functional tests.

PR #1061, opened 2018-02-22
Establish an API for getting and setting the server log level

phlogistonjohn:
These patches have been dredged out of my virtual patch basement and cleaned up (and posted). I started this a while back, got distracted by other tasks, but I didn't want this to linger forever.

This will allow those with access to Heketi to change the log level on the fly. Examples:

```
$ ./client/cli/go/heketi-cli loglevel get
info
$ ./client/cli/go/heketi-cli loglevel set debug
Server log level updated
$ ./client/cli/go/heketi-cli loglevel get
debug
```

Since this is a rather administrative/debugging task I opted to make the PR w/o first adding to the python client for it. Current TODOs include documenting the API and CLI command changes.


phlogistonjohn:
I wouldn't want to cram too much "unrelated" stuff into the request and response, but it might make sense to namespace this and make the url something like /server/logging rather than /logging. Then we can say any new server runtime things should go under /server ?

obnoxxx:
yeah, I mainly wanted to prevent clobbering the main API level with random runtime-config options in the future. Namespace is good! `/server` seems somewhat redundant. what about `/setup` ? just thinking aloud....

phlogistonjohn:
Bikeshedding namespace name: /heketi/  /internal/

To me "setup" often sounds too verb-y. I also am hoping to avoid "settings" or "configuration" so as to avoid indicating that anything related to generic settings (for gluster, or kube, or whatever) are implied. Ideally the name would reflect that the namespace is going to impact the behavior of the heketi daemon and that's all (ideally - I hope).

PR #1062, opened 2018-02-23
control behavior of refusing to start with stale pending operation db entries

phlogistonjohn:
This series aims to improve the behavior of the server w.r.t. starting up and having existing pending operations entries in the db from a previous instance of the server. This means the operation was previously in progress and the db may not be consistent with what's on gluster.

First change makes it so there's no panic externally visible. The server just exits non-zero.

Then we add a way for users to force Heketi to start even when these items are in the db. This isn't clean but exists for the "I don't care, just start it, I'll deal with the problems later". Hopefully, they will try to deal with the problems later.

### DIFF
--- a/apps/glusterfs/app_test.go
+++ b/apps/glusterfs/app_test.go
@@ -284,3 +284,29 @@ func TestCannotStartWhenPendingOperations(t *testing.T) {
 
 	t.Fatalf("Test should not reach this line")
 }
+
+func TestCanStartWhenPendingOperationsIgnored(t *testing.T) {
+	dbfile := tests.Tempfile()
+	defer os.Remove(dbfile)
+
+	// create a app that will only be used to set up the test
+	app := NewTestApp(dbfile)
+	tests.Assert(t, app != nil)
+
+	// populate the db with a "dummy" pending op entry.
+	// without the environment var we're setting later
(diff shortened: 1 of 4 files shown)


apps/glusterfs/app_test.go
```diff
@@ -284,3 +284,28 @@ func TestCannotStartWhenPendingOperations(t *testing.T) {
...
+	// trigger a panic the next time an app is instantiated
+	err := app.db.Update(func(tx *bolt.Tx) error {
+		op := NewPendingOperationEntry(NEW_ID)
+		op.Save(tx)
+		return nil
+	})
+	tests.Assert(t, err == nil, "expected err == nil, got:", err)
+	app.Close()
+
+	// now creating a new app should panic
```
raghavendra-talur: the comment here is wrong. I guess even comment on line 296 should be reworded.

main.go
```diff
@@ -226,6 +225,21 @@ func setWithEnvVariables(options *Config) {
 	}
 }
 
+func setupApp(fp *os.File) (a *glusterfs.App) {
+	defer func() {
+		err := recover()
```
raghavendra-talur: Nice !

phlogistonjohn:
Comments updated.

PR #1064, opened 2018-02-23
apps:glusterfs: store the allocator in the app again.

obnoxxx:
Previously, the allocator was not stored in the app
to prevent stale data as loaded from the db. Now that
the data is always loaded freshly in GetNodes(),
the stored allocator is not really a cache any more
but basically reflects the chosen allocator methods.
So at app start, we instantiate and store the configured
allocator again.


phlogistonjohn:
Test failures look to be due to gopkg.in hiccup.

phlogistonjohn:
The more I think about it, I'd rather see the internals of the allocator (re)moved so that they exist only within GetNodes. Then we can remove the mutex and know the rings exist only within the single transaction. This might mean creating an intermediate type to hold the ring first though.

phlogistonjohn:
@obnoxxx see pr #1078 for a mostly-fleshed-out version of what I was talking about in my previous comment.

obnoxxx:
I continues this work in PR #1082 .

obnoxxx:
superseded and obsoleted by #1086 .

PR #1068, opened 2018-02-27
Add prometheus metrics collection endpoint

allen13:


allen13:
See comments from previous pull request for context: https://github.com/heketi/heketi/pull/1057

apps/app.go
```diff
@@ -11,11 +11,13 @@ package apps
 
 import (
 	"github.com/gorilla/mux"
+	"github.com/heketi/heketi/pkg/glusterfs/api"
 	"net/http"
 )
 
 type Application interface {
 	SetRoutes(router *mux.Router) error
+	TopologyInfo() (*api.TopologyInfoResponse, error)
```
phlogistonjohn: I understand what you are trying to do here, but I'm not in love with altering the Application interface. I'm going to think on it a little more and perhaps come up with a more concrete suggestion.

apps/glusterfs/app_topology.go
```diff
@@ -0,0 +1,110 @@
+//
+// Copyright (c) 2015 The heketi Authors
```
phlogistonjohn: This should be 2018.

Our rule (at least as far as I've been involved in the project) has been that new files should have the current year in the copyright line, but old files do not need to be updated when you touch them.

apps/glusterfs/app_topology.go
```diff
@@ -0,0 +1,110 @@
...
+				},
+			}
+			cluster.Id = clusterInfo.Id
+
+			for _, volume := range clusterInfo.Volumes {
+				volumeInfo, err := volumeInfo(tx, volume)
+				if err != nil {
+					return err
+				}
+				if volumeInfo.Cluster == cluster.Id {
```
phlogistonjohn: I don't understand this check. Since you are iterating over volume ids on the cluster object how would you have volumes belonging to another cluster here?

apps/glusterfs/app_topology_test.go
```diff
@@ -0,0 +1,58 @@
+//
+// Copyright (c) 2015 The heketi Authors
```
phlogistonjohn: 2018 again

apps/glusterfs/app_topology_test.go
```diff
@@ -0,0 +1,58 @@
...
+	topologyInfo, err := app.TopologyInfo()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	testCluster := topologyInfo.ClusterList[0]
+
+	tests.Assert(t, testCluster.Id != "")
+	tests.Assert(t, len(testCluster.Volumes) > 0)
+	tests.Assert(t, len(testCluster.Nodes) > 0)
```
phlogistonjohn: Somewhat minor nit: It's much nicer if we also provide an "expected string". Something like:
```golang
tests.Assert(t, testCluster.Id != "", `expected testCluster.Id != "", got`, testCluster.Id)
```
It is annoying to type but much better if you come back 6 weeks later or have to fix a failing test somone else wrote.

apps/glusterfs/app_topology_test.go
```diff
@@ -0,0 +1,58 @@
...
+	app := NewTestApp(tmpfile)
+	defer app.Close()
+
+	err := setupSampleDbWithTopology(app,
+		1,      // clusters
+		2,      // nodes_per_cluster
+		1,      // devices_per_node,
+		500*GB, // disksize)
+	)
+	if err != nil {
```
phlogistonjohn: We prefer to use the tests.Assert to check for these sort of things. It's not the most typical golang but its succinct and we generally like it. Also see my note below about providing expected string.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,131 @@
+package metrics
+
+import (
+	"github.com/heketi/heketi/apps"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"net/http"
+)
```
phlogistonjohn: Can you please separate the standard library packages from the external packages with an empty newline? You did it that way in app_topology_test.go

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,131 @@
...
+const (
+	namespace = "gluster"
+)
+
+var (
+	up = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, "", "up"),
+		"Was the last query of heketi successful.",
+		nil, nil,
+	)
```
phlogistonjohn: These calls to NewDesc and its arguments seems a bit repetitive. I wonder it it wouldn't be nicer to have something like:
```golang
func promDesc(name, help string,  variableLabels []string) *promethus.Desc {
    return prometheus.NewDesc(....)
}
```

In order to cut down on the boilerplate. As a 2nd consideration, do you think there's any potential of name conflicats if our namespace="gluster" rather than "heketi"? IOW, if someone was also getting metrics from actual gluster nodes as well as Heketi?

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,131 @@
...
+// Describe all the metrics exported by Heketi exporter. It implements prometheus.Collector.
+func (m *Metrics) Describe(ch chan<- *prometheus.Desc) {
+	ch <- clusterCount // done
+	ch <- volumesCount // done
+	ch <- nodesCount   // done
+	ch <- deviceCount  // done
+	ch <- deviceSize   // done
+	ch <- deviceFree   // done
+	ch <- deviceUsed   // done
+	ch <- brickCount   // done
```
phlogistonjohn: These done comments don't really add any value to understanding the code IMO.

pkg/metrics/metrics_test.go
```diff
@@ -0,0 +1,58 @@
...
+
+import (
+	"github.com/gorilla/mux"
+	"github.com/heketi/heketi/pkg/glusterfs/api"
+	"io/ioutil"
+	"net/http"
+	"net/http/httptest"
+	"regexp"
+	"testing"
+)
```
phlogistonjohn: Again, please separate the import of std library packages from externals.

pkg/metrics/metrics_test.go
```diff
@@ -0,0 +1,58 @@
...
+	res.Body.Close()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	match, err := regexp.Match("gluster_nodes_count{cluster=\"t1\"} 0", body)
+	if !match || err != nil {
+		t.Fatal("gluster_nodes_count{cluster=\"t1\"} 0 should be present in the metrics output")
+	}
+}
```
phlogistonjohn: This approach to mocking out the app is quite nice!
I'd love to see at least one more test function that has filled in more stuff (volumes, etc) to exercise metrics collection a little more.

phlogistonjohn:
Oh, and as a side note feel free to rebase/squash/whatever your working branch for this PR. You don't need to make a new PR even if you alter the current patch. The maintainers here do it frequently so we're all used to it.

allen13:
Thanks for the feedback! Already working on it.

pkg/metrics/metrics.go
allen13: I'm on board with both.

allen13:
Latest commit should address most of the issues.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,158 @@
...
+
+package metrics
+
+import (
+	"github.com/heketi/heketi/apps"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+
+	"net/http"
+)
```
phlogistonjohn: The standard in the code is std library modules up top. Sorry if I wasn't clear about that before.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,158 @@
...
+	topinfo, err := m.app.TopologyInfo()
+	if err != nil {
+		ch <- prometheus.MustNewConstMetric(
+			up, prometheus.GaugeValue, 0.0,
+		)
+	} else {
+		ch <- prometheus.MustNewConstMetric(
+			up, prometheus.GaugeValue, 1.0,
+		)
+	}
```
phlogistonjohn: Since most of this block is shared this could read better as:
```golang
upVal := 0.0
if err == nil {
    upVal := 1.0
}
ch <- prometheus.MustNewConstMetric(up, prometheus.GaugeValue, upVal)

```

allen13:
Made requested changes and cleaned things up a bit more.

john-bakker:
Any expectation when this will be merged ? Would be really nice to have prometheus endpoint in the default heketi service.

humblec:
@allen13 this is a good start and thanks for the PR ! Can you also update the wiki or readme with these details ?

allen13:
Added /metrics endpoint to the api.md docs in latest revision.

phlogistonjohn:
Thanks @humblec for looking at the PR!

@allen13 the doc change seems pretty good. I'm not in love with the length of the example output but if that's what Prometheus expects we'll just live with it for now.

In the meantime, it appears that the patch series needs rebasing against master. Some changes to the internal calls appear to be causing the travis test runs to fail. Thank you!

humblec:
@allen13 may be we can only keep "heketi" specific buckets in the wiki or readme output, if rest are default ones and not introduced by this PR. Secondly, the descriptions for below looks vague or not correct.
~~~
# TYPE heketi_device_free gauge
heketi_device_free{cluster="c1",device="d1",hostname="n1"} 1
# HELP heketi_device_size How many Devices were up at the last query.
# TYPE heketi_device_size gauge
heketi_device_size{cluster="c1",device="d1",hostname="n1"} 2
# HELP heketi_device_used How many Devices were up at the last query.
# TYPE heketi_device_used gauge
heketi_device_used{cluster="c1",device="d1",hostname="n1"} 1
# HELP heketi_nodes_count How many Nodes were up at the last query.
# TYPE heketi_nodes_count gauge
~~~

allen13:
Thanks @humblec for noticing the poor metric descriptions! Just cleaned them up to be correct and more to the point.

allen13:
Also added logging to the metrics collector if it can't get topology info. Wasn't sure how to do logging since it was outside of the app so I just used the standard log module.

phlogistonjohn:
test this please

phlogistonjohn:
ok to test

phlogistonjohn:
@centos-ci ok to test

phlogistonjohn:
The centos-ci run failed due to a missing dependency. I've created a new PR on the centos-ci repo to add it. Hopefully, we can merge that pr in soon and we can then kick off another ci run.

phlogistonjohn:
@centos-ci retest this please

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,195 @@
...
+			}
+		}
+	}
+}
+
+func NewMetricsHandler(app apps.Application) http.HandlerFunc {
+	m := &Metrics{
+		app: app,
+	}
+	prometheus.MustRegister(m)
```
obnoxxx: What does this do? Are we always requiring some prometheus service with this change?

pkg/metrics/metrics_test.go
```diff
@@ -0,0 +1,96 @@
...
+				},
+			},
+		},
+	}
+
+	ts := httptest.NewServer(NewMetricsHandler(ta))
+	defer ts.Close()
+
+	res, err := http.Get(ts.URL)
+	if err != nil {
```
obnoxxx: In heketi, we use s/th like `tests.Assert(t, err == nil)` ...

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,195 @@
...
+	upVal := 0.0
+	if err == nil {
+		upVal = 1.0
+	}
+
+	ch <- prometheus.MustNewConstMetric(up, prometheus.GaugeValue, upVal)
+
+	//Do not collect further metrics if heketi is down
+	if err != nil {
+		log.Println("Can't collect toplogy info for metrics: " + err.Error())
```
obnoxxx: In Heketi, the pattern would be s/th like `logger.LogError(..., err)`

phlogistonjohn: I figured we'd just take this bit as-is and fix it up later.

apps/app.go
phlogistonjohn: I felt it was not worth blocking this change for. Since I couldn't come up with an better approach I could articulate I figured it was something we could improve later. Probably with changes to how our url endpoints are registered.

glide.lock
```diff
@@ -1,22 +1,28 @@
-hash: 5fc263bfacc703e0a599166ccae968891dd53d7b4e42a3d8356b3194e45833bb
-updated: 2017-12-18T16:46:06.7851579+05:30
+hash: c48583fd7daa415c986e22638c3036b02f22698a0f5d5f583ce1b6c06833a3b3
+updated: 2018-02-28T14:36:47.873451024Z
```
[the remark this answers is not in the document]
phlogistonjohn: This is a good point and something I sort of overlooked. We should only see changes to packages that are brought in by the new dependency(s). Node that glide itself probably made these changes to glide.lock when glide.yaml was updated. But still it would be good to minimize any changes to the dependencies.

(I'm not going to bring up the fact that glide is no longer the favored approach to dependencies in golang here though. :-))

pkg/metrics/metrics.go
allen13: Must Register starts the gather loop for metrics. It's not really a service is the traditional sense. Here are the docs for it: https://github.com/prometheus/client_golang/blob/master/prometheus/registry.go#L89

glide.lock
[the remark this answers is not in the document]
allen13: Not sure how to address this one. Will accept any guidance here.

allen13:
To address all of these questions @obnoxxx :

**Could you explain a little bit the splitting of a pkg/metrics and an app_topology? (Maybe also an opening comment block in the go files to briefly explain the scope of the module).**

@phlogistonjohn had me split these items up in a previous pull request to lighten the bulk of the the app namespace. Additionally, it provided a great opportunity to mock out the app for testing

**Why not a new heketi API but a new separate router?**

Not sure what a router is but I feel like topology should already be an api request since the client has a function that requests topology via multiple calls. Adding a topology endpoint could reduce this to a single call.

**Do we always require prometheus now at runtime?**

It's not actually requiring prometheus but rather the prometheus metrics exposition format via the client. It's just an endpoint that returns text, no configuration required.  

**Can we make it configurable?**

Since it has such low overhead and can be useful even when there is no prometheus to collect it isn't necessary. However, throwing an if block in wouldn't be too hard.

nixpanic:
I'm interested in this as well, and would like to add other metrics once this change is merged. The main things I am thinking of, are
- count & result of volume creation requests (to get success and failure rates)
- count & result of volume delete requests (to get success and failure rates)
- details from PR #1125 

@allen13, do you need any assistance with addressing outstanding review comments on this PR?

obnoxxx:
> To address all of these questions @obnoxxx :

Thanks for the replies, @allen13 
 
> **Could you explain a little bit the splitting of a pkg/metrics and an app_topology? (Maybe also an opening comment block in the go files to briefly explain the scope of the module).**
> 
> @phlogistonjohn had me split these items up in a previous pull request to lighten the bulk of the the app namespace. Additionally, it provided a great opportunity to mock out the app for testing
> 
> **Why not a new heketi API but a new separate router?**
> 
> Not sure what a router is

Well, I guess I meant "route": In `./main.go`, we already have a line adding a `/hello` route:
```
router.Methods("GET").Path("/hello").Name("Hello").HandlerFunc(...)
```

and you add one similar for `/metrics`:

```
router.Methods("GET").Path("/metrics").Name("Metrics").HandlerFunc(metrics.NewMetricsHandler(app))
```
All other heketi (api) commands are added later with:
```
n := negroni.New(negroni.NewRecovery(), negroni.NewLogger())
...
heketiRouter := mux.NewRouter().StrictSlash(true)
...
n.UseHandler(heketiRouter)
...
router.NewRoute().Handler(n)
```

So I was just wondering why not put the metrics to the heketi piece. Maybe it just does not need all the extra setup? And I'm really asking here, because I'm not that familiar with this part of the code, since I've been mostly working on the internals. Maybe @raghavendra-talur can give some comments, too?

> but I feel like topology should already be an api request since the client has a function that requests topology via multiple calls. Adding a topology endpoint could reduce this to a single call.

You mean the cli function (`heketi-cli topology info`)?

To my simple mind at least it is very confusing to have this change mixed with the actual metrics patch. It is also only used in the metrics test code in your PR.

If you want to introduce a topology app, that might be a good idea, but I think this should be done in a different patch, maybe even in a different PR, and then the client should also use this new functionality instead of requesting the various API calls to construct the topology... See `client/cli/go/cmds/topology.go`, and also see how  `client/api/go-client/topology.go:TopologyInfo()` is almost identical to your `ToplogyInfo()` function, but does the operations at a different level. So this should probably be reconciled. 

Note that I don't want to be a stickler here, but to me it is important to have good, self-contained patches that do one thing. I get confused when too many things are mixed.

> **Do we always require prometheus now at runtime?**
> 
> It's not actually requiring prometheus but rather the prometheus metrics exposition format via the client. It's just an endpoint that returns text, no configuration required.

That's good. 

> **Can we make it configurable?**
> 
> Since it has such low overhead and can be useful even when there is no prometheus to collect it isn't necessary. However, throwing an if block in wouldn't be too hard.

That was kind of depending on the assumption that we might need to have prometheus running all the time now with this change. Since that does not seem to be the case, the need for such an option is way less now.

pkg/metrics/metrics.go
obnoxxx: ok

pkg/metrics/metrics.go
obnoxxx: ok

glide.lock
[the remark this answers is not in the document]
obnoxxx: Will try to explain in more detailed comments.

glide.lock
```diff
@@ -1,22 +1,28 @@
...
 imports:
+- name: bitbucket.org/ww/goautoneg
+  version: 75cd24fc2f2c2a2088577d12123ddee5f54e0675
```
obnoxxx: Not sure what added this, was it pulled by some of the prometheus stuff?

glide.lock
```diff
@@ -1,22 +1,28 @@
...
 - name: github.com/asaskevich/govalidator
-  version: 852d82c746b23d9b357b210ea470d99f4e023b72
+  version: 593d64559f7600f29581a3ee42177f5dbded27a9
```
obnoxxx: This hunk is probably an automatic update by glide, unrelated to the prometheus patch and should be removed from this patch.

glide.lock
```diff
@@ -1,22 +1,28 @@
...
 - name: github.com/auth0/go-jwt-middleware
   version: f3f7de3b9e394e3af3b88e1b9457f6f71d1ae0ac
 - name: github.com/Azure/go-ansiterm
   version: 70b2c90b260171e829f1ebd7c17f600c11858dbe
   subpackages:
   - winterm
+- name: github.com/beorn7/perks
+  version: 3ac7bf7a47d159a033b107610db8a1b6575507a4
+  subpackages:
+  - quantile
```
obnoxxx: Is this needed by prometheus? Otherwise remove it...

glide.lock
```diff
@@ -1,22 +1,28 @@
...
 - name: github.com/boltdb/bolt
-  version: 583e8937c61f1af6513608ccc75c97b6abdf4ff9
+  version: 2f1ce7a837dcb8da3ec595b1dac9d0632f0f99e8
```
obnoxxx: remove

glide.lock
```diff
@@ -1,22 +1,28 @@
...
 - name: github.com/davecgh/go-spew
   version: 5215b55f46b2b919f50a1df0eaa5886afe4e3b3d
   subpackages:
   - spew
 - name: github.com/dgrijalva/jwt-go
-  version: d2709f9f1f31ebcda9651b03077758c1f3a0018c
+  version: dbeaa9332f19a944acb5736b4456cfcc02140e29
```
obnoxxx: remove

glide.lock
```diff
@@ -66,6 +72,10 @@ imports:
   - sortkeys
 - name: github.com/golang/glog
   version: 44145f04b68cf362d9c4df2182967c2275eaefed
+- name: github.com/golang/protobuf
+  version: 8616e8ee5e20a1704615e6c8d7afcdac06087a67
+  subpackages:
+  - proto
```
obnoxxx: Is this needed by prometheus? Otherwise remove...

glide.lock
```diff
@@ -90,8 +100,28 @@ imports:
...
+  version: fa8ad6fec33561be4280a8f0514318c79d7f6cb6
+  subpackages:
+  - go
+- name: github.com/prometheus/common
+  version: ffe929a3f4c4faeaa10f2b9535c2b1be3ad15650
+  subpackages:
+  - expfmt
+  - model
+- name: github.com/prometheus/procfs
+  version: 454a56f35412459b5e684fd5ec0f9211b94f002a
```
obnoxxx: This prometheus block is good and probably all we need (or almost all).

glide.lock
```diff
@@ -107,7 +137,7 @@ imports:
   subpackages:
   - codec
 - name: github.com/urfave/negroni
-  version: fde5e16d32adc7ad637e9cd9ad21d4ebc6192535
+  version: 5dbbc83f748fc3ad38585842b0aedab546d0ea1e
```
obnoxxx: remove

glide.lock
```diff
@@ -267,7 +297,6 @@ imports:
   - pkg/apis/storage/install
   - pkg/apis/storage/v1
   - pkg/apis/storage/v1beta1
-  - pkg/client/clientset_generated
```
obnoxxx: remove this hunk from the patch

glide.yaml
```diff
@@ -24,3 +24,5 @@ import:
   version: v3.0.0-beta.0
 - package: github.com/go-ozzo/ozzo-validation
   version: v3.3
+- package: github.com/prometheus/client_golang
+  version: v0.8.0
```
obnoxxx: Do we need to hard-code the version of the prometheus client? If no, we should probably not need to add these two lines....

glide.lock
```diff
@@ -90,8 +100,28 @@ imports:
   - buffer
   - jlexer
   - jwriter
+- name: github.com/matttproud/golang_protobuf_extensions
+  version: fc2b8d3a73c4867e51861bbdd5ae3c1f0869dd6a
+  subpackages:
+  - pbutil
```
obnoxxx: needed for prometheus? otherwise remove...

glide.lock
obnoxxx: If not needed by prometheus, remove it from the patch

glide.lock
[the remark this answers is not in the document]
obnoxxx: (and yeah, we should probably move to "dep" at some point...)

obnoxxx:
> More than willing to do whatever it takes to get these changes through. Was waiting on guidance from @obnoxx. If someone more familiar with the heketi codebase would like to take over I would be ok with that.

I think you have already provided a great piece of code!
I am not trying to be a pain in the neck but I am 
1. trying to understand and
2. trying to come up with even better patches that need fewer follow-up fixes

This is just the normal review process, but I agree it might be a little too much for a first-time contributor to the project. If you feel the scrutiny is too much, we'll happily help you work on this some more, or even take it over. But I think you are absolutely on top of it, if you have enough patience with me. :-)

allen13:
Thanks for the reply @obnoxxx ! Spent a few hours just working on the glide requests. I'm having a lot of trouble changing anything in glide. Every small change breaks the build. Just running make will trigger an update. Any recommendations for glide workflow would be helpful.

Running into this error:

```
go build -i -ldflags "-X main.HEKETI_VERSION=v6.0.0-130-gb6cfb71-metrics -extldflags '-z relro -z now'" -o heketi
# github.com/heketi/heketi/pkg/kubernetes
pkg/kubernetes/backupdb.go:30:32: cannot use c (type *"github.com/heketi/heketi/vendor/k8s.io/client-go/rest".Config) as type *"github.com/heketi/heketi/vendor/k8s.io/kubernetes/vendor/k8s.io/client-go/rest".Config in argument to clientset.NewForConfig
pkg/kubernetes/pv.go:35:44: cannot use "github.com/heketi/heketi/vendor/k8s.io/apimachinery/pkg/api/resource".MustParse(fmt.Sprintf("%vGi", volume.VolumeInfo.VolumeCreateRequest.Size)) (type "github.com/heketi/heketi/vendor/k8s.io/apimachinery/pkg/api/resource".Quantity) as type "github.com/heketi/heketi/vendor/k8s.io/kubernetes/vendor/k8s.io/apimachinery/pkg/api/resource".Quantity in assignment
make: *** [heketi] Error 2
```

allen13:
I’ll be short on time for the next few months and won’t be able to get this PR across line in the near term. Would be really happy if someone would take this over, possibly from scratch if needed. It should be noted that most of the code was borrowed directly from the heketi-exporter project.

obnoxxx:
@allen13 wrote:
> I’ll be short on time for the next few months and won’t be able to get this PR across line in the near term. Would be really happy if someone would take this over, possibly from scratch if needed. It should be noted that most of the code was borrowed directly from the heketi-exporter project.

Thanks so much for your work!

No need at all to redo from scratch at all!
We'll take it on and will try to get it merged soon, maybe even for heketi v7.
It's mostly reducing the glide change and splitting out the topology app part in a separate part.
So your code will mostly be kept. If you agree, I will start changing the patches a bit but keep
your authorship on them.

Thanks again for your contributions!

phlogistonjohn:
@obnoxxx 

Looks like there might still be dependency issues. From travis logs:

go build -i -ldflags "-X main.HEKETI_VERSION=v6.0.0-261-g6591b41-HEAD -extldflags '-z relro -z now'" -o heketi
vendor/github.com/prometheus/client_golang/prometheus/summary.go:23:2: cannot find package "github.com/beorn7/perks/quantile" in any of:
	/home/travis/gopath/src/github.com/heketi/heketi/vendor/github.com/beorn7/perks/quantile (vendor tree)
	/home/travis/.gimme/versions/go1.10.linux.amd64/src/github.com/beorn7/perks/quantile (from $GOROOT)
	/home/travis/gopath/src/github.com/beorn7/perks/quantile (from $GOPATH)
vendor/github.com/prometheus/client_golang/prometheus/desc.go:23:2: cannot find package "github.com/golang/protobuf/proto" in any of:
	/home/travis/gopath/src/github.com/heketi/heketi/vendor/github.com/golang/protobuf/proto (vendor tree)
	/home/travis/.gimme/versions/go1.10.linux.amd64/src/github.com/golang/protobuf/proto (from $GOROOT)
	/home/travis/gopath/src/github.com/golang/protobuf/proto (from $GOPATH)
vendor/github.com/prometheus/common/expfmt/decode.go:25:2: cannot find package "github.com/matttproud/golang_protobuf_extensions/pbutil" in any of:
	/home/travis/gopath/src/github.com/heketi/heketi/vendor/github.com/matttproud/golang_protobuf_extensions/pbutil (vendor tree)
	/home/travis/.gimme/versions/go1.10.linux.amd64/src/github.com/matttproud/golang_protobuf_extensions/pbutil (from $GOROOT)
	/home/travis/gopath/src/github.com/matttproud/golang_protobuf_extensions/pbutil (from $GOPATH)
make: *** [heketi] Error 1

phlogistonjohn:
Glide updates look OK to me.

PR #1069, opened 2018-02-27
apps: split volume_entry_allocate.go into two files

phlogistonjohn:
Now all the content that was formerly all in volume_entry_allocate.go is split across volume_entry_allocate.go and brick_allocate.go.
Where volume_entry_allocate.go only contains methods on the VolumeEntry struct and brick_allocate.go only contains the standalone utility methods.

The expectation is that more code will be made command and "utility-ified" in the near future.


phlogistonjohn:
Patches squashed.

obnoxxx:
LGTM, thanks!

PR #1070, opened 2018-02-28
apps: add brick and device sets to brick allocation functions

phlogistonjohn:
These changes add new brick and device types to brick_allocate.go, and tries to use them to better organize some of the code within. If it seems like a lot of change for little benefit, the expectation is this approach will become more fundamental and useful in future series.

### DIFF
--- a/apps/glusterfs/brick_allocate.go
+++ b/apps/glusterfs/brick_allocate.go
@@ -153,7 +153,7 @@ func allocateBricks(
 	allocator Allocator,
 	cluster string,
 	v *VolumeEntry,
-	bricksets int,
+	numBrickSets int,
 	brick_size uint64) (*BrickAllocation, error) {
 
 	r := &BrickAllocation{
@@ -167,7 +167,7 @@ func allocateBricks(
 		txdb := wdb.WrapTx(tx)
 
 		// Determine allocation for each brick required for this volume
-		for brick_num := 0; brick_num < bricksets; brick_num++ {
+		for brick_num := 0; brick_num < numBrickSets; brick_num++ {
 			logger.Info("brick_num: %v", brick_num)
 
 			// Generate an id for the brick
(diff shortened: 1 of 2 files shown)


phlogistonjohn:
I'm going to rebase this on top of the patches in pr #1072 because the two conflict with each other and the changes in the other pr are smaller.

obnoxxx:
@phlogistonjohn now that #1072  is merged, this can be rebased.

apps/glusterfs/brick_allocate.go
```diff
@@ -16,17 +16,51 @@ import (
...
+
+func NewBrickSet(s int) *BrickSet {
+	return &BrickSet{SetSize: s, Bricks: []*BrickEntry{}}
+}
+
+func (bs *BrickSet) Add(b *BrickEntry) {
+	bs.Bricks = append(bs.Bricks, b)
+}
+
+func (bs *BrickSet) Full() bool {
```
obnoxxx: The `Full()` methods are not used in this PR. Would we want to use them in `Add()` and prevent from over-adding? And maybe adding a (error) return value to `Add()`?

phlogistonjohn:
Added a godbc.Require check to Add and rebased.

apps/glusterfs/brick_allocate.go
```diff
@@ -26,6 +27,7 @@ func NewBrickSet(s int) *BrickSet {
 }
 
 func (bs *BrickSet) Add(b *BrickEntry) {
+	godbc.Require(!bs.Full())
```
obnoxxx: What about adding the same check for `DeviceSet.Add()`?

phlogistonjohn: Of course. I'm just being a blockhead.

phlogistonjohn:
Rebased & Updated w/ an additional improvement to variable names.

phlogistonjohn:
Commit msg updated. Var rename patch squashed.

PR #1078, opened 2018-03-02
[RFC] apps: separate allocator from ring loader

phlogistonjohn:
Prior to recent changes the allocator object was long lived and
constantly updated as nodes and devices were updated. This
was removed leading to the allocator only being loaded during a
single transaction and then discarded. However the allocator
struct still carried around a bunch of persistent state.
This change splits that state out of the allocator and into
a "private" struct that is only ever instantiated during the
db transaction (and thus needs no additional lock) and only
pulls in items from the cluster we want devices from.

### DIFF
--- a/apps/glusterfs/allocator_simple.go
+++ b/apps/glusterfs/allocator_simple.go
@@ -10,93 +10,76 @@
 package glusterfs
 
 import (
-	"sync"
-
 	"github.com/boltdb/bolt"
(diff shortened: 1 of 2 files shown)


phlogistonjohn:
See pr #1064 for additional context.

obnoxxx:
I continued this work in PR #1082 .

obnoxxx:
This has been merged as part of #1082

PR #1079, opened 2018-03-05
allow setting HEKETI_IGNORE_STALE_OPERATIONS to false

raghavendra-talur:
Earlier, setting the env var to anything would result in true. We
should rather parse it as bool allowing one to set it to "false".


phlogistonjohn:
Looks fine.

PR #1082, opened 2018-03-06
Cleanup allocator more

obnoxxx:
This is a continuation of the work begun in PR #1078.
It takes the patch from that PR and further simplifies and cleans up the code in simple_allocator.go.
Now the ring really only exists as stack variables below GetNodes.

### DIFF
--- a/apps/glusterfs/allocator_simple.go
+++ b/apps/glusterfs/allocator_simple.go
@@ -62,10 +62,21 @@ func loadRingFromDB(tx *bolt.Tx, clusterId string) (*SimpleAllocatorRing, error)
 			})
 		}
 	}
+
 	return ring, nil
 }
 
-func getDeviceList(ring *SimpleAllocatorRing, brickId string) (SimpleDevices, error) {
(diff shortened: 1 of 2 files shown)


obnoxxx:
As a consequence, we could (safely) use the patch from PR #1064 on top of this, to store the allocator in the app again.

obnoxxx:
Or alternatively, remove the whole instantiate an allocator when we start the app business from app.go ... :-)

phlogistonjohn:
> Or alternatively, remove the whole instantiate an allocator when we start the app business from app.go ... :-)

That's the approach I prefer.

obnoxxx:
@phlogistonjohn if you like it, feel free to merge.
I am writing on getting rid of the allocator stuff from app.go.

obnoxxx:
@phlogistonjohn wrote:
> I especially think that loadRingFromDB will come in handy in the upcoming arbiter work.

I hope that the signature change that this introduces does not turn out disadvantageous.
We might want to use GetDevicesFromDB instead.

PR #1084, opened 2018-03-06
apps: heavily refactor code where the allocator and replace bricks in volume func interact

phlogistonjohn:
This (long) series breaks up some of the code within the replaceBrickInVolume function in order to better prepare it for upcoming alternate allocation approaches. It also serves somewhat as a light disentangling of some of the db and allocation logic from the system probes and exec activity.

### DIFF
--- a/apps/glusterfs/brick_allocate.go
+++ b/apps/glusterfs/brick_allocate.go
@@ -58,9 +58,13 @@ type BrickAllocation struct {
 	DeviceSets []*DeviceSet
 }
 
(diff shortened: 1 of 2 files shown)


obnoxxx:
This centos-ci run seems to be stuck

retest this please

obnoxxx:
this does not work. need to wait for test to timeout and then retrigger ...

obnoxxx:
oh, it did in fact work (with some delay) :-)

PR #1085, opened 2018-03-06
Add Go 1.9.4 & 1.10 to Travis build

### LINKED ISSUE #1083: Add more, newer, versions of Go to travis test runs
Currently our travis tests are run with Go versions 1.7.5 and 1.8.2. As of this writing Go 1.10 is the newest released version.

According to the [Go release policy](https://golang.org/doc/devel/release.html#policy) none of the versions we're using are supported versions any more. I'm personally perfectly happy testing with older versions but it seems that we should at least add to the versions that we are testing with.

Atul9:
closes #1083

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,10 @@ matrix:
     env: OPTIONS="-race"
   - go: 1.8.2
     env: COVERAGE="true"
+  - go: 1.9.4
+    env: COVERAGE="true"
+  - go: "1.10"
+    env: COVERAGE="true"
 script:
 - make
 - if [[ -z "$COVERAGE" ]]; then make test TESTOPTIONS=-v ; else make test TESTOPTIONS=-vcstdout && bash .travis-coverage; fi


obnoxxx:
@Atul9 thanks for your patch!

ok to test

obnoxxx:
@Atul9 test failed.
There is an issue with travis: https://github.com/travis-ci/travis-ci/issues/9247
(1.10 is parsed as 1.1 ...)

.travis.yml
```diff
@@ -20,6 +20,10 @@ matrix:
     env: OPTIONS="-race"
   - go: 1.8.2
     env: COVERAGE="true"
+  - go: 1.9.4
+    env: OPTIONS="-race"
+  - go: 1.10
```
obnoxxx: According to https://github.com/travis-ci/travis-ci/issues/9247 , this needs to be quoted as
```
+ - go: "1.10"
```

phlogistonjohn:
Hi @Atul9 in addition to all that @obnoxxx said, I think we also need to decide on which options should go with which versions. Without doing too much research into it I think we want coverage on most versions and I'm not sure if OPTIONS="-race" is even doing anything ATM!

This is certainly heading in the right direction and I'm impressed by the super speedy contribution. :-)

Atul9:
@obnoxxx @phlogistonjohn done. I have updated the PR.

PR #1086, opened 2018-03-06
Cleanup allocator even more

obnoxxx:
This is based on PR #1082.
It takes the changes further and removes SetAllocator and ClearAllocator and merges Allocator() with newAllocator().

### DIFF
--- a/apps/glusterfs/app.go
+++ b/apps/glusterfs/app.go
@@ -495,18 +495,6 @@ func (a *App) Allocator() Allocator {
 	return a._allocator
 }
 
-// SetAllocator manually sets the allocator for this app.
-// The specified allocator will be cached on the app and
-// subsequent calls to Allocator will return the same object.
-// Generally this should only be used in test code.
-func (a *App) SetAllocator(allocator Allocator) {
-	if allocator == nil {
-		err := errors.New("use ClearAllocator to reset cached allocator")
-		panic(err)
-	}
-	a._allocator = allocator
-}
-
 // ClearAllocator resets the cached alloctor for this app.
 // This should be paired with calls to SetAllocator in order to
 // reset the app behavior to default when the test allocator is
--- a/apps/glusterfs/app.go
+++ b/apps/glusterfs/app.go
@@ -485,19 +485,8 @@ func (a *App) NotFoundHandler(w http.ResponseWriter, r *http.Request) {
 }
 
 // Allocator returns an allocator appropriate for the configuration
-// of this app. The allocator may be dynamically provided at
-// the time of the function call or cached from a prior call to
-// SetAllocator.
+// of this app.
 func (a *App) Allocator() Allocator {
-	if a._allocator == nil {
-		return a.newAllocator()
-	}
-	return a._allocator
-}
-
-// newAllocator returns a newly created allocator based on the
-// configuration of the app.
-func (a *App) newAllocator() Allocator {
 	var alloc Allocator
 	switch {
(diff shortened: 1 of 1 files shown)


obnoxxx:
I have more reworking in the pipeline that would change Allocator not to be a function under App, but we might want to discuss that separatetly... That would introduce a signature change

```
-func (a *App) Allocator() Allocator {
+func Allocator(name string) Allocator {
```
and then move it out of app.go --- not sure this is what we want...

apps/glusterfs/app.go
```diff
@@ -84,13 +84,13 @@ func NewApp(configIo io.Reader) *App {
...
-	case app.conf.Executor == "mock":
+	switch app.conf.Executor {
+	case "mock":
 		app.xo, err = mockexec.NewMockExecutor()
 		app.executor = app.xo
-	case app.conf.Executor == "kube" || app.conf.Executor == "kubernetes":
+	case "kube", "kubernetes":
 		app.executor, err = kubeexec.NewKubeExecutor(&app.conf.KubeConfig)
-	case app.conf.Executor == "ssh" || app.conf.Executor == "":
+	case "ssh", "":
```
phlogistonjohn: This is a nice cleanup but probably could have been in its  own standalone PR.

obnoxxx:
rebased after #1082 has been merged

apps/glusterfs/app.go
```diff
@@ -83,6 +83,15 @@ func NewApp(configIo io.Reader) *App {
...
 
+	switch app.conf.Allocator {
+	case "":
+		app.conf.Allocator = "simple"
+	case "simple":
+	default:
+		panic(errors.New("Invalid allocator specified: " + app.conf.Allocator))
+	}
+	logger.Info("Using %v allocator", app.conf.Allocator)
+
```
phlogistonjohn: This is the only patch in this series that I have any quibbles with. I am thinking that we should be moving away from having the app care about the allocator at all. I agree with the idea in your comment of decoupling the Allocator function from the app but I'd go further and just drop the function entirely and make the code that needs an allocator instance "get one". This should not be a problem now since there's only one allocator ATM, and we can just drop the config option entirely.

phlogistonjohn: If you think this is too much code churn at the moment a standalone `func GetAllocator() Allocator` would also be fine. FWIW I did a test where I removed the update of app.conf.Allocator and ran the unit tests. Everything passed; so this reassignment is probably superfluous and can be dropped.

apps/glusterfs/app.go
obnoxxx: True. I will split it out.

apps/glusterfs/app.go
obnoxxx: Sure, we don't currently need the config.
I just didn't want to do too big a removal in one step,  but if you are kind of sure that we will not need it any more, I'm more than happy to remove this altogether! Just let me know...

apps/glusterfs/app.go
obnoxxx: #1087

apps/glusterfs/app.go
obnoxxx: removed this patch for now

obnoxxx:
so i removed the questionable check, and will provide a different PR based on this that removes App.Allocator() altogether...

apps/glusterfs/app.go
obnoxxx: patch removed from this series

obnoxxx:
rebased

PR #1090, opened 2018-03-08
Prevent wasteful creation of too small block-hosting volumes

### LINKED ISSUE #1089: Requesting a block volume larger than the configured BlockHostingVolumeSize fails if large enough blockhosting volume has not been created.
### Observed behavior

Currently, if a block-hosting volume is requested to be created, with size larger than the BlockHostingVolumeSize, and a large enought block-hosting volume has not been created manually by the admin, a block-hosting volume of default size is created nonetheless, but the attempt to create the blockvolume then fails on the executor side.

### Expected behavior

Heketi should fail the request earlier and not wastefully create a block-hosting volume of inappropriate size.

obnoxxx:
Fixes #1089


obnoxxx:
updated to fix 1 test case

PR #1093, opened 2018-03-08
apps: deny creation of block-hosting volumes if disabled.

### LINKED ISSUE #1092: create_block_hosting_volumes has no effect any more
### Observed behavior

Setting create_block_hosting_volumes to false in the config has no effect in that block-hosting volumes are still auto-created.

### Expected behavior

When create_block_hosting_volumes is set to false in the config, an attempt to create a block-volume when there is no block-hosting volume, or when there is not enough space left on any block hosting volume should fail with a message telling that a block-hosting volume needs to be manually created.

obnoxxx:
In the move to operations, the check for the
CreateBlockHostingVolumes setting was somehow
dropped accidentially. Re-introduce it.

Fixes #1092

### DIFF
--- a/apps/glusterfs/block_volume_entry.go
+++ b/apps/glusterfs/block_volume_entry.go
@@ -46,6 +46,12 @@ func NewVolumeEntryForBlockHosting(clusters []string) (*VolumeEntry, error) {
 
 	vol := NewVolumeEntryFromRequest(&msg)
 
+	if !CreateBlockHostingVolumes {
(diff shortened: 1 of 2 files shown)


obnoxxx:
added a specific test case

obnoxxx:
fixed gofmt issue

phlogistonjohn:
> In the move to operations, the check for the
CreateBlockHostingVolumes setting was somehow
dropped accidentially.

Aw crud. I guess I had to introduce a regression eventually. :-(

PR #1095, opened 2018-03-08
github: Add a pull request template

obnoxxx:

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,25 @@
+<!--
+Thanks for sending a pull request! Your contribution is appreciated.
+
+Here are some tips for you:
+
+1. Read the contributing guide at [https://github.com/heketi/heketi/blob/master/docs/contributing.md]
+2. Split the changes up into minimal and atomic commits.  [https://github.com/heketi/heketi/blob/master/docs/contributing.md#splitting-your-change-into-commits]
+3. Write meaningful commit messages [https://github.com/heketi/heketi/blob/master/docs/contributing.md#good-commit-messages]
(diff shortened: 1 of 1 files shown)


obnoxxx:
Note that the comment will not be shown in the PR any more, only when editing the PR description.
Maybe this is a little too much text in the comment. This could also go into the contributing guide, but I thought it might be nice to have it all in in front of you when writing a PR.

Happy to reduce it / move some of it to a doc if that's preferred.

phlogistonjohn:
>Note that the comment will not be shown in the PR any more, only when editing the PR description.
Maybe this is a little too much text in the comment. This could also go into the contributing guide, but I thought it might be nice to have it all in in front of you when writing a PR.
>
>Happy to reduce it / move some of it to a doc if that's preferred.

Yeah, I feel that at this length its well into "TL;DR" territory. IMO what would be better is to convert it into a shorter bullet-point list and have the bullets (or items w/in the bullets) link to the contributing doc. 

So it might become something more of a quick checklist like:
```
* Does your PR encompass one development topic? [ref]
* Do your individual commits accomplish a single task? [ref]
* Do your commit messages describe the rationale for your change? [ref]
* Have you run 'make test'? [ref]
```
We could throw in a little more wording pre- and post-list but I find lists are more attention grabbing and keeping it short will increase the likelihood that it will be read.

I'd also argue that "thing" is too vague in this context but I'd rather wrangle that in a change to the contributing doc.

obnoxxx:
This is kind of dependent on PR #1096 as it now contains links to sections that only exist in the PR.

obnoxxx:
heh, this can't possibly have broken centos-ci ... ;-)

retest this please

obnoxxx:
retest this please

PR #1097, opened 2018-03-09
tests: add an environment var to disable sudo in functional tests

phlogistonjohn:
It can be inconvenient or downright wrong to always use sudo in
the functional tests, it depends on the local environment.
Add a new environment variable to allow all steps to run
as the current user.

### DIFF
--- a/tests/functional/lib.sh
+++ b/tests/functional/lib.sh
@@ -10,7 +10,7 @@ println() {
 }
 
 _sudo() {
-    if [ ${UID} = 0 ] ; then
+    if [[ ${UID} = 0 || "$HEKETI_TEST_USE_SUDO" = "no" ]]; then
         "${@}"
     else
         sudo -E "${@}"
--- a/tests/functional/README.md
+++ b/tests/functional/README.md
@@ -44,6 +44,11 @@ To run a specific functional test, go into that functionl test's directory and t
 $ ./run.sh
 ```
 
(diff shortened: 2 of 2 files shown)


obnoxxx:
shellcheck

phlogistonjohn:
> shellcheck

... is way too picky about a construct I've been using successfully for at least 8 years. Regardless, I should have run the tests before pushing. ;-)

tests/functional/lib.sh
```diff
@@ -10,7 +10,7 @@ println() {
 }
 
 _sudo() {
-    if [ ${UID} = 0 ] ; then
+    if [[ ${UID} = 0 || "$HEKETI_TEST_USE_SUDO" = "no" ]]; then
```
obnoxxx: It's actually curious that you did it differently from how shellcheck proposed and it still passed:
```
In ./tests/functional/lib.sh line 13:
    if [ ${UID} = 0 -o "$HEKETI_TEST_USE_SUDO" = "no" ] ; then
                    ^-- SC2166: Prefer [ p ] || [ q ] as [ p -o q ] is not well defined.
```

phlogistonjohn:
> It's actually curious that you did it differently from how shellcheck proposed and it still passed:

Yeah, I think the shelllcheck suggestion is really ugly. So I used the bash specific double brackets which has more well-defined behavior so shellcheck won't complain.

PR #1101, opened 2018-03-11
Snapshot design

obnoxxx:

This is the proposed design for volume snapshot and clone operations.

### DIFF
--- a/docs/design/snapshot.md
+++ b/docs/design/snapshot.md
@@ -1,0 +1,229 @@
+Snapshots are read-only copies of volumes (once the snapshot is activated) that
+can be used for cloning new volumes. It is possible to create a snapshot
+through the [Snapshot Create API](#create-a-snapshot) or the commandline
+client.
+
+From the command line client, you can type the following to create a snapshot
+from an existing volume:
+
+```
+$ heketi-cli volume snapshot <vol_uuid> [--name=<snap_name>]
+```
+
+The new snapshot can be used to create a new volume:
+
+```
+$ heketi-cli snapshot clone <snap_uuid> [--name=<clone_name>]
+```
+
+The clones of snapshots are new volumes with the same properties as the
+original. The cloned volumes can be deleted with the `heketi-cli volume delete`
+command. In a similar fashion, snapshots can be removed with the `heketi-cli
+snapshot delete` command.
+
+
+## Implementation Detail for Heketi internal Snapshot objects
+The proposed API and CLI can be used for both file-volumes and block-volumes.
+In order to provide the expected outcome of the cloning, the Heketi internal
+snapshot object should have a `Type: {volume|blockvolume}`. This makes it
+possible to have a single `/snapshots/{snapshot_uuid}` endpoint that can handle
+both volume types.
+
+
+# Proposed CLI
+
+```
+$ heketi-cli volume snapshot <vol_uuid> [--name=<snap_uuid>] [--description=<string>]
+$ heketi-cli snapshot clone <snap_uuid> [--name=<vol_uuid>]
+$ heketi-cli snapshot delete <snap_uuid>
+$ heketi-cli snapshot list
+$ heketi-cli snapshot info <snap_uuid>
+```
+
+Convenience call for direct volume cloning:
+
+```
+$ heketi-cli volume clone <vol_uuid>
+```
+
+
+# API Proposal
+
+The API is layed out here for file volume types only. The same API will be
+added for block-volumes at a later time.
+
+### Create a Snapshot
+* **Method:** _POST_
+* **Endpoint**:`/volumes/{volume_uuid}/snapshot`
+* **Content-Type**: `application/json`
+* **Response HTTP Status Code**: 202, See [Asynchronous Operations](#asynchronous-operations)
+* **Temporary Resource Response HTTP Status Code**: 303, `Location` header will contain `/snapshots/{snapshot_uuid}`. See [Snapshot Info](#snapshot_info) for JSON response.
+* **JSON Request**:
+    * name: _string_, _optional_, Name of snapshot. If not provided, the name of the snapshot will be `snap_{id}`, for example `snap_728faa5522838746abce2980`
+    * description: _string_, _optional_, Description of the snapshot. If not provided, the description will be empty.
+    * Example:
+
+```json
+{
+    "name": "midnight",
+    "description": "nightly snapshot"
+}
+```
+
+### Clone a Volume from a Snapshot
+* **Method:** _POST_
+* **Endpoint**:`/snapshots/{snapshot_uuid}/clone`
+* **Content-Type**: `application/json`
+* **Response HTTP Status Code**: 202, See [Asynchronous Operations](#asynchronous-operations)
+* **Temporary Resource Response HTTP Status Code**: 303, `Location` header will depend on the type of the snapshot object:
+    * For `file-volumes` the `Location` header will contain `/volumes/{id}`. See [Volume Info](#volume_info) for JSON response.
+    * For `block-volumes` the `Location` header will contain `/blockvolumes/{id}`. The `BlockVolumeInfo` JSON resonse is not documented.
+* **JSON Request**:
+    * name: _string_, _optional_, Name of volume. If not provided, the name of the volume will be `vol_{id}`, for example `vol_728faa5522838746abce2980`
+    * Example:
+
+```json
+{
+    "name": "new-vol-from-snap"
+}
+```
+
+### Delete a Snapshot
+* **Method:** _DELETE_
+* **Endpoint**:`/snapshots/{snapshot_uuid}`
+* **Response HTTP Status Code**: 202, See [Asynchronous Operations](#async)
+* **Temporary Resource Response HTTP Status Code**: 204
+
+### List Snapshots
+* **Method:** _GET_
+* **Endpoint**:`/snapshots`
+* **Response HTTP Status Code**: 200
+* **JSON Response**:
+    * snapshots: _array strings_, List of snapshot UUIDs.
+    * Example:
+
+```json
+{
+    "snapshots": [
+        "aa927734601288237463aa",
+        "70927734601288237463aa"
+    ]
+}
+```
+
+### Get Snapshot Information
+* **Method:** _GET_
+* **Endpoint**:`/snapshots/{snapshot_uuid}`
+* **Response HTTP Status Code**: 200
+* **JSON Request**: None
+* **JSON Response**:
+    * id: _string_, Snapshot UUID
+    * name: _string_, Name of the snapshot
+    * description: _string_, Description of the snapshot
+    * created: _int_, Seconds after the Epoch when the snapshot was taken
+    * volume: _string_, UUID of the volume that this snapshot belongs to
+    * cluster: _string_, UUID of cluster which contains this snapshot
+    * type: _string_, type of the original object that was snapshot'd, either `volume` or `blockvolume`
+    * Example:
+
+```json
+{
+    "id": "70927734601288237463aa",
+    "name": "midnight",
+    "description": "nightly snapshot",
+    "created": "1518712323",
+    "volume": "aa927734601288237463aa",
+    "cluster": "67e267ea403dfcdf80731165b300d1ca",
+    "type": "volume"
+}
+```
+
+### Clone a Volume directly
+
+This is a flavor that clones a volume directly.
+Implicitly, it creates a snapshot, activates
+and clones it, and then deletes the snapshot again.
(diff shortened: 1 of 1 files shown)


docs/design/snapshot.md
```diff
@@ -0,0 +1,185 @@
...
+# API Proposal
+
+The API is layed out here for file volume types only. The samy API will be
+added for block-volumes at a later time.
+
+### Create a Snapshot
+* **Method:** _POST_
+* **Endpoint**:`/volumes/{volume_uuid}/snapshots`
+* **Content-Type**: `application/json`
+* **Response HTTP Status Code**: 202, See [Asynchronous Operations](#asynchronous-operations)
```
humblec: I think this need not be an async call, any specific reason for making it async call?

obnoxxx:
@raghavendra-talur @phlogistonjohn here's the current snapshot proposal.

One question about the cli:
Would you rather have it structured like currently, i.e.
```
heketi-cli volume snapshot-create ID
heketi-cli volume snapshot-delete ID
heketi-cli volume snapshot-list
...
```

or rather have a sub-hierarchy like this

```
heketi-cli volume snapshot [--help]
heketi-cli volume snapshot create ID
heketi-cli volume snapshot delete ID
...
```

Opinions?

This is mostly cosmetic. I personally tend towards the second option.

phlogistonjohn:
@humblec I think every call that uses the heketi executor should be async. It's both consistent with the previous implementation of calls and avoids any http issues in case the operation on the gluster side is slower than we expect. So even if he operation in gluster normally is fast its "safer" to make it async.

humblec:
@phlogistonjohn Oh.. if all the calls are async , then its fine. thanks :+1:

docs/design/snapshot.md
nixpanic: Gluster needs to do some work on different storage servers, and that may require some time. Doing this async seems appropriate?

humblec:
one last question, did we change volume create call to async recently? or is it async from start?

docs/design/snapshot.md
```diff
@@ -0,0 +1,185 @@
...
+$ heketi-cli volume snapshot-create <vol_uuid> [--name=<snap_uuid>] [--description=<string>]
+$ heketi-cli volume snapshot-clone <snap_uuid> [--name=<vol_uuid>]
+$ heketi-cli volume snapshot-delete <snap_uuid>
+$ heketi-cli volume snapshot-list <vol_uuid>
+$ heketi-cli volume snapshot-info <snap_uuid>
+```
+
+# API Proposal
+
+The API is layed out here for file volume types only. The samy API will be
```
humblec: s/samy/same

obnoxxx:
@humblec 
> one last question, did we change volume create call to async recently? or is it async from start?

From the start.

docs/design/snapshot.md
```diff
@@ -0,0 +1,185 @@
...
+```json
+{
+    "name": "midnight",
+    "description": "nightly snapshot"
+}
+```
+
+### Clone a Volume from a Snapshot
+* **Method:** _POST_
+* **Endpoint**:`/volumes/{volume_uuid}/snapshots/{snapshot_uuid}/clone`
```
humblec: Do we have a cli for this ?

docs/design/snapshot.md
```diff
@@ -0,0 +1,185 @@
...
+    "id": "70927734601288237463aa",
+    "name": "midnight",
+    "description": "nightly snapshot",
+    "created": "1518712323",
+    "volume": "aa927734601288237463aa",
+    "cluster": "67e267ea403dfcdf80731165b300d1ca"
+}
+```
+
+### Clone a Volume directly
```
humblec: Do we have a cli for this ?

docs/design/snapshot.md
obnoxxx: thx. will update

docs/design/snapshot.md
obnoxxx: yes

docs/design/snapshot.md
```diff
@@ -0,0 +1,185 @@
...
+original. The cloned volumes can be deleted with the `heketi-cli volume delete`
+command. In a similar fashion, snapshots can be removed with the `heketi-cli
+volume snapshot-delete` command.
+
+# Proposed CLI
+
+```
+$ heketi-cli volume snapshot-create <vol_uuid> [--name=<snap_uuid>] [--description=<string>]
+$ heketi-cli volume snapshot-clone <snap_uuid> [--name=<vol_uuid>]
+$ heketi-cli volume snapshot-delete <snap_uuid>
```
humblec: Can we remove all snapshots via `--all` or similar switch ?

docs/design/snapshot.md
humblec: I dont see it in the doc, may be my bad eyes. :)

docs/design/snapshot.md
obnoxxx: Interesting thought. Not planned yet, but worth a thought. (Maybe for a future addition.)

docs/design/snapshot.md
obnoxxx: Updated the doc to mention it. (we had it in the started code already)

docs/design/snapshot.md
obnoxxx: fixed and update

docs/design/snapshot.md
```diff
@@ -0,0 +1,192 @@
+Snapshots are read-only copies of volumes (once the snapshot is activated) that
+can be used for cloning new volumes from. Its is possible to create a snapshot
+through the [Snapshot Create API](../api/api.md#create-a-snapshot) or the
+commandline client.
+
+From the command line client, you can type the following to create a snapshot
+from an existing volume:
+
+```
+$ heketi-cli volume snapshot-create <vol_uuid> [-name=<snap_name>]
```
phlogistonjohn: Do you want "--name" for consistency with other commands?

phlogistonjohn:
```
heketi-cli volume snapshot [--help]
heketi-cli volume snapshot create ID
heketi-cli volume snapshot delete ID
...
```

> Opinions?
>
> This is mostly cosmetic. I personally tend towards the second option.

I strongly prefer the 2nd form as well. In fact if we do have some reason not to allow "three tiered" commands I'd put the dash between volume and snapshot rather than snapshot and VERB.
IOW: "heketi-cli volume-snapshot create ID". But again, I'd rather have them all be space separated words like in your 2nd example.

docs/design/snapshot.md
nixpanic: Yes, that was the idea. However `snap_name` is something that can only be used once in a Gluster environment (where many different volumes may exist). So we'll keep the name in the Heketi DB only, and will not pass it on the `gluster` cmdline.

If there is no value in having a name, we can drop it.

nixpanic:
I've updated the design

1. to match the `heketi-cli volume snapshot <cmd>` format (was `heketi-cli volume snapshot-<cmd>`
1. with a note about a possible future extension to delete all snapshots from a volume (but a `--all` option might be inappropriate)

obnoxxx:
The only concern I have with the current design of the snapshot API and CLI is that it is slightly inconsistent in that a snapshot can be identified by its own UUID, but the operations acting on the snapshot still need to specify the volume id of which the snapshot was taken in the API, e.g. this is the url for operating on the snapshot with uuid snapshot_uuid
```
http(s)://heketi-address:heketiport/volumes/{volume_uuid}/snapshots/{snapshot_uuid}
```
while the corresponding cli would be
```
heketi-cli volume snapshot clone snapshot_uuid
heketi-cli volume snapshot delete snapshot_uuid
```
etc.

nixpanic:
Indeed, the `volume_uuid` is not needed for all snapshot operations, only for `create` and maybe `list`.

We could use `/snapshots/{snapshot_uuid}/...` for the 'pure' snapshot commands. But, when we have snapshots for block-volumes, we'll need to figure out how that fits in the grand scheme of things. I'm happy to change the endpoint to `/snapshots/` and in future we can figure out what type the `{snapshot_uuid}` refers to.

Opinions?

nixpanic:
The [workflow with the latest change](https://github.com/obnoxxx-collab/heketi/blob/dd7f23f6aa522e1acee71594c7430f1fe176cf73/docs/design/snapshot.md) might be cleaner. It boils down to having these two `volume` commands:

1. `heketi-cli volume clone <volume_uuid>`
2. `heketi-cli volume snapshot <volume_uuid>`

The 2nd command creates a snapshot object that can be manipulated through `heketi-cli snapshot ...` commands (with endpoint `/snapshots/{snapshot_uuid}`).

obnoxxx:
@nixpanic I like this a lot, thanks!

phlogistonjohn:
> I'm happy to change the endpoint to /snapshots/ and in future we can figure out what type the {snapshot_uuid} refers to.

Maybe this is a silly question, but does the block volume do anything special with snapshots? If not, then I see no issues with this.

obnoxxx:
@phlogistonjohn wrote:
> >    I'm happy to change the endpoint to /snapshots/ and in future we can figure out what type the {snapshot_uuid} refers to.
> 
> Maybe this is a silly question, but does the block volume do anything special with snapshots? If not, then I see no issues with this.

Not silly at all. On the api side this would just be the same.
In the backend, it would call out for different operations to create/work on the snapshots for volumes and block-volumes. I.e. there would need to be some branching of the form
```
if type == file-volume {
    do_the_file_volume_snapshot_operation()
} else {
    do_the_block_volume_snapshot_operation()
}
```
The reason is that the volume and blockvolume apis are separate.
I think this would be ok.

docs/design/snapshot.md
```diff
@@ -0,0 +1,227 @@
...
+
+
+# API Proposal
+
+The API is layed out here for file volume types only. The same API will be
+added for block-volumes at a later time.
+
+### Create a Snapshot
+* **Method:** _POST_
+* **Endpoint**:`/volumes/{volume_uuid}/snapshot`
```
nixpanic: @obnoxxx this lists `/volumes/{volume_uuid}/snapshot`, in your WIP implementation you used `.../snapshots` (plural). Which would be best? Maybe at one point we would like to list all snapshots of a particular volume (that has been omitted in this design for now). The proposed "Remove all Snapshots of a Volume" might suggest that `.../snapshots` is more suitable.

nixpanic:
This PR now has 11 commits that make the initial proposal to a pretty much final form. There is no description of the database contents yet. That will likely get added later once the implementation progresses and we have a better idea what is needed.

Do a :+1: on this comment and I'll squash all the patches into a single one that can be reviewed for merging. Thanks!

docs/design/snapshot.md
```diff
@@ -0,0 +1,229 @@
+Snapshots are read-only copies of volumes (once the snapshot is activated) that
+can be used for cloning new volumes from. Its is possible to create a snapshot
+through the [Snapshot Create API](#create-a-snapshot) or the commandline
+client.
+
+From the command line client, you can type the following to create a snapshot
+from an existing volume:
+
+```
+$ heketi-cli volume snapshot <vol_uuid> [-name=<snap_name>]
```
phlogistonjohn: Should probably be a double dash (--name)

docs/design/snapshot.md
```diff
@@ -0,0 +1,229 @@
...
+```
+
+The new snapshot can be used to create a new volume:
+
+```
+$ heketi-cli snapshot clone <snap_uuid> [-name=<clone_name>]
```
phlogistonjohn: Another single dash.

docs/design/snapshot.md
```diff
@@ -0,0 +1,229 @@
...
+
+```json
+{
+    "id": "70927734601288237463aa",
+    "name": "midnight",
+    "description": "nightly snapshot",
+    "created": "1518712323",
+    "volume": "aa927734601288237463aa",
+    "cluster": "67e267ea403dfcdf80731165b300d1ca"
+    "type": "volume"
```
phlogistonjohn: The line above lacks a comma

docs/design/snapshot.md
```diff
@@ -0,0 +1,229 @@
+Snapshots are read-only copies of volumes (once the snapshot is activated) that
+can be used for cloning new volumes from. Its is possible to create a snapshot
```
phlogistonjohn: The word from at the end of the first sentence is awkward and not needed.

"Its is" should be corrected.

nixpanic:
Thanks John! All remarks should have been resolved now.

@heketi/maintainers please review this (hopefully) final document.

nixpanic:
@raghavendra-talur do you any objections or changes you would like to see?

PR #1105, opened 2018-03-14
Arbiter brick placer

phlogistonjohn:

The first step towards supporting arbiter volumes. It builds upon the changes in PR #1103 and that PR should be merged first.

### Notes for the reviewer

Currently, this PR is meant as a preview of the first parts of arbiter support. It is intended to give to complain about things early. Currently, this arbiter brick placer is only unit tested and is not used anywhere else in the code. At the very least we need to be able to "know" a volume is an arbiter volume and also make executor changes. We can discuss if those should be in this PR or a later one.

### DIFF
--- a/apps/glusterfs/placer_arbiter.go
+++ b/apps/glusterfs/placer_arbiter.go
@@ -17,6 +17,13 @@ import (
 
 var (
 	tryPlaceAgain error = fmt.Errorf("Placement failed. Try again.")
+
+	// default discount size of an arbiter brick
+	ArbiterDiscountSize uint64 = 64 * KB
+)
+
+const (
+	arbiter_index int = 2
 )
 
 // ArbiterBrickPlacer is a Brick Placer implementation that can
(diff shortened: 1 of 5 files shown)


apps/glusterfs/placer_arbiter_test.go
```diff
@@ -0,0 +1,721 @@
...
+	tests.Assert(t, err == nil, "expected err == nil, got:", err)
+	tests.Assert(t, len(ba2.BrickSets) == 1,
+		"expected len(ba2.BrickSets) == 1, got:", len(ba2.BrickSets))
+	tests.Assert(t, len(ba2.BrickSets[0].Bricks) == 3,
+		"expected len(ba2.BrickSets[0].Bricks) == 3, got:",
+		len(ba2.BrickSets[0].Bricks))
+}
+
+func init() {
+	SetLogLevel("debug") // DELETEME
```
raghavendra-talur: :)

apps/glusterfs/placer_arbiter.go
```diff
@@ -231,7 +231,11 @@ func (bp *ArbiterBrickPlacer) tryPlaceBrickOnDevice(
 
 	logger.Debug("Trying to place brick on device %v", device.Info.Id)
 
-	for _, b := range bs.Bricks {
+	for i, b := range bs.Bricks {
+		// skip the brick at the current index (in case this is a replace)
```
raghavendra-talur: I don't seem to understand this commit. Is this making arbiter placer different than standard brick placer? If yes, why?

apps/glusterfs/placer_arbiter_test.go
raghavendra-talur: :)

raghavendra-talur:
The changes look good. I just need your help in understanding last commit and one init function in unit tests to be deleted.

apps/glusterfs/placer_arbiter.go
phlogistonjohn: This change tries to get the arbiter placer to behave more like standard placer. In the current, non-arbiter, approach a series of nodes and with bricks like n1/d1 n1/d2 n2/d1 n2/d2 (and so on) you can potentially replace a brick on n1/d1 with one on n1/d2. However, my initial arbiter could would have rejected n1/d2 because the old brick is still visible in the brick set and so it would see n1 in use and block it.

In current non-arbiter code the existing brick is filtered out from the brick set. We can do that easily there because the position of the bricks in the brick sets "doesn't matter".

I'll try to quickly improve the comment a bit.

apps/glusterfs/placer_arbiter_test.go
phlogistonjohn: I needed it while debugging test failures. I promptly forgot about it. I'll drop it from these changes.

phlogistonjohn:
Patches updated.

raghavendra-talur:
@phlogistonjohn I think merge of last PR has resulted in some conflicts. Please resolve and update. I don't see any more changes required.

phlogistonjohn:
Rebased onto master.

PR #1106, opened 2018-03-18
Initial support for creating and managing arbiter volumes

phlogistonjohn:

This series builds on the changes in PR #1105 and that series should be merged first.

This series adds initial support for creating and managing arbiter volumes. The core changes are the hooks that causes Heketi to select the arbiter brick placer and to request an arbiter volume from gluster. Additionally, a new function test subdir for arbiter tests is added with some simple tests for arbiter volume create and device remove (brick replace).

### Notes for the reviewer

The key name used for the "virtual" gluster option is intentionally ugly. We never finished discussing the api changes / naming in the design PR, so I picked something funky on purpose.

As discussed offline this change does nothing to try and dedicate devices for either data or arbiter bricks. That will be done later in a separate PR.

### DIFF
--- a/apps/glusterfs/volume_entry.go
+++ b/apps/glusterfs/volume_entry.go
@@ -14,6 +14,8 @@ import (
 	"encoding/gob"
 	"fmt"
 	"sort"
+	"strconv"
+	"strings"
 
 	"github.com/boltdb/bolt"
 	"github.com/heketi/heketi/executors"
@@ -36,6 +38,8 @@ const (
 	DEFAULT_EC_DATA               = 4
 	DEFAULT_EC_REDUNDANCY         = 2
 	DEFAULT_THINP_SNAPSHOT_FACTOR = 1.5
+
+	HEKETI_ARBITER_KEY = "user.heketi.arbiter"
 )
 
 // VolumeEntry struct represents a volume in heketi. Serialization is done using
@@ -217,6 +221,20 @@ func (v *VolumeEntry) Unmarshal(buffer []byte) error {
 	}
 
 	return nil
+}
+
+// HasArbiterOption returns true if this volume is flagged for
+// arbiter support.
+func (v *VolumeEntry) HasArbiterOption() bool {
+	for _, s := range v.GlusterVolumeOptions {
+		r := strings.Split(s, " ")
+		if len(r) == 2 && r[0] == HEKETI_ARBITER_KEY {
+			if b, e := strconv.ParseBool(r[1]); e == nil {
+				return b
+			}
+		}
+	}
+	return false
 }
 
 func (v *VolumeEntry) BrickAdd(id string) {
--- a/apps/glusterfs/volume_entry_create.go
+++ b/apps/glusterfs/volume_entry_create.go
@@ -115,6 +115,7 @@ func (v *VolumeEntry) createVolumeRequest(db wdb.RODB,
 	vr.Name = v.Info.Name
 	v.Durability.SetExecutorVolumeRequest(vr)
 	vr.GlusterVolumeOptions = v.GlusterVolumeOptions
+	vr.Arbiter = v.HasArbiterOption()
 
 	return vr, sshhost, nil
 }
--- a/apps/glusterfs/placer_volume.go
+++ b/apps/glusterfs/placer_volume.go
@@ -10,5 +10,8 @@
 package glusterfs
 
 func PlacerForVolume(v *VolumeEntry) BrickPlacer {
+	if v.HasArbiterOption() {
+		return NewArbiterBrickPlacer()
+	}
 	return NewStandardBrickPlacer()
 }
--- a/apps/glusterfs/volume_entry_test.go
+++ b/apps/glusterfs/volume_entry_test.go
@@ -2270,3 +2270,50 @@ func TestVolumeCreateConcurrent(t *testing.T) {
 	tests.Assert(t, brickCount == 27,
 		"expected brickCount == 27, got:", brickCount)
 }
+
+func TestVolumeCreateArbiter(t *testing.T) {
+	tmpfile := tests.Tempfile()
+	defer os.Remove(tmpfile)
+
+	// Create the app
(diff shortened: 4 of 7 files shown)


apps/glusterfs/volume_entry.go
```diff
@@ -36,6 +38,8 @@ const (
 	DEFAULT_EC_DATA               = 4
 	DEFAULT_EC_REDUNDANCY         = 2
 	DEFAULT_THINP_SNAPSHOT_FACTOR = 1.5
+
+	HEKETI_ARBITER_KEY = "x.heketi.arbiter"
```
raghavendra-talur: Gluster ignores(does not interpret any but stores them in gluster database) option that is form "user.*". I was thinking of using user.heketi-* as a namespace for all things heketi.

Something like user.heketi-arbiter=yes  OR user.heketi-arbiter=1, based on whether we want it to be a int or bool.

phlogistonjohn: Excellent, I was looking for something like that but couldn't find it in the gluster docs. I did try passing it the current key which was rejected. Does it have to have only one dot? If not I slightly prefer "user.heketi.arbiter" and I would go with any value but true-ness is indicated by any value that parses to true with go's [strconv.Parsebool](https://golang.org/pkg/strconv/#ParseBool).

raghavendra-talur: Verified in gluster code that user.heketi.* should work. I prefer dot as a separator too, just hadn't see options with dot with them in gluster.

apps/glusterfs/volume_entry.go
```diff
@@ -219,6 +223,20 @@ func (v *VolumeEntry) Unmarshal(buffer []byte) error {
 	return nil
 }
 
+// HasArbiterOption returns true if this volume is flagged for
+// arbiter support.
+func (v *VolumeEntry) HasArbiterOption() bool {
+	for _, s := range v.GlusterVolumeOptions {
+		r := strings.Split(s, " ")
```
raghavendra-talur: I thought glusterVolumeOptions are comma separated?

apps/glusterfs/volume_entry_create.go
```diff
@@ -114,7 +114,14 @@ func (v *VolumeEntry) createVolumeRequest(db wdb.RODB,
 	// Setup volume information in the request
 	vr.Name = v.Info.Name
 	v.Durability.SetExecutorVolumeRequest(vr)
-	vr.GlusterVolumeOptions = v.GlusterVolumeOptions
+	// filter out undesired options from gluster (TOO HACKY?)
+	vr.GlusterVolumeOptions = []string{}
+	for _, s := range v.GlusterVolumeOptions {
```
raghavendra-talur: Let us not filter out the arbiter option. It would be really helpful in debugging if gluster stored what heketi knew about the volume. As it is user.heketi. it is harmless.

tests/functional/TestArbiterVolume/tests/heketi_test.go
```diff
@@ -0,0 +1,460 @@
...
+}
+
+func TestConnection(t *testing.T) {
+	r, err := http.Get(heketiUrl + "/hello")
+	tests.Assert(t, err == nil, "expected err == nil, got:", err)
+	tests.Assert(t, r.StatusCode == http.StatusOK)
+}
+
+func PauseBeforeTeardown() {
+	s := os.Getenv("HEKETI_TEST_PAUSE_BEFORE_TEARDOWN")
```
raghavendra-talur: We probably need a separate patch for this :)

tests/functional/TestArbiterVolume/config/heketi.json
```diff
@@ -0,0 +1,20 @@
+{
```
raghavendra-talur: As a commit wide comment, I am not sure if I would want to add one more suite here. Mainly because I would like the infra code to be the same and am already working on a patch to merge 4 Vagrant based suites we have. Do we had a valid reason for new suite?

raghavendra-talur:
@phlogistonjohn  There are some things which need changes but this is excellent work! Arbiter support in heketi, yay!

apps/glusterfs/volume_entry.go
phlogistonjohn: On the command line, between k-v pairs, sure. But from what I could tell in the code and the existing examples the key and the value are separated by a space. A couple examples:
```
apps/glusterfs/block_volume_entry.go:   msg.GlusterVolumeOptions = []string{"group gluster-block"}
tests/functional/TestSmokeTest/tests/heketi_test.go:    volReq.GlusterVolumeOptions = []string{"performance.rda-cache-limit 10MB"}
```
From what I can tell the strings are passed to the gluster command line as is and gluster expects something like "gluster --mode=script volume set VOLNAME KEY VALUE".

apps/glusterfs/volume_entry_create.go
phlogistonjohn: Sure. I will make that change.

tests/functional/TestArbiterVolume/config/heketi.json
phlogistonjohn: I was a bit on the fence about that as well. I figured in the worst case it would be easier to merge things back together rather than split them apart. Plus, heketi_test.go feels a bit crammed full of stuff. However, we can still consolidate the suites but keep separate .go files. Does that sound like a reasonable compromise?

raghavendra-talur: Yes, for now if you could rename the test file to arbiter_test.go or similar and add it to smoke test it would be sufficient.

apps/glusterfs/volume_entry.go
raghavendra-talur: You are right! I did not check properly.

phlogistonjohn:
I've updated the main body of the code but will need a little more time to reconcile the test suites.

phlogistonjohn:
I have replaced the new suite of tests with additions to the TestSmokeTest dir.

PR #1107, opened 2018-03-19
apps: add comments to the operations interface

phlogistonjohn:
This change adds comments to all of the functions defined
by the operations interface to hopefully aid future operation
authors in understanding what each part of the interface
is meant to do.


phlogistonjohn:
Ping: @nixpanic as per our earlier conversation. Let me know if there's anything unclear here and I'll try to improve the patch.

PR #1108, opened 2018-03-19
Add TLS support for Heketi Server

rsevilla87:
Adding TLS support for Heketi Server. This functionality is crucial in some environments where all communications should be encrypted.


raghavendra-talur:
@rsevilla87 thanks for the PR. I have had a look and change looks good. I am taking time to read a little more on the docs. We will have review comments soon!

phlogistonjohn:
Ideally, we'd want some tests for this feature but it's a small enough patch that looks straightforward to me.

phlogistonjohn:
@centos-ci  ok to test

rsevilla87:
@phlogistonjohn ping :)

phlogistonjohn:
Hi @rsevilla87 we usually like to have two approvals before merging. I'll ping some other team members to get a 2nd review.

raghavendra-talur:
This probably requires changes/updates in heketi client library for feature completion but the PR is complete as far as heketi server functionality is concerned.

phlogistonjohn:
> There is no client code, so it is rather useless right now. And if the server enables TLS, the current client won't be able to connect any more - correct?

The default Go client will support HTTPS automatically. If the cert is signed by a CA the client knows about I don't think any additional code changes would be needed. However, to best use the client we probably want to make it easy to feed it options for skipping verification or feeding it a custom cert for verification.

>  There is not real test yet, of functionality through a TLS connection, just a test of being able to connect. (Of course this is kind of implied by no 1.)

Sure.

> The way to specify the certificate/key files is very static through the config file. Hence it's harder in a kubernetes environment to specify them. Could we make it overrideable by environment variables in apps/glusterfs/app.go (like for HEKETI_EXECUTOR and others)?

This seems OK for this PR. It would definitely make things more easier to set up w.r.t. adding new certs/keys to the heketi pod configuration.

> I would at least want to have client code alongside the server code. Otherwise this might make it into the release. Tests and config flexibility can be added later.

Personally, I had figured we'd do the client stuff in a follow up pr, but if you feel strongly enough about I'm willing to help add some of the needed configuration ability to the client apis.

obnoxxx:
Thanks again for your contribution @rsevilla87 !

PR #1109, opened 2018-03-20
consolidate server configuration

phlogistonjohn:

This creates a new package "server/config" and locates all the logic around parsing configuration files into this single location. Instead of parsing the configuration file twice we parse the file once and pass the configuration object around.

Using a new package also provides a foundation for future enhancements around configuring the heketi server process.

### Notes for the reviewer

In order to keep this initial series small it does not tackle re-organizing the environment variables that are spread throughout the codebase. I'm happy to bikeshed the new location but I do think it should be a new package for handling the config loading.

### DIFF
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -15,7 +15,7 @@ import (
 	"io"
 	"os"
 
-	_ "github.com/heketi/heketi/apps/glusterfs"
+	"github.com/heketi/heketi/apps/glusterfs"
 	"github.com/heketi/heketi/middleware"
 )
 
@@ -27,6 +27,9 @@ type Config struct {
 	EnableTls            bool                     `json:"enable_tls"`
 	CertFile             string                   `json:"cert_file"`
 	KeyFile              string                   `json:"key_file"`
+
+	// pull in the config sub-object for glusterfs app
+	GlusterFS *glusterfs.GlusterFSConfig `json:"glusterfs"`
 }
 
 func ParseConfig(input io.Reader) (config *Config, e error) {
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -1,0 +1,54 @@
+//
+// Copyright (c) 2018 The heketi Authors
+//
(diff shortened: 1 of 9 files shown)


obnoxxx:
That looks really nice! Need to look a little deeper for a proper review.

phlogistonjohn:
FWIW there are a bunch of other outstanding PRs that touch configuration. I'm not going to rebase this until some of those get merged and I think some of the near-term config churn is quieting down.

phlogistonjohn:
Rebased.

obnoxxx:
Sorry @phlogistonjohn , this needs another rebase..

phlogistonjohn:
Rebased.

server/config/config.go
```diff
@@ -0,0 +1,57 @@
...
+	Port                 string                   `json:"port"`
+	AuthEnabled          bool                     `json:"use_auth"`
+	JwtConfig            middleware.JwtAuthConfig `json:"jwt"`
+	BackupDbToKubeSecret bool                     `json:"backup_db_to_kube_secret"`
+	EnableTls            bool                     `json:"enable_tls"`
+	CertFile             string                   `json:"cert_file"`
+	KeyFile              string                   `json:"key_file"`
+
+	// pull in the config sub-object for glusterfs app
+	GlusterFS *glusterfs.GlusterFSConfig `json:"glusterfs"`
```
obnoxxx: Why is this hunk not added when this module is introduced, but in the patch "server: consolidate...." ?

phlogistonjohn: I was trying to refactor one area at a time. So the first patch touched on stuff that was in main.go and then later patches brings in the stuff from apps/glusterfs. Do you think there's a better approach than that?

obnoxxx: Ok, that makes sense... :-)

phlogistonjohn:
Updated.

obnoxxx:
retest this please

PR #1111, opened 2018-03-22
client: Add 'volume clone' CLI and API

nixpanic:
This change depends on the service-side implementation that is currently still being cleaned up. The API and CLI for the volume cloning is expected to be stable, other projects will be able to consume the current state.

@humblec, is this what you need for the [gluster-fiile external storage provider](https://github.com/kubernetes-incubator/external-storage/tree/master/gluster/file)?

### DIFF
--- a/client/api/go-client/volume.go
+++ b/client/api/go-client/volume.go
@@ -232,3 +232,52 @@ func (c *Client) VolumeDelete(id string) error {
 
 	return nil
 }
+
+func (c *Client) VolumeClone(id string, request *api.VolumeCloneRequest) (*api.VolumeInfoResponse, error) {
+	// Marshal request to JSON
+	buffer, err := json.Marshal(request)
+	if err != nil {
+		return nil, err
+	}
+
+	// Create a request
+	req, err := http.NewRequest("POST", c.host+"/volumes/"+id+"/clone", bytes.NewBuffer(buffer))
+	if err != nil {
+		return nil, err
+	}
+	req.Header.Set("Content-Type", "application/json")
+
+	// Set token
+	err = c.setToken(req)
+	if err != nil {
+		return nil, err
+	}
+
+	// Send request
+	r, err := c.do(req)
+	if err != nil {
+		return nil, err
+	}
+	defer r.Body.Close()
+	if r.StatusCode != http.StatusAccepted {
+		return nil, utils.GetErrorFromResponse(r)
+	}
+
+	// Wait for response
+	r, err = c.waitForResponseWithTimer(r, time.Second)
+	if err != nil {
+		return nil, err
+	}
+	if r.StatusCode != http.StatusOK {
+		return nil, utils.GetErrorFromResponse(r)
+	}
+
+	// Read JSON response
+	var volume api.VolumeInfoResponse
+	err = utils.GetJsonFromResponse(r, &volume)
+	if err != nil {
+		return nil, err
+	}
+
+	return &volume, nil
+}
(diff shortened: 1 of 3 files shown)


nixpanic:
@heketi/maintainers please add the [do not merge] label for now

phlogistonjohn:
"do not merge" label added as you requested. Just let us know when you want to drop it.

pkg/glusterfs/api/types.go
```diff
@@ -330,6 +330,16 @@ func (volExpandReq VolumeExpandRequest) Validate() error {
 	)
 }
 
+type VolumeCloneRequest struct {
+	Name string `json:"name,omitempty"`
```
phlogistonjohn: We don't actually clone by name do we? Generally, I would have expected a heketi volume Id here.

nixpanic: No, this is the name of the volume that is being created from the original. By default it is `vol_{UUID}`, but users can pass it an optional `--name=...` just like with a `heketi-cli volume create --name=...` command.

nixpanic: Note that the `VolumeCloneRequest` is sent to the `/volume/{uuid}/clone` endpoint, where the `{uuid}` is the reference to the source/original volume to be cloned.

phlogistonjohn: Yup. Got it. Thanks for the clarification. Something like this may be useful as a comment.

client/api/go-client/volume.go
```diff
@@ -232,3 +232,52 @@ func (c *Client) VolumeDelete(id string) error {
 
 	return nil
 }
+
+func (c *Client) VolumeClone(id string, request *api.VolumeCloneRequest) (*api.VolumeInfoResponse, error) {
```
humblec: Ideally we can only keep `request` as in volume create(), we dont need `id` ( which is the source volume ID iic) and also `request` together. @phlogistonjohn @nixpanic  thoughts? or its required to have it compulsory for REST URL route?

pkg/glusterfs/api/types.go
humblec: The volumeClone request can also have `size` field in it and also `cluster` field , I am not sure cloning to a different cluster is possible , if not we dont need cluster field. Also this could have an optional `snap` field, so if user specify the `snap` name, we could take clone from it. @nixpanic @phlogistonjohn comments?

client/api/go-client/volume.go
phlogistonjohn: It seems like the new call simply follows the existing pattern (for example VolumeExpand). I don't see an issue with it as-is.

pkg/glusterfs/api/types.go
nixpanic: It is not possible to do clones from one cluster to an other. Even the clones use the same block devices as the original bricks. This is just how Gluster implements volume cloning.

The extended snapshot API that is not included in the PR, offers the full create-snapshot and clone-volume-from-snapshot options. It is still planned to add that, but at a later stage. See the design from commit 353e8bdb8 for all operations.

obnoxxx:
here is a beginning of test code:

https://github.com/obnoxxx/heketi/commit/2f66d9eabbabb2111aab7db3c4d7b0b84003c286
https://github.com/obnoxxx/heketi/commits/clone-api-client-test

nixpanic:
In case someone feels like testing this out, the server-side changes are available as in a work-in-progress tree at https://github.com/obnoxxx-collab/heketi (snapshot-wip branch). Some of those internals will change, but API and functionality wise it should be stable.

nixpanic:
This change is also part of #1134. Just merging that one is sufficient and makes things a little easier.

PR #1118, opened 2018-03-28
executor: revert add-brick if rebalance fails

### LINKED ISSUE #1116: roll back of volume expansion is incomplete
### Kind of issue

Bug

### Observed behavior
When a executor operation fails for volume expansion, rollback is not performed for add-brick commands and only the bricks are cleaned up.

### Expected/desired behavior
A proper roll back of add-brick has to be performed to bring the state of volume on gluster to what it was before add-brick and then the newly created bricks should be cleaned up.

Create a volume using heketi. Kill one of the bricks belonging to the volume on a node. Perform volume expansion and check the volume info on heketi and gluster. Both don't match. Gluster shows all bricks(including the newly added ones) while heketi shows only old bricks.

- Heketi version used :  v6.0.0

raghavendra-talur:
If rebalance fails, earlier we were considering the whole add-brick
operation to be a failure. However, this would clean the bricks without
reverting the state of volume on gluster.

With this change, if rebalance on a volume fails after add-brick, we try
to revert the add-brick operation. If successful to revert, we can go
ahead with a proper Rollback. If not successful, we fake a successful
add-brick; this ensures gluster and heketi are in sync and gives admin a
chance to initiate rebalance manually.

### Does this PR fix issues?
Fixes #1116 

### Notes for the reviewer

Note the fake success done to bring gluster and heketi in sync.

### DIFF
--- a/executors/cmdexec/volume.go
+++ b/executors/cmdexec/volume.go
@@ -12,6 +12,7 @@ package cmdexec
 import (
 	"encoding/xml"
 	"fmt"
+	"time"
 
 	"github.com/heketi/heketi/executors"
 	"github.com/lpabon/godbc"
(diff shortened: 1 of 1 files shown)


executors/cmdexec/volume.go
```diff
@@ -78,6 +79,99 @@ func (s *CmdExecutor) VolumeCreate(host string,
...
+		return fmt.Errorf("Could not revert add-brick")
+	}
+	logger.Debug("%+v\n", startOutput)
+	if startOutput.OpRet != 0 {
+		logger.LogError("failed to start remove-brick, manual cleanup might be required")
+		return fmt.Errorf("Could not revert add-brick")
+	}
+
+	commands = s.createRemoveBrickCommands(volume, 0, inSet, maxPerSet, "status --xml")
+	ticker := time.NewTicker(time.Second * 10)
```
phlogistonjohn: I'd prefer to see `ticker.Stop()` in a defer statement rather than embedded in the later code blocks.

executors/cmdexec/volume.go
```diff
@@ -78,6 +79,99 @@ func (s *CmdExecutor) VolumeCreate(host string,
...
+				logger.LogError("Unable to determine remove-brick commit output")
+				return fmt.Errorf("Could not revert add-brick")
+			}
+			logger.Debug("%+v\n", commitOutput)
+			if commitOutput.OpRet != 0 {
+				return fmt.Errorf("Could not revert add-brick")
+			}
+			return nil
+		}
+		logger.Warning("remove-brick status is not complete yet, will recheck after few seconds")
```
phlogistonjohn: I'd print the number of seconds or nothing. The words "a few seconds" don't add a lot of value here.

executors/cmdexec/volume.go
```diff
@@ -78,6 +79,99 @@ func (s *CmdExecutor) VolumeCreate(host string,
...
+
+	commands = s.createRemoveBrickCommands(volume, 0, inSet, maxPerSet, "status --xml")
+	ticker := time.NewTicker(time.Second * 10)
+	select {
+	case <-ticker.C:
+		removeBrickStatusComplete = s.isRemoveBrickComplete(commands, host)
+		if removeBrickStatusComplete {
+			ticker.Stop()
+			commands := s.createRemoveBrickCommands(volume, 0, inSet, maxPerSet, "commit --xml")
+			commitOutputXML, err := s.RemoteExecutor.RemoteCommandExecute(host, commands, 10)
```
phlogistonjohn: It seems to me that all the command execution and xml parsing should be rolled into yet another helper func.

executors/cmdexec/volume.go
```diff
@@ -205,6 +308,35 @@ func (s *CmdExecutor) createAddBrickCommands(volume *executors.VolumeRequest,
...
 }
 
+func (s *CmdExecutor) createRemoveBrickCommands(volume *executors.VolumeRequest,
+	start, inSet, maxPerSet int, verb string) []string {
+
+	commands := []string{}
+	var cmd string
+
+	// Go through all the bricks and create remove-brick commands
+	for index, brick := range volume.Bricks[start:] {
```
phlogistonjohn: FWIW, I find this loop somewhat difficult to read. The way the code flows feels a bit "backwards" to me. I don't have any concrete suggestions at the moment but I wanted to point this out because if now me finds it more challenging a future reader may too.

executors/cmdexec/volume.go
raghavendra-talur: thanks, this was early comment left out as is. Will add specific value.

obnoxxx:
Part of this has been superseded by PR #1119. There is a plan to re-do this differently by not calling rebalance from expansion at all, but flagging the expanded volume as `requires-rebalance` (or so) and having rebalance triggered (and possibly monitored) lazily from a thread watching for volumes requiring rebalance.

@raghavendra-talur @phlogistonjohn : do we want to re-use this PR for this or create a new one?

raghavendra-talur:
@obnoxxx I think closing this PR makes sense. Going by the new approach there is nothing common here. I will close this.

PR #1122, opened 2018-03-29
PoC: monitor node status and update ring

Madhu-1:
Will add more info in next update

Co-authored-by: Raghavendra Talur <rtalur@redhat.com>

### DIFF
--- a/apps/glusterfs/allocator_simple.go
+++ b/apps/glusterfs/allocator_simple.go
@@ -84,8 +84,7 @@ func loadRingFromDeviceSource(dsrc DeviceSource) (
 	return ring, nil
 }
 
-func getDeviceListFromDB(db wdb.RODB, clusterId,
-	brickId string) (SimpleDevices, error) {
+func getDeviceListFromDB(db wdb.RODB, clusterId string) (*SimpleAllocatorRing, error) {
 
 	var ring *SimpleAllocatorRing
 	err := db.View(func(tx *bolt.Tx) (e error) {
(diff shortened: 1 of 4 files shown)


apps/glusterfs/allocator_simple.go
```diff
@@ -95,23 +94,56 @@ func getDeviceListFromDB(db wdb.RODB, clusterId,
...
+	return ring, nil
+}
 
+func GetRebalancedDeviceList(ring *SimpleAllocatorRing, brickId string) (SimpleDevices, error) {
 	devicelist := ring.GetDeviceList(brickId)
 
 	return devicelist, nil
 }
 
+func filterOfflineDevices(ring *SimpleAllocatorRing, clusterId, brickId string) *SimpleAllocatorRing {
```
phlogistonjohn: To me, it feels like this logic ought to be part of the same logic that populates the ring, rather than a secondary step that updates the node when it is built.

IOW, every place we're testing node.isOnline as input to the ring we can add a check on this node health.

apps/glusterfs/allocator_simple.go
```diff
@@ -95,23 +94,56 @@ func getDeviceListFromDB(db wdb.RODB, clusterId,
...
+	//edit ring
+	NodesStatuses.Mutex.RLock()
+	nodelist := NodesStatuses.Status
+	NodesStatuses.Mutex.Unlock()
```
phlogistonjohn: IMO, the locking should be "hidden" as part of the node status implementation.

apps/glusterfs/app_config.go
```diff
@@ -36,6 +36,7 @@ type GlusterFSConfig struct {
 
 	// server behaviors
 	IgnoreStaleOperations bool `json:"ignore_stale_operations"`
+	MonitorGlusterNodes   bool `json:"monitor_gluster_nodes"`
```
phlogistonjohn: Nice. My attempt did not have much in the way of configuration.

apps/glusterfs/watcher.go
```diff
@@ -0,0 +1,92 @@
...
+
+			if err != nil {
+				//pass on to next node
+				continue
+
+			}
+
+			// should we check status of nodes disabled in db
+			//Ignore if the node is not online
+			if !newNode.isOnline() {
```
phlogistonjohn: I like this! My attempt was not checking this value but it makes sense not check on nodes that are administratively disabled.

obnoxxx:
Was this superseded by #1125 ?

phlogistonjohn:
> Was this superseded by #1125 ?

IMO, yes it is.

raghavendra-talur:
Superseded by #1125

PR #1123, opened 2018-03-29
Internally retry volume creation

phlogistonjohn:

This set of changes adds an internal retry to the volume create operation, based on the previously submitted generic operations retry mechanism. It cleans up a few issues found while testing the rety and adds a specific test for the volume create retry.

This change depends on PR #1056.

### Notes for the reviewer

I've opted to use a default value of 4 retries. This means a total of 5 tries. In the test itself I upped the retry count to ten because we want to make sure the test always passes and the mock executor doesn't really take any time.


obnoxxx:
This needs rebase

phlogistonjohn:
Rebased.

PR #1128, opened 2018-04-04
middleware: new middleware to add request id

raghavendra-talur:
Each request that hits the router will get a request id attached to its
context. We can later use it to tie all operations happening for the
request together.

Once the https://github.com/heketi/rest/pull/2 is merged, we will use it to pass the same request id a location id for redirect function.


middleware/request_id.go
```diff
@@ -0,0 +1,36 @@
...
+// GetRequestID returns the request id from HTTP context.
+func GetRequestID(ctx context.Context) string {
+	reqID, _ := ctx.Value(requestIDKey).(string)
+	return reqID
+}
+
+func (reqID *RequestID) ServeHTTP(w http.ResponseWriter, r *http.Request, next http.HandlerFunc) {
+
+	newCtx := context.WithValue(r.Context(), requestIDKey, utils.GenUUID())
+	next(w, r.WithContext(newCtx))
```
phlogistonjohn: This seems pretty cool. The only thing that I wonder about is if we should avoid generating these IDs for GETs. GETs should not need and IDs and they're probably the most common request. Skipping the ID assignment on a GET might be more efficient and might help us avoid accidentally making changes in response to GETs.

raghavendra-talur: Good one, Thanks! Done.

phlogistonjohn:
Travis test failures looks real.

glide.yaml
```diff
@@ -8,6 +8,7 @@ import:
 - package: github.com/gorilla/context
 - package: github.com/gorilla/mux
 - package: github.com/heketi/rest
+  version: aa6a652074133e6b0836bb095e847021fcb94b6a
```
obnoxxx: This is a remnant of the original patch in this PR.
The required change (to glide.lock) got merged in PR #1133.
This hunk is not needed.
@Madhu-1 please remove this patch.

obnoxxx: This hunk is a remnant of the original patch that got merged as PR #1133.
This hunk is not needed (as discussed).

@raghavendra-talur please remove this hunk!

raghavendra-talur:
The main intention was to get an ID for a request which could later be used as a prefix for log lines etc to track all operations performed for a request.

Instead of using a request ID generated in middleware we should probably use the id which is generated for operations. 

I will close this PR now.

PR #1129, opened 2018-04-04
Add concept of tags to devices and nodes

phlogistonjohn:

Tags are short-ish user controlled metadata
that Heketi can choose to interpret but are otherwise free-form. We will be using our first well known tag "arbiter" to control how the system places arbiter bricks.

The following features have been implemented:
* Ability to store and retrieve tags from the node and device entries in the DB.
* Utility file for tags management
* Server API for setting tags on nodes and devices
* Go client api for setting tags on nodes and devices
* heketi-cli subcommands to set and remove tags on nodes and devices
* heketi-cli support for tags within the topology file
* Support for the arbiter tag
* Hooks for the arbiter placer to use the tag
* Functional tests for tags and arbiter tag

### DIFF
--- a/client/api/python/heketi/heketi.py
+++ b/client/api/python/heketi/heketi.py
@@ -23,6 +23,12 @@ import time
 import json
 
 
+# Constants related to setting tags
+TAGS_SET = 'set'
+TAGS_UPDATE = 'update'
+TAGS_DELETE = 'delete'
+
+
 class HeketiClient(object):
 
     def __init__(self, host, user, key):
@@ -158,6 +164,16 @@ class HeketiClient(object):
         return req.status_code == requests.codes.ok
         '''
 
+    def node_set_tags(self, node_id, tags_options):
+        '''Set, update, or delete tags on a node.
+        Specify tags with options key "tags" and a dict of tags & values,
+        specify change type with options key "change_type" and a
+        string value of "set", "update", "delete" (or use TAGS_* constants).
+        '''
+        uri = '/nodes/' + node_id + '/tags'
+        req = self._make_request('POST', uri, tags_options)
+        return req.status_code == requests.codes.ok
+
     def device_add(self, device_options={}):
         ''' device_options is a dict with parameters to be passed \
             in the json request: \
@@ -191,6 +207,16 @@ class HeketiClient(object):
         uri = '/devices/' + device_id + '/resync'
         req = self._make_request('GET', uri)
         return req.status_code == requests.codes.NO_CONTENT
+
+    def device_set_tags(self, device_id, tags_options):
+        '''Set, update, or delete tags on a device.
+        Specify tags with options key "tags" and a dict of tags & values,
+        specify change type with options key "change_type" and a
+        string value of "set", "update", "delete" (or use TAGS_* constants).
+        '''
+        uri = '/devices/' + device_id + '/tags'
+        req = self._make_request('POST', uri, tags_options)
+        return req.status_code == requests.codes.ok
 
     def volume_create(self, volume_options={}):
         ''' volume_options is a dict with volume creation options:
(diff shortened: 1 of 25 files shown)


phlogistonjohn:
The PR is nearly complete so I decided to throw it out there for wider circulation. What is still to-do and/or needs more discussion:

* Whether we need the tag updates to be atomic (done on the server) rather than done in the CLI as it is now.
* Discussion around the priority of tags on nodes vs. devices. I'm happy with it as is, but I'm willing to be convinced that additional changes are still needed.
* Finalizing the CLI interface for the new sub-commands (iow: last minute bike-shedding ;-))
* Python client api for tags
* CLI documenation

client/cli/go/cmds/device.go
```diff
@@ -301,3 +315,52 @@ var deviceResyncCommand = &cobra.Command{
...
+				return heketi.DeviceSetTags(id, &api.TagsChangeRequest{
+					SetTags: tags,
+				})
+			})
+	},
+}
+
+var deviceRmTagsCommand = &cobra.Command{
+	Use:     "rmtags [device_id] tag1:value1 tag2:value2...",
+	Aliases: []string{"deltags", "removetags"},
```
raghavendra-talur: nice!

client/cli/go/cmds/utils.go
```diff
@@ -0,0 +1,115 @@
...
+	//ensure proper number of args
+	s := cmd.Flags().Args()
+	if len(s) < 1 {
+		return errors.New("id required")
+	}
+	if len(s) < 2 {
+		return errors.New("at least one tag:value pair expected")
+	}
+
+	exact, err := cmd.Flags().GetBool("exact")
```
raghavendra-talur: This is what I had missed in the previous review. My bad.

phlogistonjohn:
As discussed in a call we've decided to stick with the current approach to tag priority/inheritance.
We also decided that we need to update the tags api for support transactional updates.

phlogistonjohn:
I've updated the API and related code to support "change types" of set/update/delete when changing tags on an existng item. Please let me know if this seems sufficient and I'll document the API. Also, please let me know if you think these additions need squashing or not.

phlogistonjohn:
Travis failed due to a network timeout. I will push again soon and hopefully it will resolve the error.

client/cli/go/cmds/utils.go
```diff
@@ -0,0 +1,101 @@
...
+		req = &api.TagsChangeRequest{Tags: setTags, Change: api.SetTags}
+	} else {
+		// to keep the api simple and consistent delete takes a map
+		// however the values in that map are ignored in the
+		// delete case
+		setTags := map[string]string{}
+		for _, k := range s[1:] {
+			setTags[k] = ""
+		}
+		req = &api.TagsChangeRequest{Tags: setTags, Change: api.SetTags}
```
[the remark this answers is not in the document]
phlogistonjohn: Wow. Good catch, will fix.

client/api/python/heketi/heketi.py
```diff
@@ -23,6 +23,12 @@
 import json
 
 
+# Constants related to setting tags
+TAGS_SET = 'set'
```
raghavendra-talur: If you could also comment why these three lines are required here?

phlogistonjohn:
Patches updated.

client/api/python/heketi/heketi.py
phlogistonjohn: I've been annoyed that the pyhton api code does not provide constants. This is the start of that. Instead of "remembering" the right strings for everything a set of constants remind one what options are available and misspelling of a constant will be an early error rather than a later confusion about what didn't work. I hope that explanation makes sense.

raghavendra-talur: Thanks, that explains :)

PR #1130, opened 2018-04-04
cli: support alternative names for setup-openshift-heketi-storage

phlogistonjohn:
This is an only partially serious suggestion triggered by a
comment @jarrpa made a while back.
(see
https://github.com/gluster/gluster-kubernetes/issues/434#issuecomment-366006365
)

Using cobra's aliases feature we no longer have to be embarrassed by
an overly openshift specific command name, but can now refer to it
via much more generic names. In the future we could even swap
the "true" name for one of these fine aliases.

### DIFF
--- a/client/cli/go/cmds/heketi_storage.go
+++ b/client/cli/go/cmds/heketi_storage.go
@@ -282,7 +282,11 @@ func createHeketiCopyJob(volume *api.VolumeInfoResponse) *batch.Job {
 }
 
 var setupHeketiStorageCommand = &cobra.Command{
-	Use:   "setup-openshift-heketi-storage",
+	Use: "setup-openshift-heketi-storage",
+	Aliases: []string{
+		"setup-heketi-db-storage",
+		"setup-kubernetes-heketi-storage",
+	},
 	Short: "Setup OpenShift/Kubernetes persistent storage for Heketi",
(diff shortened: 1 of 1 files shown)


phlogistonjohn:
@nixpanic that's pretty nice. I'll keep that in mind for next time. Thanks!

client/cli/go/cmds/heketi_storage.go
```diff
@@ -282,7 +282,11 @@ func createHeketiCopyJob(volume *api.VolumeInfoResponse) *batch.Job {
 }
 
 var setupHeketiStorageCommand = &cobra.Command{
-	Use:   "setup-openshift-heketi-storage",
+	Use: "setup-openshift-heketi-storage",
+	Aliases: []string{
+		"setup-heketi-self-storage",
```
obnoxxx: I am thinking how about "setup-heketi-db-storage" instead of "setup-heketi-self-storage"? ...

phlogistonjohn:
1. Updated name to  "setup-heketi-db-storage" as suggested
2. Updated format of commit message

jarrpa:
As the inspiration/instigator of this change, chiming in with an LGTM. :)

Also, it looks like the PR needs a rebase.

PR #1133, opened 2018-04-06
glide.lock update heketi rest version

Madhu-1:
updated heketi rest commit version
from 738570ea73f1cc2a0888c7c49fe80991c6050f34
to aa6a652074133e6b0836bb095e847021fcb94b6a

This PR updates the heketi rest package


phlogistonjohn:
Perhaps I don't understand glide as well as I think I do, but I would have expected a change to glide.yaml and then a corresponding edit to the glide.lock file. AFAIK, glide.lock is generated by glide and isn't meant to be "relied" upon.

Madhu-1:
added heketi/rest version in glide.yaml file, some of the packages are not under version control in glide.yaml 
glide update may update other packages  also so manually added heketi/rest version to glide.lock file

phlogistonjohn:
Newest push looks good. I think it would be best if you squashed the two patches together, please.

> added heketi/rest version in glide.yaml file, some of the packages are not under version control in glide.yaml

I'm not certain I understand you correctly, but I think you mean that not all (indirect) dependencies are listed in glide.yaml. If so, then yes I agree we only should be listing our direct dependencies in glide.yaml and the tool will do the rest.

> glide update may update other packages also so manually added heketi/rest version to glide.lock file

Great. In the future we may want to add "and any later version" specifiers in some of these place but heketi/rest is so low activity that it isn't hard to keep up with.

(Also, every time I read the glide project page it starts off with a reminder to migrate to (go) dep but that's a whole 'nother story)

glide.yaml
```diff
@@ -8,6 +8,7 @@ import:
 - package: github.com/gorilla/context
 - package: github.com/gorilla/mux
 - package: github.com/heketi/rest
+  version: aa6a652074133e6b0836bb095e847021fcb94b6a
```
obnoxxx: We should not do this.
If we don't lock the version down, and heketi/rest changes, the next glide update will update the glide.lock file to reflect the new version. Then we should commit just the diff of the glide.lock file.

I would personally prefer not to check in the glide.lock file at all.
The glide.yaml should lock whatever we need to lock down, and
then generate the .lock file it always. IMHO.
But not sure I understand glide well enough either...

phlogistonjohn: Err, OK. It seems that glide's mechanisms for saying "I need at least this version" are kind of lousy.

obnoxxx: I think glide doesn't want to allow you to say this at all.

If you don't say anything, then glide uses the latest.
If you don't specify something, then glide locks down to that exactly.

I think the bug is to check the glide.lock file into the repo.

@phlogistonjohn - i remember you saying that glide is not the up-to-date mechanism. What is a better/newer/hipper/.... mechanism?

phlogistonjohn: > I think glide doesn't want to allow you to say this at all.

I think there are [ways of expressing this](https://glide.readthedocs.io/en/latest/versions/), however it's not the default and I can't tell if it works with commit IDs rather than versions.

> If you don't say anything, then glide uses the latest.

Agreed.

> If you don't specify something, then glide locks down to that exactly.

I think you meant "if you do". If so, I agree (given the caveats in my earlier comment). 

> I think the bug is to check the glide.lock file into the repo.

I think the common practice is to commit the lock file. I'm not entirely sure why.

> @phlogistonjohn - i remember you saying that glide is not the up-to-date mechanism. What is a better/newer/hipper/.... mechanism?

Yeah, its [dep](https://github.com/golang/dep). The glide project page even recommends migration.

obnoxxx: > > If you don't specify something, then glide locks down to that exactly.
> 
> I think you meant "if you do". 

yes :-)

> > I think the bug is to check the glide.lock file into the repo.
> 
> I think the common practice is to commit the lock file. I'm not entirely sure why.

Well, more to investigate....

> > @phlogistonjohn - i remember you saying that glide is not the up-to-date mechanism. What is a better/newer/hipper/.... mechanism?
> 
> Yeah, its [dep](https://github.com/golang/dep). The glide project page even recommends migration.

Ok.... quote: "`dep` is the official experiment, but not yet the official tool." :-D

phlogistonjohn: OK, I understand your points. I'll think more on this but I think we'd eventually want to to do both things. That is, encode the hard versioning in the description (yaml) and soft versioning in the state file (lock) file. This is fine for now though.

obnoxxx: Hmm. This sucks, imho.

I think we want (at least I):
* as few locking down of versions for master (i.e. use dependent projects current master whenever we build)
* lock the dependent projects down for the release branches

Anyways, given you say yaml is hard versioning, and lock is soft, why don't we just commit the update to the .lock file? I see no reason to have the version of heketi/rest locked down in the yaml.

Cheers - Michael

phlogistonjohn: Odd, I could have sworn that's the way the change was the last time I looked at this PR. At this point I'd be happy with whatever minimal change satisfies the ability to build. We can discuss futures as a separate topic.

obnoxxx: Well, it adds a line to the glide.yaml file.
The minimal change would be only the hunk to glide.lock.
Can we get the patch updated to skip the change to glide.yaml or else explain why we need it?

raghavendra-talur: It would be wrong to just update the lock file. As explained in my previous comment, projects should 
1. Have only dependencies without versions listed in glide.yaml and let glide handle glide.lock(this is not checked into repo)  
OR  
2. Update glide.yaml with dependencies and versions both and let glide handle glide.lock(this is not checked into repo)  
OR  
3. Update glide.yaml with dependencies and versions both and check in glide.lock in order to lock the dependencies of dependencies. 

To satisfy @obnoxxx 's request for taking master branches of dependencies for master branch of heketi, we must check in only glide.yaml with only dependencies(without version). When a release is cut, we will have to check in glide.lock and make the release. 

Downside to this is that we won't have reproducible builds for CI but we will have reproducible builds for releases. I accept this as we don't really need reproducible builds for CI.

If agreed, we should still take the patch in its current form. A new PR should remove glide.lock from source control and remove versions from glide.yaml.

obnoxxx: > It would be wrong to just update the lock file. As explained in my previous comment, 

I don't seem to find that comment any more.
So I don't have the **why**, just the **what**... 

> projects should 
> 1. Have only dependencies without versions listed in glide.yaml and let glide handle glide.lock(this is not checked into repo)  
> OR  
> 2. Update glide.yaml with dependencies and versions both and let glide handle glide.lock(this is not checked into repo)  
> OR  
> 3. Update glide.yaml with dependencies and versions both and check in glide.lock in order to lock the dependencies of dependencies. 

So what happens if (as before), we have only the package in the yaml, but not the version? In that case, glide update should still update the glide.lock with latest versions, right? This was what was done before. And we could always commit a new version to glide.lock if we wanted. Or what am I missing. I still don't understand **why** we need the version in glide.yaml if we don't **want** to lock the version down. At least I would want to go with the latest version whenever I run `glide update`.

> To satisfy @obnoxxx 's request for taking master branches of dependencies for master branch of heketi, we must check in only glide.yaml with only dependencies(without version).  When a release is cut, we will have to check in glide.lock and make the release. 
> 
> Downside to this is that we won't have reproducible builds for CI but we will have reproducible builds for releases. I accept this as we don't really need reproducible builds for CI.

I think another downside for master would be that make would usually take a lot longer, or wouldn't it?

> If agreed, we should still take the patch in its current form. A new PR should remove glide.lock from source control and remove versions from glide.yaml.

I have been experimenting with glide on our repo a bit recently. I can't seem to have it run to completion if I removed the glide.lock file. It is running since roughly 8 hours for me by now. Wit the `--debug` switch, I see that a lot of stuff is still going on. Might be some kind of loop. Not sure if cleaning all of `vendor` and the glide cache would help here. But maybe it's also just time to move on to something else.... ("dep"?)

obnoxxx:
ok to test

Issue #413, opened 2016-07-05
Snapshot management


lpabon:
Heketi should have a snapshot management API

lpabon:
https://bugzilla.redhat.com/show_bug.cgi?id=1351450

nixpanic:
The [external-storage snapshot user-guide](https://github.com/kubernetes-incubator/external-storage/blob/master/snapshot/doc/user-guide.md) describes the common functionality:

- create snapshot from a PV
- promote snapshot to PV (sortof PV cloning functionality)
- restore a PV from a snapshot
- delete snapshot

humblec:
Yes, it would be good to implement snapshot functionality in heketi so that it can be plugged with the controller. WIP patch is available for kube side of things @https://github.com/humblec/external-storage/commit/1ee96627f14fb092575d6b0f89e50aba87755944

akram:
@humblec snpashot is an offline operation, how does the k8s volume plugin manage it ?

MartinForReal:
@akram I think the user scenario is clearly stated in external-storage snapshot user-guide.

phlogistonjohn:
I do not think we will pursue this item. Closing issue. Please reopen if this is still relevant.

PR #1134, opened 2018-04-06
Add support for cloning Gluster Volumes

nixpanic:

Volume cloning can be used for creating new volumes from a pre-existing volume. This can be seen as creating a template that can be cloned for consumption by apps that do not need to share the modified data with eachother. The current main consumer of volume cloning is [kubevirt](https://github.com/kubevirt), but others are expected to follow.

The functionality is restricted to Gluster Volumes, and will not work for blockvolumes. Once [gluster-block](https://github.com/gluster/gluster-block) gains a clone functionality, this can be revisited.

[The design](https://github.com/heketi/heketi/blob/master/docs/design/snapshot.md) for cloning volumes has been merged earlier. This PR provides the "Clone a Volume directly" feature, the additional snapshot operations will be sent in a separate PR.

### Does this PR fix issues?

Updates: #413 

### Notes for reviewers

The 1st patch in this PR (client: Add 'volume clone' CLI and API) has been reviewed in #1111 already.


executors/executor.go
```diff
@@ -82,6 +87,66 @@ type VolumeRequest struct {
...
+	Description string
+}
+
+type SnapshotCloneRequest struct {
+	Volume   string
+	Snapshot string
+}
+
+type Snapshot struct {
+	XMLName xml.Name `xml:"snapshot"`
```
phlogistonjohn: I'm a bit puzzled as to why we have xml.Name types in what I feel is the interface to the "outside world." To me it seems like use of xml is an implementation detail. If these are not meant to be part of the public api, perhaps they could be moved to a different file?

apps/glusterfs/operations.go
```diff
@@ -457,6 +457,152 @@ func (vdel *VolumeDeleteOperation) Finalize() error {
...
+	}
+
+	if err := updateCloneBrickPaths(vc.bricks, orig, clone); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (vc *VolumeCloneOperation) Rollback(executor executors.Executor) error {
+	return vc.db.Update(func(tx *bolt.Tx) error {
```
phlogistonjohn: Looks like some stuff is missing from rollback. I think we need to go to the executor and clean up any snapshots/bricks/etc that the operation may have created before the server was terminated. If you don't think you can fit it in thss PR I'd be OK with that, but we should put in a TODO comment at the very least.

apps/glusterfs/brick_entry.go
```diff
@@ -213,6 +214,7 @@ func (b *BrickEntry) Destroy(db wdb.RODB, executor executors.Executor) error {
 	req.Size = b.Info.Size
 	req.TpSize = b.TpSize
 	req.VgId = b.Info.DeviceId
+	req.Path = strings.TrimSuffix(b.Info.Path, "/brick")
```
phlogistonjohn: There are a bunch of brick & path helper functions in pkg/utils/paths.go. Would any of them be appropriate for this?

apps/glusterfs/brick_create.go
```diff
@@ -10,35 +10,22 @@
...
-	CREATOR_CREATE CreateType = iota
-	CREATOR_DESTROY
-)
-
-func createDestroyConcurrently(db wdb.RODB,
-	executor executors.Executor,
-	brick_entries []*BrickEntry,
-	create_type CreateType) error {
-
+func CreateBricks(db wdb.RODB, executor executors.Executor, brick_entries []*BrickEntry) error {
```
phlogistonjohn: Thank you for attacking this crazy function and working toward a sane interface! :-)

executors/cmdexec/brick.go
```diff
@@ -150,6 +152,9 @@ func (s *CmdExecutor) BrickDestroy(host string,
 	_, err = s.RemoteExecutor.RemoteCommandExecute(host, commands, 5)
 	if err != nil {
 		logger.Err(err)
+	} else {
+		// no space freed when tp sticks around?
```
phlogistonjohn: Why the question mark?

executors/cmdexec/brick_test.go
```diff
@@ -300,6 +300,7 @@ func TestSshExecBrickDestroy(t *testing.T) {
 	}
 
 	// Create Brick
-	err = s.BrickDestroy("myhost", b)
-	tests.Assert(t, err == nil, err)
+	_, err = s.BrickDestroy("myhost", b)
+	// TODO: this tests fail because BrickDestroy checks the output of 'mount'.
+	//tests.Assert(t, err == nil, err)
```
phlogistonjohn: Ideally, we'd update the test to match the reality of the code's new behavior.

apps/glusterfs/app_volume.go
```diff
@@ -328,3 +328,49 @@ func (a *App) VolumeExpand(w http.ResponseWriter, r *http.Request) {
...
+			http.StatusBadRequest)
+		logger.LogError("validation failed: " + err.Error())
+		return
+	}
+
+	var volume *VolumeEntry
+	err = a.db.View(func(tx *bolt.Tx) error {
+		var err error
+		volume, err = NewVolumeEntryFromId(tx, vol_id)
+		if err == ErrNotFound {
```
phlogistonjohn: I think we need a check here for volume visibility. See VolumeInfo at around line 195 in app_volume.go.
This will protect us from accidentally starting a clone operation on a pending volume and will protect us from stuff like a clone coming in on a volume that's being deleted.

executors/executor.go
nixpanic: Maybe that would have been nicer, but there are other structs in the same file that already do it like this.

executors/cmdexec/brick.go
nixpanic: The question mark can be removed.

apps/glusterfs/app_volume.go
nixpanic: Ah, got it!

executors/executor.go
phlogistonjohn: OK, I missed that there was a precedent for this already in that file. Cleanups can always come later.

apps/glusterfs/operations.go
nixpanic: For now a TODO comment is all I can do. Once the more advanced Snatshot operations are available, the VolumeClone operation can move from the CmdExecutor to `operations.go` where each step can be rolled back.

executors/cmdexec/brick_test.go
nixpanic: I'm looking into extending the commands in `FakeConnectAndExec` above, that might be sufficient to have the test pass again.

nixpanic:
Reminder for @heketi/maintainers that the comments have been addressed and further reviewing can be done

client/cli/go/cmds/volume.go
```diff
@@ -383,3 +388,46 @@ var volumeListCommand = &cobra.Command{
...
+	Example: "  $ heketi-cli volume clone 886a86a868711bef83001",
+	RunE: func(cmd *cobra.Command, args []string) error {
+		//ensure proper number of args
+		s := cmd.Flags().Args()
+		if len(s) < 1 {
+			return errors.New("Volume id missing")
+		}
+
+		// Set volume id
+		volumeId := cmd.Flags().Arg(0)
```
raghavendra-talur: I would prefer not to have positional args. Refer to any other sub command of volume.

executors/cmdexec/snapshot.go
```diff
@@ -0,0 +1,166 @@
...
+	if err != nil {
+		return fmt.Errorf("Unable to activate snapshot %v: %v", snapshot, err)
+	}
+
+	var snapActivate CliOutput
+	err = xml.Unmarshal([]byte(output[0]), &snapActivate)
+	if err != nil {
+		return fmt.Errorf("Unable to parse output from activate snapshot %v: %v", snapshot, err)
+	}
+	logger.Debug("%+v\n", snapActivate)
```
raghavendra-talur: not checking for opRet here...Command could have been executed but return value could indicate unsuccessful execution.

executors/cmdexec/snapshot.go
```diff
@@ -0,0 +1,166 @@
...
+	if err != nil {
+		return fmt.Errorf("Unable to deactivate snapshot %v: %v", snapshot, err)
+	}
+
+	var snapDeactivate CliOutput
+	err = xml.Unmarshal([]byte(output[0]), &snapDeactivate)
+	if err != nil {
+		return fmt.Errorf("Unable to parse output from deactivate snapshot %v: %v", snapshot, err)
+	}
+	logger.Debug("%+v\n", snapDeactivate)
```
raghavendra-talur: not checking for opRet here...Command could have been executed but return value could indicate unsuccessful execution.

executors/cmdexec/snapshot.go
```diff
@@ -0,0 +1,166 @@
...
+	if err != nil {
+		return fmt.Errorf("Unable to delete snapshot %v: %v", snapshot, err)
+	}
+
+	var snapDelete CliOutput
+	err = xml.Unmarshal([]byte(output[0]), &snapDelete)
+	if err != nil {
+		return fmt.Errorf("Unable to parse output from delete snapshot %v: %v", snapshot, err)
+	}
+	logger.Debug("%+v\n", snapDelete)
```
raghavendra-talur: not checking for opRet here...Command could have been executed but return value could indicate unsuccessful execution.

executors/executor.go
raghavendra-talur: Yes, blame shows it was me who introduced it and I don't seem to recall why I named it so. I don't like it either. We can have cleanup patch for this.

apps/glusterfs/volume_entry.go
```diff
@@ -154,6 +155,34 @@ func NewVolumeEntryFromId(tx *bolt.Tx, id string) (*VolumeEntry, error) {
...
+	entry := NewVolumeEntry()
+
+	entry.Info.Id = utils.GenUUID()
+	if name == "" {
+		entry.Info.Name = "vol_" + entry.Info.Id
+	} else {
+		entry.Info.Name = name
+	}
+
+	entry.GlusterVolumeOptions = v.GlusterVolumeOptions
```
raghavendra-talur: Does a cloned volume get the same volume options too? If yes then the code is fine. Can we add a comment for the same? If it does not then let us make this empty.

apps/glusterfs/volume_entry.go
```diff
@@ -154,6 +155,34 @@ func NewVolumeEntryFromId(tx *bolt.Tx, id string) (*VolumeEntry, error) {
...
+	entry.Info.Cluster = v.Info.Cluster
+	entry.Info.Durability = v.Info.Durability
+	entry.Info.Durability.Type = v.Info.Durability.Type
+	entry.Info.Gid = v.Info.Gid
```
raghavendra-talur: Same here.

apps/glusterfs/volume_entry.go
```diff
@@ -154,6 +155,34 @@ func NewVolumeEntryFromId(tx *bolt.Tx, id string) (*VolumeEntry, error) {
...
+	entry.Info.Mount = v.Info.Mount
+	entry.Info.Size = v.Info.Size
+	entry.Info.Snapshot = v.Info.Snapshot
+	copy(entry.Info.Mount.GlusterFS.Hosts, v.Info.Mount.GlusterFS.Hosts)
+	entry.Info.Mount.GlusterFS.MountPoint = v.Info.Mount.GlusterFS.Hosts[0] + ":" + entry.Info.Name
+	entry.Info.Mount.GlusterFS.Options = v.Info.Mount.GlusterFS.Options
+	entry.Info.BlockInfo.FreeSize = v.Info.BlockInfo.FreeSize
+	copy(entry.Info.BlockInfo.BlockVolumes, v.Info.BlockInfo.BlockVolumes)
```
raghavendra-talur: I am very unsure about this. Should we even allow for clone of blockhosting volumes? What happens to the IQN of the blocks that reside on them. @pkalever  How unique is the IQN? If blockhosting volume is cloned along with the metadata and the block files on it, would it be fine from gluster-block perspective?

apps/glusterfs/operations.go
phlogistonjohn: That's fine. Please add the comment when you have a moment. Thanks.

client/cli/go/cmds/volume.go
ansiwen: @raghavendra-talur other sub-commands like `info` and `delete` also use positional args. Only the sub-commands that need more than a volume id seem to use named arguments. (This is apparently copy and pasted from `info`.)

obnoxxx:
@nixpanic After latest merges, this needs (nontrivial) rebase. Sorry...
@phlogistonjohn Can you check, whether your requests have been addressed.

obnoxxx:
@nixpanic could you rebase?
@phlogistonjohn have your requests been addressed?

phlogistonjohn:
> @phlogistonjohn have your requests been addressed?

The comment I was expecting to see seems to be added to the function (just not where I expected it so I overlooked it earlier). I would like to hear responses to the various comments @raghavendra-talur made though.

apps/glusterfs/volume_entry.go
obnoxxx: Yes, I think we need to keep the GID (if that's what you mean). At least if we don't, we need to change the volume after clone to work with the new GID, right?

apps/glusterfs/volume_entry.go
obnoxxx: Yes, I think that clone should copy the options. Any reason not to?

client/cli/go/cmds/volume.go
obnoxxx: We discussed this a while ago, and I explicitly requested @nixpanic to do it like this (I think I even did the change to the CLI myself), because just as @ansiwen said, this is consistent with the other commands. And the main argument is that a CLI operation should (imho) always be a positional argument. Like the volume ID to clone, in this case.

executors/cmdexec/snapshot.go
obnoxxx: Agreed, this should be fixed.

executors/executor.go
obnoxxx: Yeah, let's cleanup later.

obnoxxx:
@nixpanic , as @phlogistonjohn pointed out, there are a couple of un-addressed requests.

executors/cmdexec/snapshot.go
ansiwen: done

executors/cmdexec/snapshot.go
ansiwen: done

apps/glusterfs/volume_entry.go
obnoxxx: Caveat: don't confuse block volumes and block-hosting volumes (glusterfs volumes that host block volumes)

phlogistonjohn: > Caveat: don't confuse block volumes and block-hosting volumes (glusterfs volumes that host block volumes)

Good point, but can you currently do anything useful with block hosting volumes that have block volumes in them with a clone? (And cloning an empty volume is pretty pointless) My first impression is no, because none of th iscsi stuff will be configured so the clone would just exist  but not be usable. 

> Maybe a HTTP code like EINVAL exists?

You're probably stuck with 400 as I don't see any of the other codes fitting this well. But please emit error text that is specific. Something like "cloning block hosting volumes is not supported" for example

tests/functional/TestSmokeTest/tests/volume_clone_test.go
```diff
@@ -0,0 +1,89 @@
...
+	clonedVol, err := heketi.VolumeClone(vol.Id, cloneReq)
+	tests.Assert(t, err == nil, "expected err == nil, got:", err)
+	tests.Assert(t, vol.Id != clonedVol.Id, "expected vol.Id != clonedVol.Id, got:", err)
+
+	cloneReq = &api.VolumeCloneRequest{
+		Name: "my_cloned_volume",
+	}
+	namedClonedVol, err := heketi.VolumeClone(vol.Id, cloneReq)
+	tests.Assert(t, err == nil, "expected err == nil, got:", err)
+	tests.Assert(t, namedClonedVol.Name == cloneReq.Name, "expected namedClonedVol.Name == cloneReq.Name, got:", err)
```
ansiwen: Same here: `err`

tests/functional/TestSmokeTest/tests/volume_clone_test.go
```diff
@@ -0,0 +1,89 @@
...
+	volReq.Durability.Type = api.DurabilityReplicate
+	volReq.Durability.Replicate.Replica = 3
+
+	vol, err := heketi.VolumeCreate(volReq)
+	tests.Assert(t, err == nil, "expected err == nil, got:", err)
+
+	cloneReq := &api.VolumeCloneRequest{}
+	clonedVol, err := heketi.VolumeClone(vol.Id, cloneReq)
+	tests.Assert(t, err == nil, "expected err == nil, got:", err)
+	tests.Assert(t, vol.Id != clonedVol.Id, "expected vol.Id != clonedVol.Id, got:", err)
```
ansiwen: Same here: `err`

tests/functional/TestSmokeTest/tests/volume_clone_test.go
```diff
@@ -0,0 +1,89 @@
...
```
ansiwen: `err` is not what should be returned in this assert message.

apps/glusterfs/volume_entry.go
ansiwen: @nixpanic, I implemented a check for Info.Block in prepareVolumeClone() in an additional commit. Please review.

obnoxxx:
what's the status here?
It's a little unclear whether all comments have now been addressed... :-)

obnoxxx:
@phlogistonjohn @ansiwen @nixpanic

obnoxxx:
I think from reading through the comments and updates carefully, all requests have been addressed. We need this to pass tests though.

ansiwen:
@obnoxxx:
> what's the status here?
It's a little unclear whether all comments have now been addressed... :-)

I think that's because we don't use a proper reviewing tool. :smiling_imp:

> I think from reading through the comments and updates carefully, all requests have been addressed. We need this to pass tests though.

I added a "unit"-test for not-allowing-block-volume-cloning and I think we are done now. The failing test was not related to this change I think (something with arbiter). Let's see what the next test run will show.

obnoxxx:
> @obnoxxx:
> 
> >  what's the status here?
> >  It's a little unclear whether all comments have now been addressed... :-)
> 
> I think that's because we don't use a proper reviewing tool. smiling_imp

true :-)

> > I think from reading through the comments and updates carefully, all requests have been addressed. We need this to pass tests though.
> 
> I added a "unit"-test for not-allowing-block-volume-cloning and I think we are done now. The failing test was not related to this change I think (something with arbiter). Let's see what the next test run will show.

I respun one travis build, but it failed again at TestDeviceRemoveSizeAccounting/standard. Need to investigate

phlogistonjohn:
I just pushed a change (msg: "apps: correctly handle space reclaimed on brick destroy in replace path") to fix this. This actually caught a real regression in the behavior around device remove!

PR #1138, opened 2018-04-11
Provide options to change arbiter brick size

raghavendra-talur:
This PR provides options to change arbiter brick size calculation in heketi. This is needed because as explained below the default in heketi might not work for all use cases. 

Calculation of arbiter brick size is dependent on

* size required for storing metadata for a each file
* maximum number of files that user may need to store on the volume.

We fix the metadata size for each file to 1KiB based on 
* https://gist.github.com/pfactum/e8265ca07f7b19f30bb3
* http://gluster.readthedocs.io/en/latest/Administrator%20Guide/arbiter-volumes-and-quorum/

Hence the only variable that remains is maximum number of files that user may need to store on the volume. This is a little difficult to estimate but can be derived from average file size and volume size. As we already know the volume size, it is easier to just ask for average file size from user.

If we assume average file size is 64KiB then there can be maximum 16384 files in a 1GiB volume and arbiter brick would need 16MiB. This seems like a good default; if the arbiter node has only one 512 GiB SSD, it can still take about 32K bricks. 

At this point, there are three levels where average file size is defined
a. each volume create can come with user.heketi.average-file-size value and it applies only to that volume
b. admin can provide a different average file size in config file and it applies to all volumes which are further created and don't have explicit value for average file size.
c. heketi internally has set average file size as 64 KiB.

### DIFF
--- a/apps/glusterfs/placer_arbiter.go
+++ b/apps/glusterfs/placer_arbiter.go
@@ -17,9 +17,6 @@ import (
 
 var (
 	tryPlaceAgain error = fmt.Errorf("Placement failed. Try again.")
-
-	// average size of files on a volume
-	AverageFileSize uint64 = 64 * KB
 )
 
 const (
@@ -57,7 +54,7 @@ func newArbiterOpts(opts PlacementOpts) *arbiterOpts {
 func (aopts *arbiterOpts) discount(index int) (err error) {
 	if index == arbiter_index {
 		aopts.brickSize, err = discountBrickSize(
-			aopts.brickSize, AverageFileSize)
+			aopts.brickSize, aopts.o.AverageFileSize())
 	}
 	return
 }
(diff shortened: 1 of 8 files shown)


apps/glusterfs/placer_arbiter.go
```diff
@@ -54,10 +51,10 @@ func newArbiterOpts(opts PlacementOpts) *arbiterOpts {
 	}
 }
 
-func (aopts *arbiterOpts) discount(index int) (err error) {
+func (aopts *arbiterOpts) discount(index int, averageFileSize uint64) (err error) {
```
phlogistonjohn: You shouldn't need to make it an extra argument here. arbiterOpts refers back to the placement opts which you already enhanced with the AverageFileSize method. So, something like:
```golang
func (aopts *arbiterOpts) discount(index int) (err error) {
 	if index == arbiter_index {
 		aopts.brickSize, err = discountBrickSize(
			aopts.brickSize, aopts.o.AverageFileSize())
 	}
 	return
}
```
should work.

raghavendra-talur: That is nice, I missed it. Updated.

PR #1141, opened 2018-04-11
Improve create rollback behavior when cleanup fails

phlogistonjohn:

When the underlying gluster system is able to create a volume but not start it heketi can get into a weird state. This seems to have been made worse by the recent volume create retry code being able to "spam" many volume create commands in a shorter time period.
This change aims to make rollback more accurately reflect the state of the system and fail in the face of a condition that it can not clean up.

### DIFF
--- a/apps/glusterfs/volume_entry.go
+++ b/apps/glusterfs/volume_entry.go
@@ -777,7 +777,15 @@ func (v *VolumeEntry) runOnHost(db wdb.RODB,
 		return err
 	}
 
+	nodeUp := currentNodeHealthStatus()
 	for nodeId, host := range hosts {
+		if up, found := nodeUp[nodeId]; found && !up {
+			// if the node is in the cache and we know it was not
+			// recently healthy, skip it
+			logger.Debug("skipping node. %v (%v) is presumed unhealthy",
+				nodeId, host)
+			continue
+		}
 		logger.Debug("running function on node %v (%v)", nodeId, host)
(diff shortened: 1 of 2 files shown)


phlogistonjohn:
Need to finish this one up but I thought I'd throw it out there for early feedback.

phlogistonjohn:
Ready for review now.

apps/glusterfs/volume_entry.go
```diff
@@ -726,3 +726,43 @@ func eligibleClusters(db wdb.RODB, req ClusterReq,
...
+	})
+	if err != nil {
+		logger.LogError("runOnHost failed to get hosts: %v", err)
+		return err
+	}
+
+	for nodeId, host := range hosts {
+		logger.Debug("running function on node %v (%v)", nodeId, host)
+		tryNext, err := cb(host)
+		if !tryNext {
```
obnoxxx: I would have found it a little more natural to do the logic roughly like this:
```
func (v *VolumeEntry) runOnHost(db wdb.RODB,
	cb func(host string) (error), bool tryNexy) error {
    ...
    for nodeId, host := range hosts {
        err := cb(host)
        if err == nil {
            return nil
        }
        if !tryNext {
            return err
        }
    }
    return fmt.Errorf(...)
}
```
Since I can't imagine the callback adapting the value of tryNext for different error cases. But maybe I'm just not seeing it. Won't block the merge for this. Functionally, it's perfectly fine as it is.

PR #1146, opened 2018-04-19
cleanup some unused code post arbiter features merge

phlogistonjohn:

In the Heketi code base the changes that lead to the arbiter support (brick placer) defined a new way of getting devices and nodes out of the db and into the placers (the DeviceSource). However, the old approach of loading the simple allocator directly from the db remained in the code base and some of the unit tests. This change removes that approach leaving us with one way to load the placer(s) with devices.

### Notes for the reviewer

This change is a prerequisite for additional refactoring that can allow us to split the placer logic out of the glusterfs app and into its own package.


phlogistonjohn:
> The cleanup of the allocator_simple.go code removes all of allocator_simple_test.go. This shows that the few remaining functions have not (directly) been tested in unit tests yet. Should we add tests for them? ...

Yes, but I want to do that after the move to package PR. Once in a package, using the coverage tools it will be clearer what is being tested as a side effect of other tests and what tests are directly exercising this code.

PR #1149, opened 2018-04-23
create a new package for placers

phlogistonjohn:

As part of the effort to add arbiter support to Heketi we defined a new higher level abstraction for determining where bricks would be placed on nodes/devices. This logic was all part of the existing "apps/glusterfs" package. This series creates a new package "apps/glusterfs/placer" after additional cleanups and refactoring to prepare for the split.

By splitting this code into a new package we clearly delineate what makes up the placer logic and can allow us to (unit) test the placer algorithms independently of the glusterfs package. This also could, in theory, make the package useful to other code besides the glusterfs package.

### Notes for the reviewer

This is a fairly large-ish series with a number of individual changes that are larger than I find ideal. However, in order to make sure the main code  base compiles while disentangling things it lead to some bigger changes. The conversion of <Brick/Node/Volume>Entry types to the placer interfaces is the biggest of these and is even split into the main code and unit test fixes. This does mean that the unit tests fail to compile between the two patches. I can squash them if the reviewers wish, but it seems reasonable to keep them apart for now so that it would be easier to review.

This series depends on PR #1146.


obnoxxx:
@phlogistonjohn conflicts...

phlogistonjohn:
Rebased on master.

obnoxxx:
@phlogistonjohn could you rebase this, please?

phlogistonjohn:
I'm going to suggest we hold off on this until after we release heketi 7. Does that sound OK?

obnoxxx:
@phlogistonjohn I was kind of trying to get to it before v7, but if you prefer to hold it off, it's fine by me. Maybe add the do-not-merge label for the time being?

obnoxxx:
@phlogistonjohn v7 is out since a while now ... do we want to pick it up again? :-)

phlogistonjohn:
too much effort to rebase for too little value. closing.

PR #1150, opened 2018-04-24
tests: allow each test.go file to have heketi config

raghavendra-talur:
With so many new features being added to heketi that have config
options to be specified in heketi.json, it is useful to start heketi
server with a different config for each test.

To keep things simple, we expect the config file to be named same as the
test file but with "json" as extension instead of "go". In case where
such a file isn't found, we use the default config file heketi.json in
config dir of test dir.

### Notes for the reviewer
Going forward, this will help in consolidating all the test folders into one thereby creating only one set of test VMs for all tests.

### DIFF
--- a/tests/functional/lib.sh
+++ b/tests/functional/lib.sh
@@ -19,14 +19,20 @@ _sudo() {
 
 HEKETI_PID=
 start_heketi() {
+    local config_filename=$1
     ( cd "$HEKETI_SERVER_BUILD_DIR" && make && cp heketi "$HEKETI_SERVER" )
     if [ $? -ne 0 ] ; then
         fail "Unable to build Heketi"
     fi
 
+    if [ ! -f config/"${config_filename}" ]
+    then
+        config_filename="heketi.json"
+    fi
+
     # Start server
     rm -f heketi.db > /dev/null 2>&1
-    $HEKETI_SERVER --config=config/heketi.json &
+    $HEKETI_SERVER --config=config/${config_filename} &
     HEKETI_PID=$!
     sleep 2
 }
@@ -45,8 +51,24 @@ teardown_vagrant() {
 
 run_go_tests() {
     cd tests || fail "Unable to 'cd tests'."
-    go test -timeout=1h -tags functional -v
-    gotest_result=$?
+    for t in $(find . -name '*_test.go' | LC_COLLATE=C sort)
+    do
+        config_file=${t%.go}.json
+        cd ..
+        start_heketi "$config_file"
+        cd tests || fail "Unable to 'cd tests'."
+        testfuncs=$(grep "func Test.*(t\\ \\*testing.T)[\\ ]*{" ./* | awk '{print $2}' | cut -d"(" -f1)
+        for testfunc in "${testfuncs[@]}"
+        do
+            go test -timeout=1h -tags functional -run "$testfunc"
+            if [ $? -ne 0 ]
+            then
+                gotest_result=$?
+            fi
+        done
+        testfuncs=""
+        kill $HEKETI_PID
+    done
     cd ..
 }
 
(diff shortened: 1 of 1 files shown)


tests/functional/lib.sh
```diff
@@ -19,14 +19,20 @@ _sudo() {
...
 
+    if [ ! -f config/"${config_filename}" ]
+    then
+        config_filename="heketi.json"
+    fi
+
     # Start server
     rm -f heketi.db > /dev/null 2>&1
-    $HEKETI_SERVER --config=config/heketi.json &
+    $HEKETI_SERVER --config=config/${config_filename} &
```
phlogistonjohn: All the changes to this function look fine to me. It makes the function more general.

tests/functional/lib.sh
```diff
@@ -45,8 +51,24 @@ teardown_vagrant() {
 
 run_go_tests() {
     cd tests || fail "Unable to 'cd tests'."
-    go test -timeout=1h -tags functional -v
-    gotest_result=$?
+    for t in $(find . -name '*_test.go' | LC_COLLATE=C sort)
```
phlogistonjohn: So, I'll admit to being a bit skeptical about the value that this change adds to the code. It seems like we're going to drastically slow down all of the functional tests (which are already quite slow). I am also having a hard time seeing what the benefit of the change is.

I think it would be valuable to have some examples of how you imagine this would be working in the future.

Without seeing your response one thought that has crossed my mind is to: create new sibling packages where we currently only have "tests". We might have "foo_tests" and "bar_tests". You could then vary the the config on the package level rather than splitting things at the go file level (which seemed a bit unnatural to me).

raghavendra-talur:
```
So, I'll admit to being a bit skeptical about the value that this change adds to the code. It seems like we're going to drastically slow down all of the functional tests (which are already quite slow). I am also having a hard time seeing what the benefit of the change is.

I think it would be valuable to have some examples of how you imagine this would be working in the future.
```
I am not sure why you say this will slow down the functional tests. The only thing that we change is number of times that heketi is started and stopped. Comparing to how long the tests run it is negligible.

The main reason for the patch is to have multiple config files for different test files. This is also going to help with collating all 4 or 5 vagrant environments into one.


```
Without seeing your response one thought that has crossed my mind is to: create new sibling packages where we currently only have "tests". We might have "foo_tests" and "bar_tests". You could then vary the the config on the package level rather than splitting things at the go file level (which seemed a bit unnatural to me).
```
What you suggest is actually the right way of doing things. I aimed for minimal changes and sent this PR. However, I haven't thought of all the changes required to get there, it might mean quite a lot of changes.

phlogistonjohn: > I am not sure why you say this will slow down the functional tests. The only thing that we change is number of times that heketi is started and stopped. Comparing to how long the tests run it is negligible.

I think this will also compile the tests every time the go test command runs. But, you're right that the additional time it takes will largely be swamped by actual tests themselves.  That said in an ideal I'd prefer to reduce the time to run.

> The main reason for the patch is to have multiple config files for different test files. This is also going to help with collating all 4 or 5 vagrant environments into one.

Got it. I was hoping you were going in that direction but I wasn't sure. :-)

> What you suggest is actually the right way of doing things. I aimed for minimal changes and sent this PR. However, I haven't thought of all the changes required to get there, it might mean quite a lot of changes.

I think running things at the granularity of the package is probably the right way to go. We can structure it either tests/foo, tests/bar or more flat like tests_foo or tests_bar too. 

One other thought I had is that if we really wanted to manage the heketi server at the granularity of the go test case we ought to start and stop the server inside the go code. I slightly prefer the first approach (packages) but I could also live with the latter.

raghavendra-talur:
This PR is outdated now with the Server Control feature in the functional tests. Closing.

PR #1156, opened 2018-04-25
apps: treat an empty string log level name as a no-op

phlogistonjohn:
Some heketi config files lack a specific log level and thus
the configuration value is defaulted to the empty string.
If this is passed to the glusterfs app SetLogLevel it produces
an ugly message on start up. Simply treat the empty string
as a no-op.


phlogistonjohn:
Unfortunately, the logging subsystem that heketi uses doesn't lend itself well to being tested, ATM. I'll think about it though.

PR #1159, opened 2018-04-26
Add client side tls configuration

phlogistonjohn:

With pr #1108 now merged Heketi server can now be enabled for TLS/HTTPS. However, if the client APIs need to access an HTTPS secured heketi they may run into issues if the server is using self-signed or non well known CAs. To ease this situation the following changes add the ability to enable or disable verification OR provide cert files to the python client, the go client, and the heketi cli.

### Notes for the reviewer

Please feel free to bikeshed the cli option names and the user facing variable names if they don't seem clear.

### DIFF
--- a/client/api/go-client/client.go
+++ b/client/api/go-client/client.go
@@ -14,7 +14,11 @@ package client
 
 import (
 	"crypto/sha256"
+	"crypto/tls"
+	"crypto/x509"
 	"encoding/hex"
+	"fmt"
+	"io/ioutil"
 	"net/http"
 	"time"
 
@@ -26,12 +30,22 @@ const (
 	MAX_CONCURRENT_REQUESTS = 32
 )
 
+type ClientTLSOptions struct {
+	// directly borrow the field names from crypto/tls
+	InsecureSkipVerify bool
+	// one or more cert file paths (best for self-signed certs)
+	VerifyCerts []string
+}
+
 // Client object
 type Client struct {
 	host     string
 	key      string
 	user     string
 	throttle chan bool
+
+	// configuration for TLS support
+	tlsClientConfig *tls.Config
 }
 
 // Creates a new client to access a Heketi server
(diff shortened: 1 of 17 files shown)


phlogistonjohn:
I need to add tests for the go lang changes, so I set do not merge. Please feel free to provide early feedback though.

CC @obnoxxx @rsevilla87

client/api/go-client/client.go
```diff
@@ -48,11 +63,49 @@ func NewClient(host, user, key string) *Client {
 	return c
 }
 
+func NewClientTLS(host, user, key string, tlsOpts *ClientTLSOptions) (*Client, error) {
+	c := NewClient(host, user, key)
+	if err := c.SetTLSOptions(tlsOpts); err != nil {
```
Madhu-1: do we need to check host schema (whether its HTTP or HTTPS before creating HTTPS client)

phlogistonjohn: I don't think so. Internally, the Go http client will use the protocol indicated by the url, but we don't need to key to conditionalize our logic. If HTTPS is configured but not used by the request these options will just be ignored by the http client code.

client/cli/go/cmds/root.go
```diff
@@ -64,6 +66,13 @@ func init() {
 		"\n\tPrint response as JSON")
 	RootCmd.Flags().BoolVarP(&version, "version", "v", false,
 		"\n\tPrint version")
+	RootCmd.PersistentFlags().BoolVarP(&options.InsecureTLS,
```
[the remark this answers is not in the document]
phlogistonjohn: Good thought, but I think it makes sense to continue to use the --server option as the source of truth for that. It seems a bit too complicated to switch protocols based on the presence of the tls config options, especially when we can get the server URL either from the cli option or from an env variable.

phlogistonjohn:
retest this please

phlogistonjohn:
rebased

obnoxxx:
not sure why we have a conflict now.

phlogistonjohn:
> not sure why we have a conflict now.

I made a change in a different PR to improve the reliability of the tests that conflicted with this series. Rebased.

phlogistonjohn:
Yum failure in centos ci

retest this please

phlogistonjohn:
Ansible/yum failures in centos ci. Not related to change.

retest this please

PR #1160, opened 2018-04-27
fixes and improvements for functional test runner

phlogistonjohn:

Functional test runner had a couple of issues, the biggest being
not correctly exiting non-zero on a test failure. This change
fixes that along with some other general cleanups.

### DIFF
--- a/apps/glusterfs/volume_entry_test.go
+++ b/apps/glusterfs/volume_entry_test.go
@@ -2430,3 +2430,45 @@ func TestVolumeCreateArbiterSizingCustom(t *testing.T) {
 		return nil
 	})
 }
+
+func TestVolumeCreateBoundarySizing(t *testing.T) {
+	tmpfile := tests.Tempfile()
+	defer os.Remove(tmpfile)
+
+	app := NewTestApp(tmpfile)
+	defer app.Close()
+
+	err := setupSampleDbWithTopology(app,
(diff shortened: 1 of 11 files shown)


phlogistonjohn:
I have reason to believe that the TestSmokeTest suite is still flaky. It's passing ATM but was failing regularly on centos ci before I pushed my extra debugging changes. I'm going to continue looking into this.

phlogistonjohn:
This PR has become somewhat of a hodgepodge of changes but they all ultimately are trying to make the functional tests correctly report errors, succeed more reliably, and easier to debug when it does fail. Please let me know if think this needs more cleanup or should be split into more PRs. I think the state is OK to review now.

phlogistonjohn:
I'd like to ask that this PR be reviewed with priority.
I had to do PR #1171 because the functional tests failed to report that the smoke test suite failed. If we had been alerted to that problem by the functional tests I could have fixed the previous PR more quickly.

obnoxxx:
retest this please

phlogistonjohn:
This is failing now because master lacks the fix in PR #1171 
These PRs are now mutually dependent on each other for full successful runs of the functional suite, unfortunately.

obnoxxx:
I wanted to let the test suite pass. Interestingly, #1171 passed without this. Waiting for the result here now #1171 is merged.

PR #1161, opened 2018-04-30
HEKETI_DB_PATH environment variable overrides the heketi.json setting for the daemon

ateleshev:
It is unfortunate that if we choose to use a different place for the DB,
and if we do the backup-to-kube-secret thing, that we then need to
specify it in two places now. This solution fixs this problem.

Related to: https://github.com/heketi/heketi/pull/1155

### DIFF
--- a/apps/glusterfs/app_test.go
+++ b/apps/glusterfs/app_test.go
@@ -75,12 +75,14 @@ func TestAppAdvsettings(t *testing.T) {
 	defer os.Remove(dbfile)
 	os.Setenv("HEKETI_EXECUTOR", "mock")
 	defer os.Unsetenv("HEKETI_EXECUTOR")
+	os.Setenv("HEKETI_DB_PATH", dbfile)
+	defer os.Unsetenv("HEKETI_DB_PATH")
 
 	data := []byte(`{
 		"glusterfs" : {
 			"executor" : "crazyexec",
 			"allocator" : "simple",
-			"db" : "` + dbfile + `",
+			"db" : "/path/to/nonexistent/heketi.db",
 			"brick_max_size_gb" : 1024,
 			"brick_min_size_gb" : 4,
 			"max_bricks_per_volume" : 33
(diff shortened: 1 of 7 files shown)


ateleshev:
@obnoxxx - this is the next part of https://github.com/heketi/heketi/pull/1155

apps/glusterfs/app_test.go
```diff
@@ -228,12 +230,14 @@ func TestAppBlockSettings(t *testing.T) {
...
 	defer os.Unsetenv("HEKETI_EXECUTOR")
+	os.Setenv("HEKETI_DB_PATH", dbfile)
+	defer os.Unsetenv("HEKETI_DB_PATH")
 
 	data := []byte(`{
 		"glusterfs" : {
 			"executor" : "crazyexec",
 			"allocator" : "simple",
-			"db" : "` + dbfile + `",
+			"db" : "/path/to/nonexistent/heketi.db",
```
obnoxxx: Same here: Why change the db settings at all if you're not testing them explicitly?

apps/glusterfs/app_test.go
```diff
@@ -75,12 +75,14 @@ func TestAppAdvsettings(t *testing.T) {
...
```
obnoxxx: This is potentially problematic: Despite its name, his path may exist by coincidence...

Also, since you don't seem to specifically test the db setting here, why change this at all?

apps/glusterfs/app_test.go
ateleshev: Test case has added, please see below.

apps/glusterfs/app_test.go
ateleshev: Test case has added, please see below.

ateleshev:
@obnoxxx - It has been done like as executor environment variable. In configuration has specified nonexistent executor - "crazyexec" and this parameter changed to "mock" via passing new value in environment variable. I don't think it is a good idea to add a specific test case only for the HEKETI_DB_PATH variable. Maybe should we add a specific test case for all environment variables, and moves into this test case all variables: HEKETI_EXECUTOR, HEKETI_DB_PATH, HEKETI_IGNORE_STALE_OPERATIONS and other what will be add in future?

obnoxxx:
ok to test

obnoxxx:
retest this please

obnoxxx:
retest this please

PR #1168, opened 2018-05-07
cli: fix copy'n'paste error in usage of heketi-cli device info

obnoxxx:

This PR fixes a copy-and-paste error in the `heketi-cli device info --help` message, where `node` would be printed instead of `device`.


phlogistonjohn:
FWIW, travis failure does not appear related to this change. Will restart it again later.

PR #1172, opened 2018-05-08
create arbiter bricks sets in reverse order

phlogistonjohn:

Creating the arbiter bricks in reverse order (starting with the 3rd
brick in the brick set and working back to the 1st) means we create the arbiter brick first
and this prevents the data bricks from "blocking" the arbiter brick
being placed on a particular node. This makes placing the
arbiter bricks more reliable.

### Does this PR fix issues?

See [rhbz #1572561](https://bugzilla.redhat.com/show_bug.cgi?id=1572561)

### Notes for the reviewer

This PR relies on the changes in PR #1171 so that tests can pass. Please review that PR first.

### DIFF
--- a/apps/glusterfs/brick_allocate.go
+++ b/apps/glusterfs/brick_allocate.go
@@ -24,10 +24,16 @@ type BrickSet struct {
 	Bricks  []*BrickEntry
 }
 
+// NewBrickSet creates an empty brick set tracking object with
+// the expected final size of s.
 func NewBrickSet(s int) *BrickSet {
 	return &BrickSet{SetSize: s, Bricks: []*BrickEntry{}}
 }
 
+// NewSparseBrickSet creates a brick set tracking object with
+// the expected final size of s and the brick tracking slice
+// pre-sized to s. This means that the array can be directly
+// indexed but may contain null brick entries.
 func NewSparseBrickSet(s int) *BrickSet {
 	return &BrickSet{
 		SetSize: s,
@@ -67,6 +73,8 @@ func (bs *BrickSet) Contents() []*BrickEntry {
 	return out
 }
 
+// IsSparse returns true if the brick tracking slice contains
+// any nulls, indicating that the BrickSet is not fully populated.
 func (bs *BrickSet) IsSparse() bool {
 	for _, b := range bs.Bricks {
 		if b == nil {
@@ -76,6 +84,8 @@ func (bs *BrickSet) IsSparse() bool {
 	return false
 }
 
+// Full returns true if the brick slice contains the same
+// number of items as the expected set size.
 func (bs *BrickSet) Full() bool {
 	return len(bs.Bricks) == bs.SetSize
 }
@@ -102,10 +112,16 @@ type DeviceSet struct {
 	Devices []*DeviceEntry
 }
 
+// NewDeviceSet creates a device set tracking object with
+// the expected final size of s.
 func NewDeviceSet(s int) *DeviceSet {
 	return &DeviceSet{SetSize: s, Devices: []*DeviceEntry{}}
 }
 
+// NewSparseDeviceSet creates a device set tracking object with
+// the expected final size of s and the device tracking slice
+// pre-sized to s. This means that the array can be directly
+// indexed but may contain null brick entries.
 func NewSparseDeviceSet(s int) *DeviceSet {
 	return &DeviceSet{
 		SetSize: s,
@@ -135,10 +151,14 @@ func (ds *DeviceSet) Insert(index int, d *DeviceEntry) {
 	}
 }
 
+// Full returns true if the device slice contains the same
+// number of items as the expected set size.
 func (ds *DeviceSet) Full() bool {
 	return len(ds.Devices) == ds.SetSize
 }
 
+// IsSparse returns true if the device tracking slice contains
+// any nulls, indicating that the DeviceSet is not fully populated.
 func (ds *DeviceSet) IsSparse() bool {
 	for _, d := range ds.Devices {
 		if d == nil {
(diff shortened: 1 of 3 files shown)


apps/glusterfs/placer_arbiter.go
```diff
@@ -103,6 +103,12 @@ func (bp *ArbiterBrickPlacer) PlaceAll(
 		if err != nil {
 			return r, err
 		}
+		if bs.IsSparse() {
```
nixpanic: The 'Sparse' naming might be confusing me. Why would a SparseBrick/Device be an error here?

apps/glusterfs/brick_allocate.go
```diff
@@ -28,6 +28,13 @@ func NewBrickSet(s int) *BrickSet {
 	return &BrickSet{SetSize: s, Bricks: []*BrickEntry{}}
 }
 
+func NewSparseBrickSet(s int) *BrickSet {
```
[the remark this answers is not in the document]
phlogistonjohn: Heh. I always get in trouble with names. It's not that the brick set is sparse at the gluster level, its that the brick set object in heketi starts off "sparse" in that the array holds nulls but is sized up to the expected brick set set, rather than being an empty array.

It's not arbiter specific because I had wanted something like this earlier for some other re-factoring I did, but I convinced myself not to do it then. I will try to see if I can come up with a better name quickly.

apps/glusterfs/placer_arbiter.go
```diff
@@ -171,15 +177,20 @@ func (bp *ArbiterBrickPlacer) newSets(
...
 	}
 	defer dscan.Close()
 
-	for index := 0; index < ssize; index++ {
+	// work backwards from the last item in the brick set (typically index 2)
+	// in order to place the most special brick, the arbiter brick, first.
+	// Placing the arbiter brick first means we get a more reliable distribution
+	// of bricks because the nodes will not be "reserved" by the two data
+	// bricks before we try to place the arbiter brick.
+	for index := ssize - 1; index >= 0; index-- {
```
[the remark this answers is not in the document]
phlogistonjohn: It just means taken. I should have just wrote that. I will update the comment.

apps/glusterfs/brick_allocate.go
```diff
@@ -24,10 +24,23 @@ type BrickSet struct {
...
+// NewBrickSet creates an empty brick set tracking object with
+// the expected final size of s.
 func NewBrickSet(s int) *BrickSet {
 	return &BrickSet{SetSize: s, Bricks: []*BrickEntry{}}
 }
 
+// NewSparseBrickSet creates a brick set tracking object with
+// the expected final size of s and the brick tracking slice
+// pre-sized to s. This means that the array can be directly
+// indexed but may contains null brick entries.
```
ansiwen: typo: ...may contain null...

apps/glusterfs/brick_allocate.go
```diff
@@ -24,10 +24,23 @@ type BrickSet struct {
...
+func NewSparseBrickSet(s int) *BrickSet {
+	return &BrickSet{
+		SetSize: s,
+		Bricks:  make([]*BrickEntry, s),
```
ansiwen: I'm new to golang, but wouldn't it be better style, to make a new type from []*BrickEntry, like `type BrickSlice = []*BrickEntry` and use that instead?

apps/glusterfs/brick_allocate.go
```diff
@@ -76,10 +112,23 @@ type DeviceSet struct {
...
 }
 
+// NewSparseDeviceSet creates a device set tracking object with
+// the expected final size of s and the device tracking slice
+// pre-sized to s. This means that the array can be directly
+// indexed but may contains null brick entries.
+func NewSparseDeviceSet(s int) *DeviceSet {
+	return &DeviceSet{
+		SetSize: s,
+		Devices: make([]*DeviceEntry, s),
```
ansiwen: Same as for []*BrickEntry

apps/glusterfs/placer_arbiter.go
```diff
@@ -148,6 +160,11 @@ func (bp *ArbiterBrickPlacer) Replace(
 		wds.Insert(i, d)
 	}
 	aopts := newArbiterOpts(opts)
+	// this is a mildly hacky way to deal with the higher level replace
```
ansiwen: I assume a non-hacky way is not feasible right now? ;-)

apps/glusterfs/placer_arbiter.go
```diff
@@ -234,7 +256,7 @@ func (bp *ArbiterBrickPlacer) tryPlaceBrickOnDevice(
 
 	logger.Debug("Trying to place brick on device %v", device.Info.Id)
 
-	for i, b := range bs.Bricks {
+	for i, b := range bs.Contents() {
```
ansiwen: Contents() is not a newly introduced function. Why the change?

apps/glusterfs/placer_arbiter.go
phlogistonjohn: Yup

apps/glusterfs/placer_arbiter.go
phlogistonjohn: Contents was introduced in the previous change. It returns a copy of the Bricks with all nulls removed for easy iteration.

apps/glusterfs/brick_allocate.go
phlogistonjohn: It is certainly do-able... I just don't see the value of adding that now ATM. If we were to do it I'd add it to the file where BrickEntry was defined.

apps/glusterfs/placer_arbiter.go
obnoxxx: Because newSets() is supposed to return a fully populated set (i.e. one without nil entries).
These checks might be more natural at the end of newSets().

Issue #931, opened 2017-12-05
kubernetes documentation is seriously outdated or wrong


metal3d:
Hi,
Glade to use heketi on OpenShift, it seems that it's a bit more complicated to deploy heketi on kubernetes with the provided documentation.

I followed https://github.com/heketi/heketi/blob/master/doc/admin/install-kubernetes.md and see some problems:

1. https://github.com/heketi/heketi/tree/master/extras/kubernetes => there is no `deploy-heketi-deployment.json` file, only a one deployment. So that, there is no way to know how to add secret and client name. At this time, installation worked with the provided deployment, but there is only one
2. daemonset are wrong for glusterfs pods, we need to use gluster/gluster-centos:latest
3. we also need to change :dev to :latest on heketi images in deployment
2. it is not explained how to add rolebinding to let heketi list pods, I did it to finalize installaiton:
   ```
    # change gluster:heketi-service-account to YOUR_NAMESPACE:heketi-service-account
    kubectl create clusterrolebinding gluster-role-binding \
    --clusterrole=cluster-admin \
    --serviceaccount=gluster:heketi-service-account
   ```
3. maybe an example of storageClass could be nice (at this time, without any user/secret I've got some problems)

I am not here to criticize the work, I am impressed by all that is proposed in this repository and I thank you for this service which greatly helps to make volume claims work in Kubernetes and Openshift.

I'm just a bit confused now that I've got installed everything and that I'm not able to know if I finished installation as expected :)

I'm ok to adapt documentation and maybe to change yaml files, but I'm not sure that the install process is ok on my own cluster.

phlogistonjohn:
Thank you for the feedback! I recently converted a bunch of wiki pages to these documentation files so I am not shocked to find they're outdated or have errors. Off the top of my head I know that your item 2 is correct because the old container images were recently removed and that generated other bug reports.

In addition to improving this documentation, are you aware of the [gluster-kubernetes project](https://github.com/gluster/gluster-kubernetes)? The deployment bits there are more up to date than this documentation AFAIK. The documentation there also goes over creating a storageClass, pvc, and so forth. Look over the readme, quickstart guide, and hello world example there for more info.

That said, I do want to improve this documentation but I also don't want you to be blocked on whatever you are working on.

metal3d:
I've forgotten that repo. 
I used it to create ansible playbook to install heketi on OpenShift when official repository didn't provide it :)

[...]

Issue #873, opened 2017-10-03
 deploy-heketi-deployment.json is missing when you do git clone


dimthe:
Hello i am trying to get heketi/glusterfs to work on a kube cluster deployed with kubespray on centos 7.3 and docker 1.13.1 , kubectl 1.7.4

i found the missing file from here 
/master/deploy/kube-templates/deploy-heketi-deployment.yaml

i changed the version heketi:tag to heketi:latest and although the yaml seems to be working 
```
kubectl create -f deploy-heketi-deployment.yaml
service "deploy-heketi" created
deployment "deploy-heketi" created
```
it does not create a heketi pod

 ```
kubectl get pods
NAME              READY     STATUS    RESTARTS   AGE
glusterfs-2jnkx   1/1       Running   0          23m
glusterfs-352wm   1/1       Running   0          23m
glusterfs-gt2qx   1/1       Running   0          23m
glusterfs-tf5pj   1/1       Running   0          23m
```

i have manually done a docker pull heketi/heketi:latest and i see the image
```
docker images | grep heketi
heketi/heketi
```

having the image , i deleted the above deployment and re run kubectl create -f hoping it would work but i still see no heketi pod

is there a fix ? thanks

obnoxxx:
As mentioned in issue #881, have you had a look at https://github.com/gluster/gluster-kubernetes ?

obnoxxx:
@dimthe , any luck with gluster-kubernetes?

phlogistonjohn:
Last comment was made nearly a year ago. Unless you want to hack on heketi directly it is recommended to deploy via gluster-kubernetes. Feel free to reopen if you wish to discuss further.

Issue #820, opened 2017-08-02
update documentation for kubernetes


roffe:
The documentation is faulty:

https://github.com/heketi/heketi/wiki/Kubernetes-Integration

deploy-heketi-deployment.json  <-- Where does this file come from, it's not in your repo.

roffe:
i used heketi-deployment.json in the first step as the file you are pointing to is missing.

roffe:
so it turns out i found the file in the heketi-cli release package and not a github repo as you state in the documentation. 

also it's in a openshift folder and not kubernetes.

roffe:
New try today. This time even more fail:

```
Creating cluster ... ID: b279a7477e4f22f3749c896484f22e0f
	Creating node 10.1.201.181 ... ID: e75a1d0facd9e8a19c929714e181eb5d
		Adding device /dev/sdb ... Unable to add device: No pods with the label 'glusterfs-node=10.1.201.181' were found
	Creating node 10.1.201.182 ... Unable to create node: No pods with the label 'glusterfs-node=10.1.201.181' were found
	Creating node 10.1.201.183 ... Unable to create node: No pods with the label 'glusterfs-node=10.1.201.181' were found
```

```
➜  heketi git:(master) ✗ kc get po -o wide
NAME                             READY     STATUS    RESTARTS   AGE       IP             NODE
deploy-heketi-4120581439-h88k0   1/1       Running   0          4m        10.10.7.44     10.1.201.204
glusterfs-5bb82                  1/1       Running   0          6m        10.1.201.181   10.1.201.181
glusterfs-6wd3w                  1/1       Running   0          6m        10.1.201.183   10.1.201.183
glusterfs-w0pd7                  1/1       Running   0          6m        10.1.201.182   10.1.201.182
```

roffe:
the glusterf daemonset is installed from glusterfs-daemonset.json provided by heketi

Please update your docs page with a big warning that HEKETI is currently broken and cannot be deployed on kubernetes to prevent more people from wastng their time until your docs examples are up to speed and working again

wileeam:
@roffe Maybe you want to be more explicit in your problem and if the documentation is that faulty a PR can help you and everyone else.
I am actually running GlusterFS+Heketi in a cluster of two nodes (and not three) following mostly the  documentation available here and debugging the errors on the way (because you are right that the documentation is not exact).
[...]

Issue #807, opened 2017-07-15
No deploy-heketi-deployment.json


Lazarus404:
I'm following the Kubernetes integration docs, but have found there is no `deploy-heketi-deployment.json` file.  Where might I get it?  Is the integration docs out of date?

Thanks

raghavendra-talur:
Are you looking for this file? https://github.com/gluster/gluster-kubernetes/blob/master/deploy/kube-templates/deploy-heketi-deployment.yaml

Lazarus404:
I thought that, but was following a blog post and the getting started guide that both said the file should already be in extras/kubernetes in this repo, so wasn't certain.

huguesalary:
Same here. I've been following the instructions here https://github.com/heketi/heketi/wiki/Kubernetes-Integration and they mention the  `deploy-heketi-deployment.yaml` file, which indeed does not exist in `extras/`. Although the file does exist at the location @raghavendra-talur pointed out, it'd be great to update the wiki and/or the `extras/` folder.

There's also conflicting information between the `README.md` in `https://github.com/heketi/heketi/tree/master/extras/kubernetes` and `https://github.com/heketi/heketi/wiki/Kubernetes-Integration`.

One mentions using `glusterfs-daemonset.json` but not the other, and most of the steps in `https://github.com/heketi/heketi/tree/master/extras/kubernetes/README.md` are missing. Removing a set of instructions and consolidating the doc in 1 single place would be helpful.

amolsh:
@raghavendra-talur  As you said , i used https://github.com/gluster/gluster-kubernetes/blob/master/deploy/kube-templates/deploy-heketi-deployment.yaml

But, it did not create required pod. As per documentation, it creates one running pod.

Looks like documentation is all messy. Any idea where can I find its proper installation steps

prapdm:
+1

wernergillmer:
Trying to use the deploy-heketi-deployment.yaml file pointed out by @raghavendra-talur but it's using a secret, that I don't have, and I don't see where it get's created. What should be in the secret? Anyone got some documentation that's up to date?

obnoxxx:
Please refer to the documentation and tools in the github.com/gluster/gluster-kubernetes repository for instructions, artifacts, and tools to deploy heketi and gluster into kubernetes.

There are several duplicates of this issue, including #873.

obnoxxx:
The kubernetes related documentation in this repo is outdated  and needs to be updated/fixed/removed.

[...]

PR #1179, opened 2018-05-14
Direct kubernetes users to gluster-kubernetes

### LINKED ISSUE #1177: update DOC for extras/kubernetes
Hi,

in extras/kubernetes. deploy-heketi-deployment.json still used in the Readme.md file otherwise it does not exist in the files. as consequence i am unable to install it properly.

thank you

joejulian:
This documentation has been deprecated. Instead, direct users to install
using gluster-kubernetes.

Fixes #1177, #931, #873, #820, #807, #587


nixpanic:
ok to test

joejulian:
Sounds good. Done.

phlogistonjohn:
I'd like to preserve some of this document (under a different name perhaps) as its the only prose documentation that shows how heketi works with kubernetes. Directing people away from this doc for production is great. However, I'd hate to lose some of the institutional knowledge it contains (git history doesn't count due it's low discover-ability).

Can we rename the doc to something like "how heketi integrates with kubernetes" and orient it more towards a document for those who are curious about how heketi works and/or potential contributors?

joejulian:
I tried following those docs and when I failed, I looked in issues and saw that everyone was being pointed to gk-deploy. Leaving them in would imply some sort of maintained correctness which, afaict, isn't the case.

Do any of the maintainers have the time or inclination to maintain any of this? If so, which bits should remain and be maintained? I'll happily make the change to this PR.

phlogistonjohn:
I went through that guide and updated issues that I found within the last 4 months or so. But like I mentioned, I see this as more developer oriented docs telling people how the peices fit together rather than a guide on how to get heketi running in kubernetes for production. I think most people want the latter and that's why they get pointed at things like g-k  or openshift ansible.

Personally, I"d just start by renaming the doc (both the filename and the title). Perhaps moving it out from under admin and under design and putting more emphasis in the into that this is background rather than a complete HOW-TO. I'd retain the majority of the body of the doc. As for the templates, I think we can discuss if we want to simply refer to the templates that live in g-k and completely remove them from this repo or move them somewhere under the docs tree.

phlogistonjohn:
@joejulian any interest in this still? If you've lost interest in this PR I can make an alternative PR that does as I requested: rename original doc, add more strenuous pointer to g-k, reword starting section of current doc to indicate is is for developer & educational purposes.

Any thoughts?

jarrpa:
@phlogistonjohn I'd say ten days is more than enough time to wait for a response. Probably fine to go ahead with your idea.

phlogistonjohn:
Good point. It will probably take me a few to get my PR together anyway. :-)

Issue #1181, opened 2018-05-15
heketi is not creating volumes when we give very large average file size


iamniting:
### Kind of issue
Bug

### Observed behavior
Heketi is not creating a volume.

### Expected/desired behavior
It should create volume with minimum (16 mb) arbiter brick size.

### Details on how to reproduce (minimal and precise)
$ heketi-cli volume create --size=10 --gluster-volume-options='user.heketi.arbiter true,user.heketi.average-file-size 1024'

### Information about the environment:

- Heketi version used (e.g. v6.0.0 or master): 6.0.0-12 
- Operating system used: RHEL
- Heketi compiled from sources, as a package (rpm/deb), or container: Container & RPM 
- If container, which container image: 3.3.1-12
- Using kubernetes, openshift, or direct install: brew
- If kubernetes/openshift, is gluster running inside kubernetes/openshift or outside: inside
- If kubernetes/openshift, how was it deployed (gk-deploy, openshift-ansible, other, custom): ansible

### Other useful information

```
Error: Unable to execute command on glusterfs-storage-dz65l: agsize (3072 blocks) too small, need at least 4096 blocks
Usage: mkfs.xfs
/* blocksize */		[-b log=n|size=num]
/* metadata */		[-m crc=0|1,finobt=0|1,uuid=xxx]
/* data subvol */	[-d agcount=n,agsize=n,file,name=xxx,size=num,
			    (sunit=value,swidth=value|su=num,sw=num|noalign),
			    sectlog=n|sectsize=num
/* force overwrite */	[-f]
/* inode size */	[-i log=n|perblock=n|size=num,maxpct=n,attr=0|1|2,
			    projid32bit=0|1]
/* no discard */	[-K]
/* log subvol */	[-l agnum=n,internal,size=num,logdev=xxx,version=n
			    sunit=value|su=num,sectlog=n|sectsize=num,
			    lazy-count=0|1]
/* label */		[-L label (maximum 12 characters)]
/* naming */		[-n log=n|size=num,version=2|ci,ftype=0|1]
/* no-op info only */	[-N]
/* prototype file */	[-p fname]
/* quiet */		[-q]
/* realtime subvol */	[-r extsize=num,size=num,rtdev=xxx]
/* sectorsize */	[-s log=n|size=num]
/* version */		[-V]
			devicename

```

obnoxxx:
Thanks, @iamniting , I am working on a fix for this.

PR #1183, opened 2018-05-15
arbiter: fix volume creation with too big average file size

obnoxxx:
The arbiter brick size for a brick set is calculated as
(dataBrickSize)/(averageFileSize). If the data brick is
rather small, and the averageFileSize is rather big, this
can get smaller than the minimum filesystem size of 16MiB
that xfs supports. Hence always use at least 16 MiB for
the arbiter brick.

Note this required to change two of the test cases
slightly: The sizes are actually in KB. The tests use
some fantasy numbers like 800 KiB brick sizes that now
contradict for some checks the new mode of having a
minimum arbiter brick size of 16 MiB.

Fixes: #1181

### DIFF
--- a/apps/glusterfs/placer_arbiter.go
+++ b/apps/glusterfs/placer_arbiter.go
@@ -371,13 +371,25 @@ func (dscan *arbiterDeviceScanner) Scan(index int) <-chan string {
 	return dscan.dataDevs
 }
 
-func discountBrickSize(dataBrickSize, averageFileSize uint64) (uint64, error) {
+func discountBrickSize(dataBrickSize, averageFileSize uint64) (brickSize uint64,
+	err error) {
+
 	if dataBrickSize < averageFileSize {
 		return 0, fmt.Errorf(
 			"Average file size (%v) is greater than Brick size (%v)",
 			averageFileSize, dataBrickSize)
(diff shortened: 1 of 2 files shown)


obnoxxx:
Updated to fix gofmt issue.

apps/glusterfs/placer_arbiter.go
```diff
@@ -371,13 +371,24 @@ func (dscan *arbiterDeviceScanner) Scan(index int) <-chan string {
...
 	if dataBrickSize < averageFileSize {
 		return 0, fmt.Errorf(
 			"Average file size (%v) is greater than Brick size (%v)",
 			averageFileSize, dataBrickSize)
 	}
-	return (dataBrickSize / averageFileSize), nil
+
+	brickSize = dataBrickSize / averageFileSize
+
+	if brickSize < 16*MB {
```
nixpanic: I would have loved a comment that this is the minimal device size for XFS. Oh well, its in the commit message at least :-)

obnoxxx: Good point. Updated the patch to a more verbose log message instead of adding a comment. This way, it is even apparent from the log why this happened...

PR #1185, opened 2018-05-16
Add an option to destroy data while adding a device to a node

nixpanic:

Some users are certain that the data on the disks that they add to a node can be destroyed. It can be useful to force adding a device, even if `pvcreate` detects existing data. This PR adds an option to the "device add" API and implements the functionality to provide this feature.

### Does this PR fix issues?

BUG: https://bugzilla.redhat.com/1433614

### Notes for the reviewer

A short discussion was also held on the heketi-devel list: http://lists.gluster.org/pipermail/heketi-devel/2018-May/000267.html

### DIFF
--- a/client/cli/go/cmds/topology.go
+++ b/client/cli/go/cmds/topology.go
@@ -30,8 +30,13 @@ const (
 var jsonConfigFile string
 
 // Config file
-type ConfigFileDevice struct {
+type ConfigFileDeviceOptions struct {
 	api.Device
+	DestroyData bool `json:"destroydata,omitempty"`
+}
+
+type ConfigFileDevice struct {
+	ConfigFileDeviceOptions
 }
 type ConfigFileNode struct {
 	Devices []*ConfigFileDevice `json:"devices"`
@@ -59,13 +64,18 @@ func (device *ConfigFileDevice) UnmarshalJSON(b []byte) error {
 		device.Name = s
 		return nil
 	}
-	var d api.Device
+
+	// ConfigFileDevice embeds the ConfigFileDeviceOptions struct which has
+	// additional members compared to the standard api.Device. Structuring
+	// it this way, prevents a recursive call to UnmarshalJSON().
+	var d ConfigFileDeviceOptions
 	err = json.Unmarshal(b, &d)
 	if err != nil {
 		return err
 	}
 	device.Name = d.Name
 	device.Tags = d.Tags
+	device.DestroyData = d.DestroyData
 	return nil
 }
 
@@ -244,6 +254,7 @@ var topologyLoadCommand = &cobra.Command{
 						req.Name = device.Name
 						req.NodeId = nodeInfo.Id
 						req.Tags = device.Tags
+						req.DestroyData = device.DestroyData
 						err := heketi.DeviceAdd(req)
 						if err != nil {
 							fmt.Fprintf(stdout, "Unable to add device: %v\n", err)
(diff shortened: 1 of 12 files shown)


Madhu-1:
@nixpanic need to update topology.json in the gluster-kubernetes also (once this PR is merged)

client/cli/go/cmds/topology.go
```diff
@@ -59,13 +60,22 @@ func (device *ConfigFileDevice) UnmarshalJSON(b []byte) error {
 		device.Name = s
 		return nil
 	}
-	var d api.Device
+
+	// Using ConfigFileDevice will cause an endless loop while calling
+	// Unmarshal() on it. A copy of the struct with a different name
+	// prevents the recursive Unmarshal() call.
+	type _ConfigFileDevice struct {
```
phlogistonjohn: I think that you can do an embed rather than a literal copy-n-paste like so:
```golang
type _ConfigFileDevice struct {
  ConfigFileDevice
}
```

This way any other changes to the "parent" ConfigFileDevice don't need to be manually copied. If this doesn't work let me know because there's something along these lines that should.

phlogistonjohn:
> need to update topology.json in gluster-kubernetes (once this PR is merged).

Only if you want to advertise the new feature. The topology parsing is backwards compatible with the older contents of the topology, it just works as before... no destroy will take place.

client/cli/go/cmds/topology.go
nixpanic: Good idea, but it does not seem to work. This calls `ConfigFileDevice.Unmarshal()` and results in an recursive loop again, resulting in a stack overflow.

phlogistonjohn: Ah. I think we need to create an intermediate type with the fields you want (say ConfigFileDevice) and then the type that actually has the custom unmarhal method (_ConfigFileDevice) rather than the current approach which is has that relationship inverted. If I'm not making sense let me know and I'll try to demonstrate what I mean with a patch.

nixpanic: Ok, I think I understood what you mean. How about the updated version?

phlogistonjohn:
> But do have a few concerns:
>
 > * I am not too fond of the required topology changes. Ideally, without changing anything, the
> behavior should stay the same, i.e. not specifying anything apart from the device name, should not wipe the devices, just as before. Not entirely sure what happens in this current form. So I would like to not change the topology files in the test cases.

I agree that none of the topology files uses for test cases should be changed. I did't think any of the files @nixpanic changed are used in test cases but I guess I missed that there were two last time I looked at the patches. 

We should probably revert those parts, and/or make explicit tests for this. I'd be OK for that to be in a followup PR (& if @obnoxxx  is OK with that).

> * We should have a case that tests this, in the sens of: bring up a setup, take it down and bring up a new one over it with the destroy option.

See above. FWIW, we don't do much in the way of testing heketi-cli right now. Our tests mainly cover the server code and the client APIs. 

> * requiring the flag on each single device is somewhat cumbersome. wouln't it be more convenient to be able to pass a --destroy-data option to the topology load command that would propagate into all implied device add operations?

I was thinking along these lines and was wondering if that's actually useful AFAIK, the two common ways of installing a cluster for use with heketi are gluster-kubernetes and openshift-ansible. Neither method exposes the heketi-cli topology load command directly to the user. The g-k method has you put a topology file in place but runs the command for you. The openshift-ansible method generates a topology automatically behind the scenes.

In the former case we'd need to plumb the option thru gk-deploy.  In the latter method I don't think it would be an adantage at all because the part of the ansible that generates the topology could just generate the "destroydata" keys.

I don't know if I'm making sense. :-)

obnoxxx:
retest this please

jarrpa:
Should be a relatively easy change in openshift-ansible. I already have a variable that triggers basically this same behavior, so it'd be a matter of changing from doing it manually to doing it through heketi.

PR #1187, opened 2018-05-18
executors: do not create backups in /etc/lvm/archive by default

nixpanic:

With the automation that Heketi provides and the redundancy that Gluster
offers there is no need for the LVM-metadata backups under
/etc/lvm/archive. It is not expected that users have a need for these
archives, and they can take up a considerable amount space on the /
filesystem in active environments.

The heketi.json used by the service daemon can be configured to create
the LVM-metadata archives by setting "backup_lvm_metadata" to true. This
option defaults to false if not set.

### Does this PR fix issues?

BUG: https://bugzilla.redhat.com/1561680

### DIFF
--- a/executors/cmdexec/brick_test.go
+++ b/executors/cmdexec/brick_test.go
@@ -18,13 +18,7 @@ import (
 	"github.com/heketi/tests"
 )
 
-func TestSshExecBrickCreate(t *testing.T) {
-	f := NewCommandFaker()
-	s, err := NewFakeExecutor(f)
-	tests.Assert(t, err == nil)
-	tests.Assert(t, s != nil)
-	s.portStr = "100"
-
+func doTestSshExecBrickCreate(t *testing.T, f *CommandFaker, s *FakeExecutor) {
 	// Create a Brick
 	b := &executors.BrickRequest{
 		VgId:             "xvgid",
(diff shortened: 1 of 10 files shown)


executors/cmdexec/brick_test.go
```diff
@@ -52,8 +46,12 @@ func TestSshExecBrickCreate(t *testing.T) {
 					cmd == "mkdir -p /var/lib/heketi/mounts/vg_xvgid/brick_id", cmd)
 
 			case 1:
+				autobackup := "n"
```
phlogistonjohn: You do this in a bunch of places. Perhaps create a short helper function?

nixpanic: I could do that, does `utils.BoolToYN(b bool)` sound acceptible?

phlogistonjohn: I was originally thinking something more specific to the BackupLVM option, but that'd be fine too. Anything to cut down on the boilerplate and reduce the chance of typos.

nixpanic:
Now with added `BoolToYN(b bool) string` function in `pkg/utils/conv.go` (+ `conv_test.go`).

This looks indeed a little cleaner :-)

pkg/utils/conv.go
```diff
@@ -0,0 +1,18 @@
...
+//
+// This file is licensed to you under your choice of the GNU Lesser
+// General Public License, version 3 or any later version (LGPLv3 or
+// later), or the GNU General Public License, version 2 (GPLv2), in all
+// cases as published by the Free Software Foundation.
+//
+
+package utils
+
+// BrickDevNode returns a "y" (yes) or "n" (no) for the passed bool.
```
phlogistonjohn: BrickDevNode ??

nixpanic: oh, yay for copy/paste!

nixpanic:
Comments for `BoolToYN()` should be good now too.

PR #1191, opened 2018-05-18
bias the arbiter placer to prefer dedicated devices before general devices

phlogistonjohn:

In testing it became apparent that tagging scenarios where a user
tags some nodes or devices for arbiter:required but leaves other
items untagged seems natural to a user. However, prior to this change
it had the effect of making placement failures highly likely on a
small cluster. This was due to the arbiter brick "stealing" a node
that needed to be used by a data brick, when the arbiter brick
was placed on an (implicit) arbiter:supported device rather
than the arbiter:required device.

What this change does is to prioritize devices that are tagged for
one type of brick when placing a brick of that type. In other words,
if any devices are tagged arbiter:required the placer prefers
to use those devices before it uses arbiter:supported devices.
The same goes for preferring to place data bricks on arbiter:disabled
devices before trying arbiter:supported devices.

### Does this PR fix issues?

Fixes [RHBZ #1578658](https://bugzilla.redhat.com/show_bug.cgi?id=1578658)

### Notes for the reviewer

Let me know if you have any ideas about if there are any additional test cases we could add for this.

### DIFF
--- a/apps/glusterfs/placer_arbiter.go
+++ b/apps/glusterfs/placer_arbiter.go
@@ -306,11 +306,28 @@ func (bp *ArbiterBrickPlacer) tryPlaceBrickOnDevice(
 	return nil
 }
 
+type deviceFeed struct {
+	Devices <-chan string
+	Done    chan struct{}
+}
+
 type arbiterDeviceScanner struct {
-	arbiterDevs <-chan string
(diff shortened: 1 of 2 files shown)


phlogistonjohn:
retest this please

PR #1199, opened 2018-05-29
Improve error handling on volume create

### LINKED ISSUE #712: Need a better error message when a cluster id with no bricks is provided
When a cluster is created manually without any nodes or bricks and when such a cluster id is defined as part of storage class, pvc creation fails with 'Error calling v.allocBricksInCluster: Id not found'. Looks like we just return the error received from the function. A better error message should be provided.

ProvisioningFailed	Failed to provision volume with StorageClass "voltypenegative": glusterfs: create volume err: error creating volume Error calling v.allocBricksInCluster: Id not found.

phlogistonjohn:

This series improves the handling of error messages on volume create by:
1. Adding specific errors for when the cluster has no nodes or none of the nodes have usable devices
2. Adding the ability to capture and report on each cluster's error by aggregating them in a new MultiClusterError type, as opposed to coercing all errors to "No Space"
3. Adds tests for the above

Example output:
```
heketi-cli volume create --size=4
Error: Failed to allocate new volume: Unable to create volume on any cluster:
Cluster 4081998cadcd0c9093f2c2f0fc813430: No nodes in cluster
Cluster a54cc7f2711c1711b8a6aef195681893: No online storage devices in cluster
```

### Does this PR fix issues?

Fixes #712 

### Notes for the reviewer

Type name and error text wording bike-shedding welcome.

### DIFF
--- a/apps/glusterfs/errors.go
+++ b/apps/glusterfs/errors.go
@@ -25,4 +25,8 @@ var (
 	ErrKeyExists        = errors.New("Key already exists in the database")
 	ErrNoReplacement    = errors.New("No Replacement was found for resource requested to be removed")
 	ErrCloneBlockVol    = errors.New("Cloning of block hosting volumes is not supported")
+
+	// well known errors for cluster device source
+	ErrEmptyCluster = errors.New("No nodes in cluster")
+	ErrNoStorage    = errors.New("No online storage devices in cluster")
 )
--- a/apps/glusterfs/operations_test.go
+++ b/apps/glusterfs/operations_test.go
@@ -321,7 +321,8 @@ func TestVolumeCreatePendingNoSpace(t *testing.T) {
 
 	e := vc.Build()
 	// verify that we failed to allocate due to lack of space
-	tests.Assert(t, e == ErrNoSpace, "expected e == ErrNoSpace, got", e)
+	tests.Assert(t, strings.Contains(e.Error(), ErrNoSpace.Error()),
+		"expected strings.Contains(e.Error(), ErrNoSpace.Error()) got", e)
 
 	// verify no volumes, bricks or pending ops in db
 	app.db.View(func(tx *bolt.Tx) error {
--- a/apps/glusterfs/volume_entry.go
+++ b/apps/glusterfs/volume_entry.go
@@ -952,3 +952,69 @@ func (v *VolumeEntry) cloneVolumeRequest(db wdb.RODB, clonename string) (*execut
 
 	return vcr, sshhost, nil
 }
+
+type MultiClusterError struct {
+	prefix string
+	errors map[string]error
+}
+
+// NewMultiClusterError returns a MultiClusterError with the given
+// prefix text. Prefix text will be used in the error string if
+// more than one error is captured.
+func NewMultiClusterError(p string) *MultiClusterError {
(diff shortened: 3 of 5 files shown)


phlogistonjohn:
retest this please

apps/glusterfs/errors.go
```diff
@@ -25,4 +25,8 @@ var (
 	ErrKeyExists        = errors.New("Key already exists in the database")
 	ErrNoReplacement    = errors.New("No Replacement was found for resource requested to be removed")
 	ErrCloneBlockVol    = errors.New("Cloning of block hosting volumes is not supported")
+
+	// well known errors for cluster device source
+	EmptyClusterErr = errors.New("No nodes in cluster")
+	NoStorageErr    = errors.New("No online storage devices in cluster")
```
obnoxxx: I would prefer to stick with the existing naming scheme:
```
ErrEmptyCluster = ...
ErrNoStorage     = ...
```

apps/glusterfs/operations_test.go
```diff
@@ -321,7 +321,8 @@ func TestVolumeCreatePendingNoSpace(t *testing.T) {
 
 	e := vc.Build()
 	// verify that we failed to allocate due to lack of space
-	tests.Assert(t, e == ErrNoSpace, "expected e == ErrNoSpace, got", e)
+	tests.Assert(t, strings.Contains(e.Error(), ErrNoSpace.Error()),
+		"expected trings.Contains(e.Error(), ErrNoSpace.Error()) got", e)
```
obnoxxx: typo "trings -> strings"

apps/glusterfs/volume_entry.go
```diff
@@ -952,3 +968,72 @@ func (v *VolumeEntry) cloneVolumeRequest(db wdb.RODB, clonename string) (*execut
...
+}
+
+// Shorten returns a simplified version of the errors that
+// the MultiClusterError may have captured. It returns nil if
+// no errors were captured. It returns itself if more than one
+// error was captured. It returns the original error if only
+// one error was captured.
+func (m *MultiClusterError) Shorten() error {
+	if len(m.errors) == 0 {
+		return nil
```
obnoxxx: this if block is redundant: it is also in the switch statement under `case 0:`

apps/glusterfs/volume_entry.go
```diff
@@ -952,3 +968,72 @@ func (v *VolumeEntry) cloneVolumeRequest(db wdb.RODB, clonename string) (*execut
...
+	}
+	switch len(m.errors) {
+	case 0:
+		return nil
+	case 1:
+		for _, err := range m.errors {
+			return err
+		}
```
obnoxxx: it is slightly counter-intuitive for me to use a for loop to access the only element

apps/glusterfs/errors.go
phlogistonjohn: Oops. Don't know how I missed that.

apps/glusterfs/volume_entry.go
phlogistonjohn: Me too. Unfortunately, go does not have any good mechanisms for this. This was the most succinct thing I could come up with. Suggestions welcome, though.

PR #1202, opened 2018-06-04
extras: add heketi containers for the CentOS Registry

nixpanic:
### What does this PR achieve? Why do we need it?

Add container images for the CentOS Container Registry that is available at https://registry.centos.org/containers/ . The images are built from RPMs that are available from the CentOS Storage SIG.

Currently two flavours are provided:
1. built from the CentOS Storage SIG stable/released repository
2. built from the CentOS Storage SIG testing repository

The description to build the containers for the CentOS Container Registry is available from [the add-gluster-heketi branch](https://github.com/nixpanic/centos-container-index/commit/5476a48b72a4e192318f5301a6da155ca1b93908) in my fork of the [container-index repository](https://github.com/CentOS/container-index). Once this PR has been merged, a PR for the container-index will be created.

PR #1205, opened 2018-06-05
Dockerfile: update the Maintainer

obnoxxx:

The dockerfile lists Luis as the maintainer. But Luis is no longer working on heketi, so update the dockerfile.

### DIFF
--- a/extras/docker/fromsource/Dockerfile
+++ b/extras/docker/fromsource/Dockerfile
@@ -1,6 +1,6 @@
 # set author and base
 FROM fedora
-MAINTAINER Luis Pabón <lpabon@redhat.com>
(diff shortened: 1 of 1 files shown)


extras/docker/fromsource/Dockerfile
```diff
@@ -1,6 +1,6 @@
 # set author and base
 FROM fedora
-MAINTAINER Luis Pabón <lpabon@redhat.com>
+MAINTAINER Michael Adam <obnox@redhat.com>
```
nixpanic: What about using "Heketi Developers" and the mailinglist for contact details? I'v done that in #1202 as well.

obnoxxx: Done, thanks!

PR #1206, opened 2018-06-05
Fix heketi startup with orphaned cluster->volume and volume-> brick links

obnoxxx:

Currenly, when a cluster lists a volumeid, that does not exist in the db, and when a volume lists a brick id that does not exist in the db, heketi will refuse to start with the upgrade failing. This is unnecessary since those orphan backlinks are actually harmless.  This PR improves the upgrade code to  not fail over these.

### DIFF
--- a/apps/glusterfs/brick_entry.go
+++ b/apps/glusterfs/brick_entry.go
@@ -274,6 +274,12 @@ func addVolumeIdInBrickEntry(tx *bolt.Tx) error {
 		}
 		for _, brick := range volumeEntry.Bricks {
 			brickEntry, err := NewBrickEntryFromId(tx, brick)
+			if err == ErrNotFound {
+				logger.Warning("Volume [%v] links to "+
+					"nonexistent brick [%v]. Ignoring.",
+					volume, brick)
+				continue
+			}
 			if err != nil {
 				return err
 			}
--- a/apps/glusterfs/brick_entry.go
+++ b/apps/glusterfs/brick_entry.go
@@ -263,34 +263,28 @@ func BrickEntryUpgrade(tx *bolt.Tx) error {
 }
 
 func addVolumeIdInBrickEntry(tx *bolt.Tx) error {
-	clusters, err := ClusterList(tx)
+	volumes, err := VolumeList(tx)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


phlogistonjohn:
centos ci failed due to yum/network issues

retest this please

obnoxxx:
error starting vms. 

retest this please

obnoxxx:
@phlogistonjohn heh, i did not see your update... sorry for repeating :-)

apps/glusterfs/brick_entry.go
```diff
@@ -274,6 +274,11 @@ func addVolumeIdInBrickEntry(tx *bolt.Tx) error {
 		}
 		for _, brick := range volumeEntry.Bricks {
 			brickEntry, err := NewBrickEntryFromId(tx, brick)
+			if err == ErrNotFound {
+				logger.Warning("Volume [%v] liks to orphaned "+
```
phlogistonjohn: Typo: liks -> links
Also, Heketi doesn't really have the habit of putting brackets around strings. Since it only ever prints IDs I don't think they aid readability. This is minor tho.

obnoxxx:
@phlogistonjohn fixed the typo, and also replaced "orphaned" by "inexistent" in the same log message.

While not all of heketi is using [...] quoting, I like it since it improves the readability to me, especially if in some cases, the string is empty, etc... So I am trying to use it wherever I can. :-D

apps/glusterfs/brick_entry.go
```diff
@@ -274,6 +274,12 @@ func addVolumeIdInBrickEntry(tx *bolt.Tx) error {
 		}
 		for _, brick := range volumeEntry.Bricks {
 			brickEntry, err := NewBrickEntryFromId(tx, brick)
+			if err == ErrNotFound {
+				logger.Warning("Volume [%v] links to "+
+					"inexistent brick [%v]. Ignoring.",
```
phlogistonjohn: Ok, I won't block on the brackets.

However, inexistant is not a normal english word (see https://en.wiktionary.org/wiki/inexistant - which considers it obsolete usage). Do you mind changing it to nonexistent?

obnoxxx:
fixed inexistent --> nonexistent

PR #1213, opened 2018-06-06
don't let DeleteBricksWithEmptyPath stumble over broken links in the DB

obnoxxx:

If the DB has broken links (list of node-ids in the cluster, list of device-ids in the node, list of brick-ids in the device), DeleteBricksWithEmptyPath will fail. This PR lets DeleteBricksWithEmptyPath ignore these harmless links pointing to nowhere, but logging warning messages about them.

### DIFF
--- a/apps/glusterfs/cluster_entry.go
+++ b/apps/glusterfs/cluster_entry.go
@@ -218,6 +218,11 @@ func (c *ClusterEntry) DeleteBricksWithEmptyPath(tx *bolt.Tx) error {
 
 	for _, nodeid := range c.Info.Nodes {
 		node, err := NewNodeEntryFromId(tx, nodeid)
+		if err == ErrNotFound {
+			logger.Warning("Ignoring nonexisting node [%v] in "+
+				"cluster [%v].", nodeid, c.Info.Id)
+			continue
+		}
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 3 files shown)


apps/glusterfs/cluster_entry.go
```diff
@@ -218,6 +218,10 @@ func (c *ClusterEntry) DeleteBricksWithEmptyPath(tx *bolt.Tx) error {
 
 	for _, nodeid := range c.Info.Nodes {
 		node, err := NewNodeEntryFromId(tx, nodeid)
+		if err == ErrNotFound {
+			logger.Warning("Ignoring nonexisting node [%v] in "+
+				"cluster [%v].", nodeid, c.Info.Id)
+		}
```
phlogistonjohn: Did you mean to add a `continue` statement to this code block?

obnoxxx: Gosh, thanks! Done...

obnoxxx: Maybe I should add unit tests for this new behavior?!... :-)

phlogistonjohn: I'd be all in favor of that. Or if you wish you could add to the functional tests for upgrade @ tests/functional/TestUpgrade

phlogistonjohn:
Fix looks good. Please let me know if you want to add tests to this PR or else we can just merge it.

obnoxxx:
I'd rather add later. There is no existing test to just add this aspect to.

PR #1216, opened 2018-06-08
Ignore nonexistent block volumes in deletion of block-hosting volume.

obnoxxx:

Don't fail the deletion of a block-hosting volume if
it references a block volume id, that does not
exist in the database.

### DIFF
--- a/apps/glusterfs/app_volume.go
+++ b/apps/glusterfs/app_volume.go
@@ -252,17 +252,22 @@ func (a *App) VolumeDelete(w http.ResponseWriter, r *http.Request) {
 			return nil
 		}
 
-		if volume.Info.BlockInfo.BlockVolumes == nil {
-			return nil
+		for _, bvId := range volume.Info.BlockInfo.BlockVolumes {
+			_, err = NewBlockVolumeEntryFromId(tx, bvId)
+			if err == nil {
+				err = logger.LogError("Cannot delete a block hosting volume containing block volumes")
+				http.Error(w, err.Error(), http.StatusConflict)
+				return err
+			}
+			if err != ErrNotFound {
+				err = logger.LogError("Refusing to delete block-hosting volume: "+
+					"Error loading block-volume [%v]: %v", bvId, err)
+				http.Error(w, err.Error(), http.StatusInternalServerError)
+				return err
+			}
 		}
 
-		if len(volume.Info.BlockInfo.BlockVolumes) == 0 {
-			return nil
-		}
-
-		err = logger.LogError("Cannot delete a block hosting volume containing block volumes")
-		http.Error(w, err.Error(), http.StatusConflict)
-		return err
+		return nil
 	})
 	if err != nil {
 		return


apps/glusterfs/app_volume.go
```diff
@@ -252,17 +252,22 @@ func (a *App) VolumeDelete(w http.ResponseWriter, r *http.Request) {
...
+			_, err = NewBlockVolumeEntryFromId(tx, bvId)
+			if err == nil {
+				err = logger.LogError("Cannot delete a block hosting volume containing block volumes")
+				http.Error(w, err.Error(), http.StatusConflict)
+				return err
+			}
+			if err != ErrNotFound {
+				err = logger.LogError("Refusing to delete block-hosting volume: "+
+					"Error loading block-volume [%v]: %v", bvId, err)
+				http.Error(w, err.Error(), http.StatusConflict)
```
phlogistonjohn: Should this be conflict or just an internal server error?

apps/glusterfs/app_volume.go
```diff
@@ -252,17 +252,22 @@ func (a *App) VolumeDelete(w http.ResponseWriter, r *http.Request) {
 			return nil
 		}
 
-		if volume.Info.BlockInfo.BlockVolumes == nil {
```
phlogistonjohn: Does the range below work as expected if the BlockVolumes value is nil? (I honestly don't remember)

apps/glusterfs/app_volume.go
obnoxxx: You are probably right, thanks! Will change.

apps/glusterfs/app_volume.go
obnoxxx: yes, I checked the following in the sandbox in golang.org ;-)
```
package main

import "fmt"

func main() {
	fmt.Println("Hello!")
	
	var mylist []string
	mylist = nil
	
	for _, myitem := range mylist {
		fmt.Println("myitem: %v", myitem)
	}
	
	fmt.Println("Bye...")
}
```

obnoxxx: it prints
```
Hello!
Bye...
```

phlogistonjohn: Great!

apps/glusterfs/app_volume.go
```diff
@@ -252,17 +252,22 @@ func (a *App) VolumeDelete(w http.ResponseWriter, r *http.Request) {
...
+		for _, bvId := range volume.Info.BlockInfo.BlockVolumes {
+			_, err = NewBlockVolumeEntryFromId(tx, bvId)
+			if err == nil {
+				err = logger.LogError("Cannot delete a block hosting volume containing block volumes")
+				http.Error(w, err.Error(), http.StatusConflict)
+				return err
+			}
+			if err != ErrNotFound {
+				err = logger.LogError("Refusing to delete block-hosting volume: "+
+					"Error loading block-volume [%v]: %v", bvId, err)
```
obnoxxx: @phlogistonjohn one small doubt I had is whether this overriding err this way actually works?! ... :-)

phlogistonjohn: I'm not sure it is useful, but I think it will behave as you expect.

obnoxxx:
@phlogistonjohn any additional thoughts? good to go?

PR #1222, opened 2018-06-13
Add support for resumable delete volume

### LINKED ISSUE #1220: Failed volume deletes should be resumable
### Kind of issue

Feature request

### Observed behavior

In current heketi master a volume delete can fail if some of the volume delete (and brick delete) steps fail, such as failing to unmount a brick. This leaves the volume in the heketi db.
This means the parts of the failed delete are not forgotten by heketi but we now have a non-functional non-deletable volume in the db. This can be removed by modifying the db and manually removing the components but this is an unpleasant process.

### Expected/desired behavior

Running 'heketi-cli volume delete' on a volume that is already partially deleted can succeed.

Manually remove a volume from gluster using the command line and then attempt to remove said volume from heketi.

phlogistonjohn:

This PR implements the ability to resume deleting partially deleted volumes. It currently involves refactoring various parts of the delete volume path. It also adds support for explicitly tracking the the LV and thinpool name for bricks. Functional tests for normal volumes are included.

### Does this PR fix issues?

Fixes #1220

### Notes for the reviewer

I still need to add tests for deleting cloned volumes. This is already a big PR so I may do that in a followup pr depending on timing/your opinions.

### DIFF
--- a/executors/cmdexec/brick.go
+++ b/executors/cmdexec/brick.go
@@ -144,6 +144,24 @@ func (s *CmdExecutor) deleteBrickLV(host, lv string) error {
 	return err
 }
 
+func (s *CmdExecutor) countThinLVsInPool(host, tp string) (int, error) {
+	// Detect the number of bricks using the thin-pool
+	commands := []string{
+		fmt.Sprintf("lvs --noheadings --options=thin_count %v", tp),
+	}
+	output, err := s.RemoteExecutor.RemoteCommandExecute(host, commands, 5)
+	if err != nil {
+		logger.Err(err)
+		return 0, fmt.Errorf("Unable to determine number of logical volumes in "+
+			"thin pool %v on host %v", tp, host)
+	}
+	thin_count, err := strconv.Atoi(strings.TrimSpace(output[0]))
+	if err != nil {
+		return 0, fmt.Errorf("Failed to convert number of logical volumes in thin pool %v on host %v: %v", tp, host, err)
+	}
+	return thin_count, nil
+}
+
 func (s *CmdExecutor) BrickDestroy(host string,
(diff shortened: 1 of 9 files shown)


executors/cmdexec/brick.go
```diff
@@ -142,13 +131,74 @@ func (s *CmdExecutor) BrickDestroy(host string,
...
 		fmt.Sprintf("umount %v", brick.Path),
 	}
 	_, umountErr = s.RemoteExecutor.RemoteCommandExecute(host, commands, 5)
 	if umountErr != nil {
-		logger.Err(err)
+		logger.Err(umountErr)
+		// check if the brick was previously unmounted
+		out, e := s.RemoteExecutor.RemoteCommandExecute(
+			host, []string{"mount"}, 5)
+		if e == nil && len(out) == 1 && !strings.Contains(out[0], brick.Path) {
```
obnoxxx: Why `len(out) == 1`? there could be more mounts.
Did you intend to use some form of `mount | grep ...` like in other places?

phlogistonjohn: The length of out in this case is the length of the slice passed to RemoteCommandExecute not the number of lines in the output of the mount command. The check is somewhat redundant but it is basically asserting that only the mount command was run by RemoteCommandExecute func. Since this is a test function and we have some control of the environment I didn't think I needed to be any fancier than that.

obnoxxx: Alright. I wasn't originally aware of the fact that we get each command's stdout in a single string in the output array. This is ok.

obnoxxx: good to not use pipes if we can avoid it!

PR #1223, opened 2018-06-14
client: disable issued at time claim in jwt

### LINKED ISSUE #646: Authentication failing with "Token used before issued"
We have a Kubernetes cluster on 3 Vagrant/VirtualBox nodes and we deployed Heketi to manage GlusterFS volumes provisioning.

All worked well, but when we activated authentication (setting users and secrets in heketi.json) heketi-cli often doesn't work, returning the error "Token used before issued".

We digged a bit into Heketi code and found that this error originates from file heketi/vendor/github.com/dgrijalva/jwt-go/claims.go
For what we can understand, at line 111:

    func verifyIat(iat int64, now int64, required bool) bool {
    	if iat == 0 {
    		return !required
    	}
    	return now >= iat
    }

the iat timestamp (that is, the client's time) is compared with the server timestamp. That is very easy to randomly fail if client and server run on different nodes. If we call heketi-cli from the same node where Heketi is running on, everything works fine.

Are we missing something? Maybe the time compare is by design and it requires the JWT client to issue calls in a controlled way?

Thanks for any feedback.

raghavendra-talur:
From https://github.com/dgrijalva/jwt-go/pull/139 it is understood
that if the machine where jwt token is generated and/or the machine
where jwt token is verified have any clock skew then there is a
possibility of getting a "Token used before issued" error. Considering
that we also check for expiration with delta of 5 minutes, disabling
iat claim until the patch is merged in jwt.

This PR disables the issued at time claim in jwt. Doing so will stop "token used before issued" errors.

### Does this PR fix issues?
Fixes #646

### DIFF
--- a/middleware/jwt.go
+++ b/middleware/jwt.go
@@ -22,8 +22,14 @@ import (
 	"github.com/gorilla/context"
 )
 
+// From https://github.com/dgrijalva/jwt-go/pull/139 it is understood
+// that if the machine where jwt token is generated and/or the machine
+// where jwt token is verified have any clock skew then there is a
+// possibility of getting a "Token used before issued" error. Considering
+// that we also check for expiration with delta of 5 minutes, disabling
+// iat claim until the patch is merged in jwt.
 var (
-	required_claims = []string{"iss", "iat", "exp"}
+	required_claims = []string{"iss", "exp"}
 )
 
 type JwtAuth struct {


phlogistonjohn:
Looks like the server may need a corresponding change. (Some of) the tests are failing due to the iat claim being missing. Perhaps we should make this configurable for backwards compatibility reasons?

phlogistonjohn:
retest this please

PR #1227, opened 2018-06-15
Set the "group gluster-block" options when manually creating block-hosting volumes

### LINKED ISSUE #1226: volume create request with block flag does not set the block-specific volume options
### Kind of issue

Bug

### Observed behavior

`heketi-cli volume create --size=X --block` does not set the `group gluster-block` volume options.
(Corresponding for the API request.)

### Expected/desired behavior

The cli or api request with the block flag should set the volume options.

Run `heketi-cli volume create --size=<X> --block`.

- Heketi version used (master): 
- Operating system used: fedora
- Heketi compiled from sources, as a package (rpm/deb), or container: source 
- Using kubernetes, openshift, or direct install: local install with mock translator

obnoxxx:
Fixes #1226

### DIFF
--- a/apps/glusterfs/volume_entry.go
+++ b/apps/glusterfs/volume_entry.go
@@ -93,12 +93,6 @@ func NewVolumeEntryFromRequest(req *api.VolumeCreateRequest) *VolumeEntry {
 	vol.Info.Size = req.Size
 	vol.Info.Block = req.Block
 
-	if vol.Info.Block {
-		vol.Info.BlockInfo.FreeSize = req.Size
-		vol.GlusterVolumeOptions = []string{"group gluster-block"}
-
-	}
-
 	// Set default durability values
 	durability := vol.Info.Durability.Type
 	switch {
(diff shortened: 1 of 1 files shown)


obnoxxx:
note: if you specify `group gluster-block` on the cmdline/in the request as well, this will have it listed twice. Should we make that unique?...

nixpanic:
I think this fails when the block hosting volume does not have replication enabled (durability=none). See kubevirt/kubevirt-ansible#260 for some details. The set of options for gluster-block is [here](https://github.com/gluster/glusterfs/blob/master/extras/group-gluster-block).

obnoxxx:
@nixpanic writes:
> I think this fails when the block hosting volume does not have replication enabled (durability=none). See kubevirt/kubevirt-ansible#260 for some details. The set of options for gluster-block is here.

Hmm, that's true, but actually unrelated to the fact that this setting is not applied when manually creating block-hosting volumes.

Group options should probably be made more intelligent in gluster... :-)
(or the options that do not apply should be made no-ops if not applicable, or at the very least we need different flavors of group options...)

nixpanic:
On Fri, Jun 29, 2018 at 12:17:51AM -0700, Michael Adam wrote:
> 
> @phlogistonjohn writes:
> > @obnoxxx writes:
> > >  note: if you specify group gluster-block on the cmdline/in the
> > >  request as well, this will have it listed twice. Should we make
> > >  that unique?...
> > 
> > I think so. My preferred approach would to be to only set the group
> > to gluster-block if vol.Info.Block is true and there are no other
> > groups specified in the options list.
> 
> I kind of disagree: I don't think we should not set it at all. It
> should be up to gluster to resolve conflicts for various group
> options. They are not all mutually exclusive either. I propose to
> **prepend** the block option to the list of provided volume options in
> this patch, instead of **appending** it, so that user-provided options
> can take precedence? -- OK?

That makes sense to me.

phlogistonjohn:
> I kind of disagree: I don't think we should not set it at all. It should be up to gluster to resolve conflicts for various group options. They are not all mutually exclusive either. I propose to prepend the block option to the list of provided volume options in this patch, instead of appending it, so that user-provided options can take precedence? -- OK?

If that's how gluster works, then OK. Do you need to make any other changes to the patches or do you want a re-review given the current consensus?

obnoxxx:
> If that's how gluster works, then OK. Do you need to make any other changes to the patches or do you want a re-review given the current consensus?

Just the change to **prepend** instead of **append**. Will update shortly.

PR #1230, opened 2018-06-18
functional tests: support custom methods for running gluster

phlogistonjohn:

This change slightly restructures the lib.sh functions and allows
for environment vars to be used to provide custom methods for
starting and stopping the gluster nodes.

The idea here is to move towards abstracting away the use of
vagrant and making it easy to try different approaches to
running the gluster test nodes.

### Notes for the reviewer

This is a starting off point for some experimentation I am doing where the test is run within a vm, and thus running vms inside that vm is undesirable. I haven't used it much yet, so you can view this as quasi-wip.

### DIFF
--- a/tests/functional/lib.sh
+++ b/tests/functional/lib.sh
@@ -67,16 +67,39 @@ stop_heketi() {
     done
 }
 
+start_gluster_nodes() {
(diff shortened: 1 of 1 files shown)


phlogistonjohn:
This PR was obsoleted by PR #1612  Closing.

PR #1231, opened 2018-06-19
Update the free size in block info on expansion

### LINKED ISSUE #1229: free size in block hosting volume info isn't updated on expansion
### Kind of issue

Bug

### Observed behavior
When a block hosting volume is expanded then the free size isn't updated.

### Expected/desired behavior
Free size should get updated

create a block volume
block hosting volume should get auto created
now expand the block hosting volume 
try creating a block volume which would cross the original limit of block hosting volume

heketi 7

raghavendra-talur:
Update the free size in block info on expansion. If a volume is expanded and has a block flag set to true, then we should also update the free size attribute of the volume in blockinfo.

### Does this PR fix issues?

Fixes #1229

### DIFF
--- a/apps/glusterfs/operations.go
+++ b/apps/glusterfs/operations.go
@@ -341,6 +341,9 @@ func (ve *VolumeExpandOperation) Finalize() error {
 			}
 		}
 		ve.vol.Info.Size += sizeDelta
+		if ve.vol.Info.Block == true {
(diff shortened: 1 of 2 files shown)


phlogistonjohn:
retest this please

phlogistonjohn:
retest this please

PR #1235, opened 2018-06-21
Add adminstrative modes to Heketi server

### LINKED ISSUE #1225: Support administrative modes
### Kind of issue

Feature request

### Observed behavior

When trying to debug heketi issues or when performing important administrative tasks it can be useful to control the types of requests heketi will serve. Similarly, this could be used to allow the system to settle before stopping the server (and avoid leaving pending operations on the plate).

### Expected/desired behavior

Add an api that allows an admin to set an administrative mode. Modes could include:
* Normal
* Read-Only - only GETs will be served
* Limit-Client - only calls from a specific client (IP) will be served

At the very least, something like the former would be very useful to quiesce the server.

phlogistonjohn:

This series creates apis, middleware and utilities to track and control a heketi
server's administrative state. Currently all the state does
is control what clients are allowed to made modifying requests (POSTS, DELETES, etc):
* normal: all clients
* read-only: no clients
* local-only: only localhost may make changes

The idea is to allow an admin to put heketi into read-only mode in order for it to finish outstanding requests but accept no more. The local-only state is a hybrid state where only changes from localhost will be accepted so as to drain any outstanding external client requests but to allow for local administrative actions like a node replacement.

As a special case a local process can send SIGUSR2 to the server to reset the administrative state to normal. This is done to support testing (TDB) as well as an "escape hatch" in case the user sets the server to read-only and decides that was not what was wanted.

### Does this PR fix issues?

Fixes #1225

### Notes for the reviewer

Currently only the server core and the API are supported. I have not done cli yet and would like feedback on the general approach first. When it comes time to do the cli I'm actually thinking that the "status" version will hit this api as well as a few others to aggregate some general information about the state of the server.

### DIFF
--- a/server/admin/reset.go
+++ b/server/admin/reset.go
@@ -1,0 +1,31 @@
+//
+// Copyright (c) 2018 The heketi Authors
+//
+// This file is licensed to you under your choice of the GNU Lesser
+// General Public License, version 3 or any later version (LGPLv3 or
+// later), or the GNU General Public License, version 2 (GPLv2), in all
+// cases as published by the Free Software Foundation.
+//
+
+package admin
+
+import (
+	"os"
+	"os/signal"
+
+	"github.com/heketi/heketi/pkg/glusterfs/api"
+)
+
+func ResetStateOnSignal(s *ServerState, sig ...os.Signal) {
+	signalch := make(chan os.Signal, 1)
+	signal.Notify(signalch, sig...)
+
+	go func() {
+		for {
+			select {
+			case <-signalch:
+				s.Set(api.AdminStateNormal)
+			}
+		}
+	}()
+}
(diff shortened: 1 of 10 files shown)


server/admin/reset.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"os"
+	"os/signal"
+
+	"github.com/heketi/heketi/pkg/glusterfs/api"
+)
+
+func ResetStateOnSignal(s *ServerState, sig ...os.Signal) {
+	signalch := make(chan os.Signal, 1)
+	signal.Notify(signalch, sig...)
```
raghavendra-talur: we need to specify signal USR2 here?

phlogistonjohn:
Added patch for heketi-cli commands.

phlogistonjohn:
NOTE: Travis networking seems to be flaky ATM and is causing package fetch failures. I am restarting some of the failed tasks.

PR #1238, opened 2018-06-26
build: add a check for mercurial/hg

nixpanic:

Glide needs to fetch a repository from bitbucket.org, which is a Mercurial hosting site. The `hg` executable needs to be available to download the requires pkgs.

If `hg` is not available running `make vendor` fails with the following error:

     [WARN] Unable to checkout bitbucket.org/ww/goautoneg
     [ERROR] Update failed for bitbucket.org/ww/goautoneg: hg is not installed
     [ERROR] Failed to install: hg is not installed
     make: *** [Makefile:74: vendor] Error 1

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,7 @@ ifeq ($(GOARCH),$(GOHOSTARCH))
 endif
 endif
 GLIDEPATH := $(shell command -v glide 2> /dev/null)
+HGPATH := $(shell command -v hg 2> /dev/null)
 DIR=.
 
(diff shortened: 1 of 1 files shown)


nixpanic:
@phlogistonjohn let me know if you want me to correct the commit message. Otherwise I assume it's ok to call bitbucket.org a Mercurial hosting site.

phlogistonjohn:
@nixpanic that was just a nitpick. You don't need to change it unless you want to.

Issue #1242, opened 2018-06-29
Robust delete fails when thin pool is fully deleted


phlogistonjohn:
### Kind of issue

Bug

### Observed behavior

Heketi is requested to delete a partially deleted volume (volume is deleted from gluster, 2 of 3 bricks fully deleted). It returns an error reporting that it is "Unable to determine number of logical volumes in thin pool"

```
$ heketi-cli volume delete 1e71bc93d8e1a6719bcee7b657b088f1
Error: Unable to determine number of logical volumes in thin pool vg_c8815688fd8d1b77b03589c6044daa39/tp_52839b42fb59d4f1c9da920cd19c4d74 on host node2
command terminated with exit code 255
```

### Expected/desired behavior

Deleting this volume should succeed removing all remaining brick lvs, etc.

### Details on how to reproduce (minimal and precise)

* Prevent umount of one brick by keeping a file open on one brick fs.
* Run heketi volume delete <volid>
* Close the open file on the brick fs (allowing future umount)
* Run heketi volume delete <volid> again

### Information about the environment:

- Heketi version used (e.g. v6.0.0 or master): master

### Other useful information

This problem is to to an incorrect refactoring that moved a "user friendly" error message into a utility function, but the check for the "lv missing" versus "other error" expects to work off of the raw error from lvm.

PR #1243, opened 2018-06-29
executors: delete path needs to check raw errors from lvm commands

phlogistonjohn:
When this function was refactored in preparation for robust delete
behavior the "friendly" error message was moved into the utility
function used for counting the number of lvs in a thin pool.
However, the errors from this function need to be checked to see
if lvm errors out because the thin pool is missing vs. another
reason. Changing the raw lvm error to a friendly error in this
case is wrong. Move the friendly error back to the upper function
call.

Fixes issue #1242


raghavendra-talur:
@centos-ci retest this please

Issue #1001, opened 2018-02-05
Device resync shows increased disk size when disk hasn't been expanded


aliscott:
Heketi became out of sync with my gluster volumes, so I ran a device resync. Before running device resync `heketi-cli node info <NODE ID>` showed:
```
State:online    Size (GiB):250     Used (GiB):234     Free (GiB):16
```

I then ran:
```
heketi-cli device resync <DEVICE ID>
```

and now `heketi-cli node info <NODE ID>` shows:
```
State:online    Size (GiB):255     Used (GiB):234     Free (GiB):21
```

But the total size of the disk has not changed. `pvs` shows:
```
/dev/vdc   vg_6e355a7fcbc4a3e1c60bedc3ffd6ac81 lvm2 a--  249.87g  21.12g
```

raghavendra-talur:
@aliscott thanks for highlighting this. I hit the same issue too, will start working on this.

dlamotte:
https://github.com/heketi/heketi/commit/6ee7ab4e3b4ea55d9967dfafaf2847b47ac2dadc#diff-c4be80418af013b96800ee9a5485e022L386 appears to be related here.  We've seen this issue in our environment as well but had some issues reproducing it specifically.

rouja:
Hi,

I hit the same issue too.

heketi-cli 6.0.0
glusterfs 3.8.4

raghavendra-talur:
Fixed by #1245

PR #1245, opened 2018-06-29
device resync should update total, used and free

raghavendra-talur:
Device resync feature currently only fetches the free space and calculates the total and used based on that and data in heketi db. If the device has been resized or if heketi has gone out of sync from back end it is better to update all three attributes.


raghavendra-talur:
@phlogistonjohn no, that is all I have for now.

raghavendra-talur:
Fixes #1001

PR #1247, opened 2018-07-02
executors: prevent (vg|pv)remove commands from prompting

### LINKED ISSUE #1244: vgremove command can "get stuck" waiting for prompt
### Kind of issue

Bug

### Observed behavior

Running a heketi command to add a device never completed without additional intervention.
Example:
```
oc exec heketi-7d88c9cd74-pp956 -- heketi-cli device add --node=87b0a2168e3c1d6ef1ce4ad5e6dc930b  --name=/dev/bob
```

### Expected/desired behavior

Command exits with success or failure within a "reasonable" time frame, regardless of the state of the system. Something like
```
$ oc exec heketi-7d88c9cd74-pp956 -- heketi-cli device add --node=87b0a2168e3c1d6ef1ce4ad5e6dc930b  --name=/dev/bob
Error: Unable to execute command on glusterfs-rzlk8:   Device /dev/bob not found.
command terminated with exit code 255
```

The key to this error is getting lvm tools in a bad state. Thus it is somewhat difficult to give minimal steps. While trying to reproduce a different scenario I:

* Manually opened a file on a brick fs so that umount commands would fail
* Did a node replacement, migrating bricks off of the brick fs in question
* Used heketi-cli to delete the device which heketi allowed because it considered the device devoid of bricks  (see below)
* Ran heketi-cli device add with a dummy device name on the same node as the "bad" brick

Looking into the state of the system I can see that multiple lvm commands are running:
* pvcreate /dev/bob (x2, because I ran the command multiple times)
* vgremove vg_b934a3f13f3dca5601931b346fa9c42c

strace revealed that pvcreate is blocked on flock of an lvm lock file. strace of vgremoved revealed that it was reading from stdin. Running the same command manually prompts like:
```
# vgremove  vg_b934a3f13f3dca5601931b346fa9c42c
Do you really want to remove volume group "vg_b934a3f13f3dca5601931b346fa9c42c" containing 2 logical volumes? [y/n]: n
  Volume group "vg_b934a3f13f3dca5601931b346fa9c42c" not removed
```

- Heketi version used (e.g. v6.0.0 or master): master

### Other useful information

We should take care to make sure we run our lvm commands in non-prompting mode. I think we can simply change the code in this case to `vgremove -qq` so that vgremove will not prompt.

phlogistonjohn:
If there is an unexpected condition on the vg or pv when
removing a device it is preferable to fail than to allow
the command to prompt for input which will block heketi
"forever" and possibly (due to lvm-level) locking other
commands as well.

Fixes #1244

### Notes for the reviewer

This ought to have a functional test but due to time limits and the fiddly nature (having to do lvm stuff directly via the test code) I skipped this for now. It may be worthwhile to do a general technical debt issue for adding tests in this general area.

### DIFF
--- a/executors/cmdexec/device.go
+++ b/executors/cmdexec/device.go
@@ -73,8 +73,8 @@ func (s *CmdExecutor) DeviceTeardown(host, device, vgid string) error {
 
 	// Setup commands
 	commands := []string{
-		fmt.Sprintf("vgremove %v", utils.VgIdToName(vgid)),
-		fmt.Sprintf("pvremove '%v'", device),
+		fmt.Sprintf("vgremove -qq %v", utils.VgIdToName(vgid)),
+		fmt.Sprintf("pvremove -qq '%v'", device),
 	}
 
 	// Execute command


mykaul:
Two issues I see with this:
1. It will fail the command, and this is probably not what we needed to do (yes, we might have a bug somewhere if it popped the interactive question in the 1st place?)
2. It will suppress output and log messages - so we won't know what happened?

FWIW, I think other implementation use '-f' when calling the commands ('vgremove' and 'pvremove').
Example: https://github.com/oVirt/vdsm/blob/4614a324da0edadfa55a51c95c35d2a641a8acc4/lib/vdsm/storage/lvm.py#L997

phlogistonjohn:
> Two issues I see with this:

>  1.  It will fail the command, and this is probably not what we needed to do (yes, we might have a bug somewhere if it popped the interactive question in the 1st place?)

I think it is more correct to fail in this situation. The failure will indicate some precondition was not met and prevent the command from possibly making the situation worse.

>  2.  It will suppress output and log messages - so we won't know what happened?

I was worried about that after reading the manpage but when I tried it live it did in fact print the message that it prints when prompting and subsequent error message. I was prepared to redirect input from /dev/null in case -qq did suppress error message. Since it did not suppress errors, it seemed cleaner.

FWIW, I think other implementation use '-f' when calling the commands ('vgremove' and 'pvremove').
Example: https://github.com/oVirt/vdsm/blob/4614a324da0edadfa55a51c95c35d2a641a8acc4/lib/vdsm/storage/lvm.py#L997

mykaul:
@raghavendra-talur - yes, the reply makes sense to me (although as opposed to 'pvremove' and friends I feel it's not as harmful).

raghavendra-talur:
The second centos-ci failure was because of virt error in ci-system. As the first run had passed, I merged the PR.

PR #1251, opened 2018-07-04
[RFC] Remove orphan bricks in DeleteBricksWithEmptyPath

obnoxxx:

If there's a brick id on a device for which the brick entry does not exist (any more), the delete-bricks-with-empty-path command would previously fail. A previous commit has made delete-bricks-with-empty-path more robust in ignoring these brick ids. This patch goes a step further in cleaning them up while we're at it.

This is more an RFC than a request for merging.

### The conceptual question is: 

Do we want to do this kind of clean up of broken link structures
* at all?
* where we come across it?
* in a dedicated function that does some kind of fsck for the db?

### Notes for the reviewer

This is based on PR #1249.

### DIFF
--- a/apps/glusterfs/device_entry.go
+++ b/apps/glusterfs/device_entry.go
@@ -573,6 +573,7 @@ func markDeviceFailed(db wdb.DB, id string, force bool) error {
 func (d *DeviceEntry) DeleteBricksWithEmptyPath(tx *bolt.Tx) error {
 	godbc.Require(tx != nil)
 	var bricksToDelete []*BrickEntry
+	var bricksToRemoveFromDevice []string
 
 	logger.Debug("Deleting bricks with empty path on device [%v].",
 		d.Info.Id)
@@ -580,8 +581,8 @@ func (d *DeviceEntry) DeleteBricksWithEmptyPath(tx *bolt.Tx) error {
 	for _, id := range d.Bricks {
 		brick, err := NewBrickEntryFromId(tx, id)
 		if err == ErrNotFound {
-			logger.Warning("Ignoring nonexistent brick [%v] on "+
-				"disk [%v].", id, d.Info.Id)
+			bricksToRemoveFromDevice =
+				append(bricksToRemoveFromDevice, id)
 			continue
 		}
 		if err != nil {
@@ -593,10 +594,22 @@ func (d *DeviceEntry) DeleteBricksWithEmptyPath(tx *bolt.Tx) error {
 			bricksToDelete = append(bricksToDelete, brick)
 		}
 	}
+
+	for _, id := range bricksToRemoveFromDevice {
+		logger.Debug("Removing unreferenced brick [%v] from device [%v].",
+			id, d.Info.Id)
+		d.BrickDelete(id)
+	}
+	err := d.Save(tx)
+	if err != nil {
+		logger.LogError("Unable to save device %v: %v", d.Info.Id, err)
+		return err
+	}
+
 	for _, brick := range bricksToDelete {
 		logger.Debug("Deleting brick [%v] which has empty path.",
 			brick.Info.Id)
-		err := brick.Delete(tx)
+		err = brick.Delete(tx)
 		if err != nil {
 			return logger.LogError("Unable to remove brick %v: %v", brick.Info.Id, err)
 		}


apps/glusterfs/device_entry.go
```diff
@@ -573,30 +573,40 @@ func markDeviceFailed(db wdb.DB, id string, force bool) error {
...
-		if err != nil {
+			bricksToRemoveFromDevice =
+				append(bricksToRemoveFromDevice, id)
+		} else if err != nil {
 			logger.LogError("Unable to fetch brick [%v] from db: %v",
 				id, err)
 			return err
-		}
-		if brick.Info.Path == "" {
+		} else if brick.Info.Path == "" {
```
raghavendra-talur: @obnoxxx ,

We can add the brick to "bricksToRemoveFromDevice" too.

apps/glusterfs/device_entry.go
```diff
@@ -573,30 +573,40 @@ func markDeviceFailed(db wdb.DB, id string, force bool) error {
...
+		logger.Debug("Removing unreferenced brick [%v] from device [%v].",
+			id, d.Info.Id)
+		d.BrickDelete(id)
+	}
+	err := d.Save(tx)
+	if err != nil {
+		logger.LogError("Unable to save device %v: %v", d.Info.Id, err)
+		return err
+	}
+
```
raghavendra-talur: Move this section below the bricksToDelete section. That way we won't leave any orphans bricks again.

If you agree with the approach but don't have time to, I would be happy to do the update.

phlogistonjohn: I talked this over with @raghavendra-talur  and we both agree that changes need to be made here so that we don't delete the brick entries but leave the references on the device. One addtional suggestion I have is to avoid looping twice and instead create a new list of "good bricks" as you go. Then after the end of the loop assign the new good-brick list to d.Bricks.

obnoxxx: @raghavendra-talur writes:
> Move this section below the bricksToDelete section. That way we won't leave any orphans bricks again.

I don't fully understand yet: How would we possibly end up leaving (new) orphan bricks again here, due to this brick from device removal being above the loop to remove the bricks without path? That loop also removes the removed bricks-with-empty-path from the referencing device.

obnoxxx: @phlogistonjohn writes:
>  I talked this over with @raghavendra-talur and we both agree that changes need to be made here so that we don't delete the brick entries but leave the references on the device.

I don't get the problem yet: In the loop over the collected bricks with empty path, we do

```
   d.BrickDelete(brick.Info.Id)
   d.Save(tx)
   ...
```
so that leaves no new references to nonexisting bricks.

>  One additional suggestion I have is to avoid looping twice and instead create a new list of "good bricks" as you go. Then after the end of the loop assign the new good-brick list to d.Bricks.

I am giving it a try, but I have to say that I find the current proposal much more natural: It is removing the brick ids that reference a nonexisting brick in one loop and then removes the brick-ids that reference bricks with empty path when it actually removes the brick entry itself. That seems like a natural grouping to me even though the multiple loops may not be highly optimized. But I am not sure that optimization is very important here, since this should not need to be a highly performant operation.

If I prepare the complete new list of bricks for the device in the first loop and just assign it over the device's brick list afterwards, I have to work a little harder in the first loop (because of the sorted string list type), and spare the loop over the (few?) brick IDs that don't reference an existing brick. The end result will be exactly the same due to all this being wrapped in a transaction.

I will do the corresponding change and look at the result, but the assignment of the device's brick list instead of operating on it makes me feel slightly uneasy.

obnoxxx:
for now, I rebased the PR on master so we see the current state. Will work on the proposed changes as discussed...

phlogistonjohn:
> I don't get the problem yet: In the loop over the collected bricks with empty path, we do

 ```
   d.BrickDelete(brick.Info.Id)
   d.Save(tx)
   ...
```

I see that now. I think somehow when I was looking at the diff I missed this part and assumed it was not there. Sorry.

I think what this really needs is some tests to demonstrate that it behaves as intended and we should be OK to merge then.

obnoxxx:
first rebased on master (and reduced the diff slightly). will do tests next

phlogistonjohn:
PR has been stale for over a year. Closing. Please feel free to reopen if needed.

PR #1256, opened 2018-07-11
Number of hosts should be equal to ha count 

raghavendra-talur:
Earlier, due to off by one error, we added one host more than ha count in list of hosts provided to gluster-block cli during blockvolume create.

This PR matches number of hosts to ha count parameter.

### DIFF
--- a/apps/glusterfs/app_block_volume_test.go
+++ b/apps/glusterfs/app_block_volume_test.go
@@ -277,6 +277,72 @@ func TestBlockVolumeCreate(t *testing.T) {
 	tests.Assert(t, info.Auth == false)
 }
 
+func TestBlockVolumeCreateHACount(t *testing.T) {
+	tmpfile := tests.Tempfile()
+	defer os.Remove(tmpfile)
+
+	// Create the app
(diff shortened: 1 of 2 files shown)


apps/glusterfs/app_block_volume_test.go
```diff
@@ -277,6 +277,72 @@ func TestBlockVolumeCreate(t *testing.T) {
...
+			tests.Assert(t, r.Header.Get("Content-Type") == "application/json; charset=UTF-8")
+			err = utils.GetJsonFromResponse(r, &info)
+			tests.Assert(t, err == nil)
+			break
+		}
+	}
+	tests.Assert(t, info.Id != "")
+	tests.Assert(t, info.Cluster != "")
+	tests.Assert(t, info.BlockHostingVolume != "")
+	tests.Assert(t, len(info.BlockVolume.Hosts) == 3)
```
phlogistonjohn: Am I correct to assume that this line fails the test if the other patch is not applied?

raghavendra-talur: yes

PR #1267, opened 2018-07-14
[RFC] Operations based throttling

phlogistonjohn:

This series adds http request throttling based on the number of operations that the server is currently processing (in-flight operations). The server will accept up to N operations that it will process at once but after that limit is reached will respond to clients with the 429 (too many requests) error code.
This limit is configurable both through the config json and an env variable.

### Notes for the reviewer

This PR is an alternative to PR #1131 which adds a generic throttling middleware to heketi. Instead of that approach, this PR aims at a more minimal set of changes that only covers the actions heketi performs that are very costly - mainly an operation's exec phase.

This PR depends on PR #1266, primarily for hand testing by running heketi-cli commands. This PR is not complete and requires tests. While that work is in progress _please_ take a look at the general approach and see if you think this is heading in the right direction or has gone off into the weeds. :-)


humblec:
@phlogistonjohn this indeed looks a better approach to bring throtlling!! and it looks like a must thing to take care of many issues we are currently encountering in this area. The env or config setup also  looks promising as we can tweak it based on the setup. so complete ack from my side and I would let others to review this in detail.

apps/glusterfs/app.go
```diff
@@ -289,6 +300,16 @@ func (a *App) setFromEnvironmentalVariable() {
 			a.conf.KubeConfig.RebalanceOnExpansion = value
 		}
 	}
+
+	env = os.Getenv("HEKETI_MAX_LIVE_OPERATIONS")
```
humblec: Not sure whether `LIVE` or `CONCURRENT` sound better here?

apps/glusterfs/operations_manage.go
```diff
@@ -0,0 +1,209 @@
...
+	"fmt"
+	"net/http"
+	"sync"
+
+	"github.com/heketi/heketi/executors"
+)
+
+// OpCounter is used to track and manage how many operations are being
+// processed by the server.
+type OpCounter struct {
```
humblec: Couple of questions on locking: The big lock is taken for each operation or start of each request , Isnt it ? second question is, how the lock is handled at server restart or respawn?

apps/glusterfs/app.go
nixpanic: How about `OUTSTANDING` or `INFLIGHT`?

phlogistonjohn: I like INFLIGHT because it matches how I speak about it. I don't want to use the word concurrent because we don't promise anything about concurrency in regards to this behavior.

apps/glusterfs/operations_manage.go
phlogistonjohn: A read lock is taken to check the current number of in-flight operations but is released when the function returns. Similarly, write locks are taken when the increment or decrement function is called but only long enough to change the counter value. The lock is not held the majority of the time when the operation is being processed. The lock is always unlocked when the struct is initialized, this is the default of the golang lock construct.

phlogistonjohn:
retest this please

humblec:
retest this please

apps/glusterfs/app.go
humblec: `INFLIGHT` looks more close to this.

PR #1268, opened 2018-07-14
extras: add a script to build heketi containers using buildah

phlogistonjohn:
I have been successfully using this script for a while to
build containers for test purposes. It has the ability to
build heketi from source (similar to extras/docker/fromsource)
but as an added bonus has an option for using using existing
heketi binaries and can thus cut down on the time needed to
build a heketi container.
All modes are based on the 'buildah' tool which is a daemon-less
alternative to 'docker build'.


phlogistonjohn:
This is a simple(ish) standalone script. Any opinions?

extras/buildah/build.py
```diff
@@ -0,0 +1,113 @@
...
+    heketi_cli_bin = pjoin(heketi_dir, 'client/cli/go/heketi-cli')
+    shutil.copy(heketi_bin, outdir)
+    shutil.copy(heketi_cli_bin, outdir)
+    buildah('umount', bc)
+
+    buildah('rm', bc)
+
+
+def build_container(sources):
+    hc = buildah_from('fedora:27')
```
nixpanic: This could probably be `fedora:latest`?

extras/buildah/build.py
```diff
@@ -0,0 +1,113 @@
...
+    return buildah('from', src, redirect=True)
+
+
+def build_binaries(outdir):
+    pkgs = 'glide golang git make mercurial'
+    hdir = '/build/src/github.com/heketi/heketi'
+    heketi_branch = 'master'
+    heketi_url = 'https://github.com/heketi/heketi.git'
+
+    bc = buildah_from('fedora:27')
```
nixpanic: I assume `fedora:27` to get a functional version of glide? A comment and maybe reference to the BZ would be  good.

extras/buildah/build.py
phlogistonjohn: There was an issue using fedora 28 IIRC. What I should do is move that to a top-level variable and make it possible to swap the base image easier for testing.

extras/buildah/build.py
phlogistonjohn: Yes, good idea.

phlogistonjohn:
Updated.

PR #1277, opened 2018-07-19
update test binaries on make

raghavendra-talur:
When heketi and heketi-cli binaries are built they are not copied to locations where test scripts start them. The test scripts have a check with bash "-x" to see if the binary exists and if it does they don't update from the build location.

Considering that the developers might want to have control on what version of binary to use while testing, we don't want to always copy over the binary in the bootstrapping part of the test.

Rather, we overwrite the binary when make is run.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -96,7 +96,10 @@ client: vendor glide.lock
 run: server
 	./$(APP_NAME)
 
-test: vendor glide.lock
+test: server client vendor glide.lock
+	cp heketi tests/functional/TestDbExportImport/heketi-server
+	cp heketi client/api/python/heketi-server
+	cp client/cli/go/heketi-cli tests/functional/TestDbExportImport/heketi-cli
 	$(TESTBIN) $(TESTOPTIONS)
 
 test-functional: vendor glide.lock
@@ -105,6 +108,9 @@ test-functional: vendor glide.lock
 clean:
 	@echo Cleaning Workspace...
 	rm -rf $(APP_NAME)
+	rm -rf tests/functional/TestDbExportImport/heketi-server
+	rm -rf client/api/python/heketi-server
+	rm -rf tests/functional/TestDbExportImport/heketi-cli
 	rm -rf dist coverage packagecover.out
 	@$(MAKE) -C client/cli/go clean
 


raghavendra-talur:
@phlogistonjohn @obnoxxx @nixpanic @Madhu-1 

This is not the only way to fix the issue. Suggestions welcome to do it in a better way.

Makefile
```diff
@@ -89,7 +89,10 @@ client: vendor glide.lock
 run: server
 	./$(APP_NAME)
 
-test: vendor glide.lock
+test: server client vendor glide.lock
+	cp heketi tests/functional/TestDbExportImport/heketi-server
+	cp heketi client/api/python/heketi-server
+	cp client/cli/go/heketi-cli tests/functional/TestDbExportImport/heketi-cli
```
nixpanic: This should be fine, there is no `$(CP)` variable to use instead of `cp`.

Makefile
```diff
@@ -98,6 +101,9 @@ test-functional: vendor glide.lock
 clean:
 	@echo Cleaning Workspace...
 	rm -rf $(APP_NAME)
+	rm -rf tests/functional/TestDbExportImport/heketi-server
+	rm -rf client/api/python/heketi-server
+	rm -rf tests/functional/TestDbExportImport/heketi-cli
```
nixpanic: Ideally `$(RM)` is used, but that by default expands to `rm -rf` too. You actually should not need `-r`, but it is not a problem either.

Makefile
```diff
@@ -89,7 +89,10 @@ client: vendor glide.lock
 run: server
 	./$(APP_NAME)
 
-test: vendor glide.lock
+test: server client vendor glide.lock
+	cp heketi tests/functional/TestDbExportImport/heketi-server
```
obnoxxx: wouldn't it be better to have this under the `test-functional` target?

Makefile
obnoxxx: Also this line should be under `test-functional` imho.

Makefile
```diff
@@ -89,7 +89,10 @@ client: vendor glide.lock
...
+	cp heketi client/api/python/heketi-server
+	cp client/cli/go/heketi-cli tests/functional/TestDbExportImport/heketi-cli
 	./test.sh $(TESTOPTIONS)
 
 test-functional: vendor glide.lock
```
obnoxxx: test-functional should also depend on `server` and `client`

obnoxxx:
@raghavendra-talur ping. Any updates?

Also I think this needs more discussion, I think @phlogistonjohn had some concerns about the general approach here...

obnoxxx:
@raghavendra-talur ping

phlogistonjohn:
can we close this couple of years old PR?

PR #1278, opened 2018-07-19
apps: avoid throttling test code much

phlogistonjohn:
Some of the tests make a lot of requests at once and throttling has
a huge impact on it (1min vs 4min). So we avoid doing much
throttling at all in the test code.

### DIFF
--- a/apps/glusterfs/testapp_mock.go
+++ b/apps/glusterfs/testapp_mock.go
@@ -20,6 +20,7 @@ func NewTestApp(dbfile string) *App {
 		DBfile:                    dbfile,
 		Executor:                  "mock",
 		CreateBlockHostingVolumes: true,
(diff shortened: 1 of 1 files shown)


raghavendra-talur:
Adding some numbers here for reference:

We found that TestTopology and TestClientVolume functions took more than 130 seconds each in 100-gotest.sh runs. On carefully analyzing, we found that having a go func() in loop of 50 slows down the operations if limit on inflight operations placed in throttling code is less than 50.

In order to keep the unit tests fast like they should be, the test framework now starts with limit of 64. Throttling tests bring it down to 5 or 8 for testing throttling itself.

As as aside, we also discovered the GOCACHE env. If set, it caches build and test results. Set the env to "off" if you don't want to rely on the cache. The cache is not used if go test is used with a package path, that is another way of avoiding cache.

PR #1280, opened 2018-07-24
apps: only consider available hosts for ha count

ansiwen:

Before the code iterated only over ha count hosts, no matter if they
were available or not. This change checks for the number of available
hosts being equal to ha count instead, before stopping iteration.


apps/glusterfs/block_volume_entry_create.go
```diff
@@ -63,14 +63,18 @@ func (v *BlockVolumeEntry) createBlockVolumeRequest(db wdb.RODB,
 		}
 
 		if v.Info.Hacount > 0 && v.Info.Hacount <= len(bhvol.Info.Mount.GlusterFS.Hosts) {
-			for i := 0; i < v.Info.Hacount && i < len(bhvol.Info.Mount.GlusterFS.Hosts); i++ {
+			v.Info.BlockVolume.Hosts = nil
```
phlogistonjohn: I can't quite seem to wrap my head around this line. What is it meant to accomplish?

ansiwen: Clear the slice, just in case it is not empty, which is what we expect at this point I assume?

phlogistonjohn: OK, it seems kind of unnecessary to me then, but I won't reject the change because of it.

obnoxxx: Yeah, this is always a question if this kind of initialization is needed, but it should not harm. The change looks correct to me.

obnoxxx:
ok to test

apps/glusterfs/block_volume_entry_create.go
ansiwen: Well, I'm a bit surprised, because I even considered this as a bug, that it is not cleared. :grin: (And would have rejected it in a code review without it.) Only if we know by code-path that it must be empty I would consider it unnecessary, because the code clearly assumes that. (Of course, if I'm in an init function that is only called by constructors, then I would also skip the initialization, because the function _is_ the initializer, but I couldn't see that here.) Alternatively, if we assume that by convention I would still put an assert there. Isn't "better be safe than sorry" a general rule? These things can cause naughty bugs and it's very cheap to clean it. Of course, my caution is certainly also caused by the fact that I'm not familiar with the code yet.
That I use `= nil` instead of `= [:0]` is just because I googled how to clear slices, and StackOverflow taught me about it: https://stackoverflow.com/a/16973160

ansiwen:
I added two tests now, but I'm a bit puzzled, that the http server returns StatusInternalServerError (500) when it runs out of available hosts. Is it really meant to be like that? (I would expect a 4xx error)

phlogistonjohn:
retest this please

phlogistonjohn:
> I added two tests now, but I'm a bit puzzled, that the http server returns StatusInternalServerError (500) when it runs out of available hosts. Is it really meant to be like that? (I would expect a 4xx error)

That sounds pretty typical for Heketi. We've been slowly adding more meaningful error codes but most error conditions end up as 500s and you have to read the body to get any meaningful data (or worse the logs)

PR #1281, opened 2018-07-24
apps: shuffle block-volume ha nodes to avoid relying on same nodes 

phlogistonjohn:

When the number of nodes available is greater than the HA count
we want to select a subset of these nodes for use with gluster
block ha support. Previous changes to the code try to select
the ha-count number of healthy nodes from this set but if the
healthy node set was unchanging it would always pick the exact
same nodes.

This change add utility code for shuffling (based on Go 1.10), shuffles the list of nodes that is tested for
liveness so different block volumes are distributed across
different nodes, and adds a test for this behavior.

### Does this PR fix issues?

Fixes (latter half of) rhbz#1595531

### Notes for the reviewer

This change depends on PR #1280 and should not be merged before it.

### DIFF
--- a/apps/glusterfs/app_block_volume_test.go
+++ b/apps/glusterfs/app_block_volume_test.go
@@ -17,6 +17,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"os"
+	"reflect"
 	"strings"
 	"testing"
 	"time"
@@ -278,6 +279,31 @@ func TestBlockVolumeCreate(t *testing.T) {
 	tests.Assert(t, info.Auth == false)
 }
 
+func blockVolumeTestResult(t *testing.T, r *http.Response) api.BlockVolumeInfoResponse {
+	var info api.BlockVolumeInfoResponse
+	tests.Assert(t, r.StatusCode == http.StatusAccepted)
+	location, err := r.Location()
+	tests.Assert(t, err == nil)
+
+	// Query queue until finished
+	for {
+		r, err = http.Get(location.String())
+		tests.Assert(t, err == nil)
+		tests.Assert(t, r.StatusCode == http.StatusOK)
+		if r.ContentLength <= 0 {
+			time.Sleep(time.Millisecond * 10)
+			continue
+		} else {
+			// Should have node information here
+			tests.Assert(t, r.Header.Get("Content-Type") == "application/json; charset=UTF-8")
+			err = utils.GetJsonFromResponse(r, &info)
+			tests.Assert(t, err == nil)
+			break
+		}
+	}
+	return info
+}
+
 func TestBlockVolumeCreateHACount(t *testing.T) {
 	tmpfile := tests.Tempfile()
 	defer os.Remove(tmpfile)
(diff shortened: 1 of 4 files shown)


obnoxxx:
retest this please

obnoxxx:
retest this please

phlogistonjohn:
rebased

pkg/utils/shuffle.go
```diff
@@ -0,0 +1,44 @@
...
+	for ; i > 0; i-- {
+		j := int(r.Int31n(int32(i + 1)))
+		swap(i, j)
+	}
+}
+
+// SeededShuffle pseudo-randomizes the order of elements
+// using a private PRNG instance seeded from the clock.
+func SeededShuffle(n int, swap func(i, j int)) {
+	r := rand.New(rand.NewSource(time.Now().UnixNano()))
```
ansiwen: This effectively disables the PRNG, because it is seeded every call with the time (which doesn't have to be random for a couple of reasons). I would suggest to put `r` into a closure and initialize it only on the first call.

obnoxxx:
retest this please

pkg/utils/shuffle_test.go
```diff
@@ -0,0 +1,63 @@
...
+	copy(a2, a)
+	SeededShuffle(len(a2), func(i, j int) {
+		a2[i], a2[j] = a2[j], a2[i]
+	})
+	same := true
+	for i := 0; i < len(a); i++ {
+		dedup[a2[i]] += 1
+		same = same && (a[i] == a2[i])
+	}
+	tests.Assert(t, !same, a, a2)
```
ansiwen: I expect this test to fail every 120th run on average, because that is the likelihood, that shuffling of 5 elements returns the identical sequence.

pkg/utils/shuffle_test.go
```diff
@@ -0,0 +1,63 @@
...
+	a2 := make([]string, len(a))
+	copy(a2, a)
+	Shuffle(r, len(a2), func(i, j int) {
+		a2[i], a2[j] = a2[j], a2[i]
+	})
+	same := true
+	for i := 0; i < len(a); i++ {
+		same = same && (a[i] == a2[i])
+	}
+	tests.Assert(t, !same, a, a2)
```
ansiwen: I expect this test to fail every 120th run on average, because that is the likelihood, that shuffling of 5 elements returns the identical sequence. In this case you could seed r with a fixed number above and make use of the deterministic behavior of the PRNG.

apps/glusterfs/app_block_volume_test.go
```diff
@@ -310,27 +336,8 @@ func TestBlockVolumeCreateHACount(t *testing.T) {
 	// Send request
 	r, err := http.Post(ts.URL+"/blockvolumes", "application/json", bytes.NewBuffer(request))
 	tests.Assert(t, err == nil)
-	tests.Assert(t, r.StatusCode == http.StatusAccepted)
-	location, err := r.Location()
-	tests.Assert(t, err == nil)
+	info := blockVolumeTestResult(t, r)
```
ansiwen: Can we do that also for the tests of PR #1280 ?

apps/glusterfs/app_block_volume_test.go
```diff
@@ -492,6 +499,51 @@ func TestBlockVolumeCreateHACountHostUnavailableFail(t *testing.T) {
...
+		r, err := http.Post(ts.URL+"/blockvolumes", "application/json", bytes.NewBuffer(request))
+		tests.Assert(t, err == nil)
+		info := blockVolumeTestResult(t, r)
+
+		// check that the hosts counts are as expected and match the ha counts
+		// verify that we do not get the same hosts every time this function is
+		// called.
+		tests.Assert(t, len(info.BlockVolume.Hosts) == 3)
+		tests.Assert(t, info.Hacount == 3)
+		tests.Assert(t, !reflect.DeepEqual(prevHosts, info.BlockVolume.Hosts),
```
ansiwen: I expect this test to fail every 3.6 millionth run on average, because that is the likelihood, that shuffling of 10 elements returns the identical sequence. That's unlikely... I know... but... it would still itch me... ;-)

pkg/utils/shuffle_test.go
obnoxxx: Gosh, accelent point! This is just plain wrong. How can we test that the hosts are not always in the same order but can sometimes be?

ansiwen: I would not test it, to be honest, else we are just retesting Go's implementation. I would probably only test it with a statically seeded case, so it's deterministic. Or repeat the test a couple of times? But this really sounds "over-tested" to me.

ansiwen: Well, repeating makes it exponentially less likely that you have a "collision", so that would work. But as I said, it sounds like too much. All we actually want to test, is that shuffle is called. So I personally would seed the global PRNG with a fixed number before the test. And in production code I would make sure that we do proper seeding of it in the main function.

phlogistonjohn:
This PR is made obsolete by PR #1286. Closing

PR #1284, opened 2018-07-26
Update volume hosts info during node removal

### LINKED ISSUE #1273: newly created volume info showing deleted node IP
### Kind of issue

> Bug

### Observed behavior
I have done node removal in heketi and created a volume, but new volume info contains the removed node IP

### Expected/desired behavior
heketi should pick up the node IP from the heketi cluster, newly created volume should have contains removed node IP

1)load topology in heketi (it has 3 nodes)
2)create 4  volumes
3)add a new node to heketi
4)disable one old node in heketi
5)remove the same node
6)remove devices
7)delete the node from heketi
8)create the new volume (check vol info)

- Heketi version used (e.g. v6.0.0 or master): 
- Operating system used: 
- Heketi compiled from sources, as a package (rpm/deb), or container: 
- If container, which container image: 
- Using kubernetes, openshift, or direct install: 
- If kubernetes/openshift, is gluster running inside kubernetes/openshift or outside: 
- If kubernetes/openshift, how was it deployed (gk-deploy, openshift-ansible, other, custom):

### LINKED ISSUE #1274: after node delete from heketi, backup-volfile-servers in volumes contains removed node IP
### Kind of issue

 Bug

### Observed behavior
after node delete from heketi, the volumes present in the heketi is having ```backup-volfile-servers``` as removed node IP

### Expected/desired behavior
once the node is deleted  from heketi, heketi should refresh volume info as it should not point out to deleted node IP

1)load topology in heketi (it has 3 nodes)
2)create 4  volumes
3)add a new node to heketi
4)disable one old node in heketi
5)remove the same node
6)remove devices
7)delete the node from heketi
8)check volume info on old volumes

- Heketi version used (e.g. v6.0.0 or master): 
- Operating system used: 
- Heketi compiled from sources, as a package (rpm/deb), or container: 
- If container, which container image: 
- Using kubernetes, openshift, or direct install: 
- If kubernetes/openshift, is gluster running inside kubernetes/openshift or outside: 
- If kubernetes/openshift, how was it deployed (gk-deploy, openshift-ansible, other, custom):

Madhu-1:

during the node removal in heketi, we replace the brick from removal node to another node, during this process we need to update the volume info such as Hosts (which may contain the removal node hostname). and we need to update the backup-volfile-servers and volume mountpoint  (this also may contain the removal hostname)

if it's a block hosting volume, we need to fetch all the block volumes belongs to it, and update the hosts in those volumes, so that next block volume creation fetches the new hosts.

consider only online nodes for a volume mountpoint update 
if the node statue is online, then consider this node for updating the mount point and backup-volfile-servers of a volume.

### Does this PR fix issues?

Fixes #1274
Fixes #1273

### DIFF
--- a/apps/glusterfs/volume_entry_allocate.go
+++ b/apps/glusterfs/volume_entry_allocate.go
@@ -11,6 +11,7 @@ package glusterfs
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/boltdb/bolt"
 	"github.com/heketi/heketi/executors"
@@ -391,6 +392,7 @@ func (v *VolumeEntry) replaceBrickInVolume(db wdb.DB, executor executors.Executo
 	// have changed
 
 	err = db.Update(func(tx *bolt.Tx) error {
+		txdb := wdb.WrapTx(tx)
 		err = newBrickEntry.Save(tx)
 		if err != nil {
 			return err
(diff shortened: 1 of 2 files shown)


apps/glusterfs/volume_entry_create.go
```diff
@@ -55,7 +55,10 @@ func (v *VolumeEntry) updateMountInfo(db wdb.RODB) error {
 			if err != nil {
 				return err
 			}
-			hosts = append(hosts, node.StorageHostName())
+			if node.isOnline() {
```
obnoxxx: This part is wrong: A node being offline just means heketi won't place new brick on it. This is typically a temporary maintenance state. It can be a step towards node deletion, but before a node is deleted, we should not remove the node from the mount info! It is still a valid node of the gluster cluster.

apps/glusterfs/volume_entry_allocate.go
```diff
@@ -425,8 +427,58 @@ func (v *VolumeEntry) replaceBrickInVolume(db wdb.DB, executor executors.Executo
 		if err != nil {
 			return err
 		}
-		err = reReadVolEntry.Save(tx)
-		if err != nil {
+
+		updateNodeHostName := true
```
obnoxxx: I'm afraid this whole block is not correct. We don't need to update the mount info when we replace a brick. The mount host and backup volfile servers should always be *all* nodes in the gluster cluster.

Upon node deletion, when the node is really taken out of the gluster cluster for good, we should update the mount info for each volume...

obnoxxx: maybe we should do a new patch altogether? sorry.

obnoxxx:
@Madhu-1 thanks for the patch, don't feel bad for the comments. I hope I explained well.

Madhu-1:
@obnoxxx, I thought we should replace all the hosts and mountpoint after removing the node.
I will send a new patch to update hosts and mountpoints after deleting the node from heketi management.

PR #1287, opened 2018-07-26
apps:glusterfs: reserve 1% of raw capacity in FreeSpace tracking

obnoxxx:

A block-hosting volume created with raw size X can not host block files of accumulated size X because the file system uses a certain small amount of the raw size. But the FreeSpace tracking does currently not take this into account. Hence trying to create a block volume of size X with FreeSize X available on the block hosting volume will fail. 

This PR fixes heketi to always track only 99% of the raw space of a newly created (or expanded) block hosting volume as the available FreeSize.

### Does this PR fix issues?

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1596018


obnoxxx:
Debt: A  test for the explicit problem is currently missing. It probably needs to be a functional test since it would need to interact with gluster.

PR #1291, opened 2018-07-27
apps: don't waste entropy for seeding

ansiwen:

Seed values that have the same remainder when divided by 2^31-1
generate the same pseudo-random sequence. So there is no need waste
entropy to generate a 64 bit random number.

Also this change removes an unnecessary UTC() call, since UnixNano() ignores the timezone info.

### DIFF
--- a/main.go
+++ b/main.go
@@ -291,7 +291,7 @@ func randSeed() {
 	max := big.NewInt(1<<31 - 1)
 	n, err := crand.Int(crand.Reader, max)
 	if err != nil {
-		rand.Seed(time.Now().UTC().UnixNano())
+		rand.Seed(time.Now().UnixNano())
 	} else {
 		rand.Seed(n.Int64())
 	}
--- a/main.go
+++ b/main.go
@@ -12,7 +12,6 @@ package main
 import (
 	crand "crypto/rand"
 	"fmt"
-	"math"
 	"math/big"
 	"math/rand"
 	"net/http"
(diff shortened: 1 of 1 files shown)


nixpanic:
add to whitelist

nixpanic:
ok to test

obnoxxx:
@ansiwen writes:
> @obnoxxx: the only entropy-consuming function is crand.Int(), 

So far, that was clear to me.

> so by setting its max argument to 2^31-1, we only consume as much entropy as can be used by the rand.Seed() function.

Yes, that was (part of) my question. But doesn't that statement imply knowledge about the implementation of `crand.Int()`? Hence my question whether just to use a 32 bit random function
instead of our 64-bit one, to be sure that this change makes a difference at all... Looking at the implementation it seems that it reads random values until one is below the given max value. Hence I don't really know whether we change anything really when doing this to the crand.Int() funciton. It will probably actually take more cycles internally to get a random number below a smaller threshold. But we will avoid producing two different value that yield the same seed effectively. 

I also have to admit that I always have some difficulties with how the word entropy is used in computer science. It seems somewhat counter-intuitive when looking from a thermodynamics angle. I know it's supposed to be essentially the same term and mean disorder or randomness in the system, but when talked about concretely, it always confuses me a bit. :-) This may be the real reason of my problem with understanding what you're saying with "trying waste less entropy" or "the function consumes entropy". (As stated above, it seems to me that we may even be consuming more entropy when restricting the initial random number to 32bits, but we get off better with different random number sequences for different seeds returned.)

>  The format, in which this number is stored doesn't influence how much entropy is used before, and since rand.Seed() takes an int64, we directly convert the big.Int into a int64.

So all that said, the main question I have is about the argument. Do we actually consume less entropy, or do we compute more entropy for a better result?

obnoxxx:
@ansiwen and thanks for splitting the patches! :-)

ansiwen:
@obnoxxx: ok, now I got your point. I have to admit I didn't check the implementation of crand.Int() before, but I did now. The implementation only reads as many entropy bytes as is required to represent max-1, and then clears all bits, that are not required. Only if then still the number is higher than max-1, it is repeated. The likelihood that this happens is always smaller than 50% (that is [next power of 2 - max]/[next power of 2]). In case of 2^31-1 that is 1:2^31, so extremely low. My conclusion is, this change cuts the entropy consumption in half. :-)

> In computing, entropy is the randomness collected by an operating system or application for use in cryptography or other uses that require random data. This randomness is often collected from hardware sources (variance in fan noise or HDD), either pre-existing ones such as mouse movements or specially provided randomness generators. A lack of entropy can have a negative impact on performance and security. (https://en.wikipedia.org/wiki/Entropy_(computing))

PR #1294, opened 2018-07-30
apps:glusterfs update volume hosts info during node deletion

### LINKED ISSUE #1273: newly created volume info showing deleted node IP

### LINKED ISSUE #1274: after node delete from heketi, backup-volfile-servers in volumes contains removed node IP

Madhu-1:
during the node delete in heketi, we remove the node from
heketi management, volume info such as Hosts (which may
contain the deleting node hostname). and we need to
update the backup-volfile-servers and volume mountpoint
(this also may contain the deleting node hostname)

if it's a block hosting volume, we need to
fetch all the block volumes belongs to it,
and update the hosts in those volumes, so that
next block volume creation fetches the new
hosts from the heketi cluster.

### Does this PR fix issues?

Fixes #1274
Fixes #1273

### DIFF
--- a/apps/glusterfs/app_node.go
+++ b/apps/glusterfs/app_node.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/boltdb/bolt"
 	"github.com/gorilla/mux"
+	wdb "github.com/heketi/heketi/pkg/db"
 	"github.com/heketi/heketi/pkg/glusterfs/api"
 	"github.com/heketi/heketi/pkg/utils"
 )
@@ -289,7 +290,6 @@ func (a *App) NodeDelete(w http.ResponseWriter, r *http.Request) {
 
 		// Remove from db
 		err = a.db.Update(func(tx *bolt.Tx) error {
-
 			// Get Cluster
 			cluster, err := NewClusterEntryFromId(tx, node.Info.ClusterId)
 			if err == ErrNotFound {
@@ -320,6 +320,11 @@ func (a *App) NodeDelete(w http.ResponseWriter, r *http.Request) {
 				return err
 			}
 
+			err = refreshVolumeNodes(tx, node)
+			if err != nil {
+				logger.Err(err)
+				return err
+			}
 			return nil
 
 		})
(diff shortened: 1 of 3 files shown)


apps/glusterfs/app_node.go
```diff
@@ -320,6 +321,49 @@ func (a *App) NodeDelete(w http.ResponseWriter, r *http.Request) {
 				return err
 			}
 
+			volumes, err := ListCompleteVolumes(tx)
```
phlogistonjohn: This feels like it should be it's own function, it's fairly long and conceptually independent from the surrounding code. Something like `refreshVolumeNodes(tx, clusterID)`

phlogistonjohn: I think this should only iterate over the volumes in the cluster that was changed rather than all volumes.

apps/glusterfs/app_node.go
```diff
@@ -320,6 +321,49 @@ func (a *App) NodeDelete(w http.ResponseWriter, r *http.Request) {
...
+						}
+
+						if err := blockEntry.Save(tx); err != nil {
+							logger.LogError("Save block entry for ID %s Failed with error %s", id, err.Error())
+							continue
+						}
+					}
+				}
+				//update mountpoint info
+				if err = volEntry.updateMountInfo(txdb); err != nil {
```
phlogistonjohn: I don't think this function will do what we want. Based on my reading of the code this may wipe out any other options in the map other than backup-volfile-servers.

Additionally, this function has the same problem as I mentioned above about repeatedly going to the db only to get the same results over and over again in a loop.

I suggest:
* Extracting (as you've already partially done) the inner parts of `updateMountInfo` into new functions:
  * One that iterates over the nodes in the cluster and returns a list of the storage host names of those nodes. This function would be standalone and not attached to any struct.
  * Make the latter half of updateMountInfo into a helper function that takes a list of hosts and sets/updates the hosts and Options.
* Call it from updateMountInfo and before the start of the loop over the volumes.
* Now the loop over the block volumes sets .Hosts to the new list
* The loop over the volumes uses the helper function to update (& not overwrite) the mountpoint and options.

apps/glusterfs/app_node.go
```diff
@@ -320,6 +321,49 @@ func (a *App) NodeDelete(w http.ResponseWriter, r *http.Request) {
...
+				if volEntry.Info.Block {
+					for _, id := range volEntry.Info.BlockInfo.BlockVolumes {
+
+						blockEntry, err := NewBlockVolumeEntryFromId(tx, id)
+						if err != nil {
+							logger.LogError("Get block volume entry for ID %s Failed with error %s", id, err.Error())
+							continue
+						}
+
+						if err = blockEntry.updateHosts(txdb); err != nil {
```
phlogistonjohn: This strikes me as rather suboptimal. We're in a nested loop calling the same "db queries" over and over again in a transaction such that we're always going to get the same results. It strikes me as more correct to simply fetch the list of nodes for the cluster and re-use that list throughout the loop. See my comment on `updateMountInfo` for more details.

phlogistonjohn:
ok to test

humblec:
LGTM.

PR #1298, opened 2018-08-01
Heketi client uses new port for every get

raghavendra-talur:
This is a WIP PR to show that heketi client library creates a new http client for every GET operation that it performs while checking the status of a async operation. It should rather create a http client object when the heketi client object is created and should be cached.

### Notes for the reviewer
Note that the PR currently ignores many other options that are stored in the client struct of heketi. I will update the patch if we all agree that the http client object should be reused.

### DIFF
--- a/apps/glusterfs/operations_manage.go
+++ b/apps/glusterfs/operations_manage.go
@@ -13,6 +13,7 @@ import (
 	"fmt"
 	"net/http"
 	"sync"
+	"time"
 
 	"github.com/heketi/heketi/executors"
 
@@ -97,6 +98,7 @@ func AsyncHttpOperation(app *App,
 		// decrement the op counter once the operation is done
 		// either success or failure
 		defer app.opcounter.Dec()
+		time.Sleep(10 * time.Second)
 		logger.Info("Started async operation: %v", label)
 		if err := op.Exec(app.executor); err != nil {
 			if _, ok := err.(OperationRetryError); ok && op.MaxRetries() > 0 {
--- a/client/api/go-client/client.go
+++ b/client/api/go-client/client.go
@@ -55,10 +55,11 @@ type ClientOptions struct {
 
 // Client object
 type Client struct {
-	host     string
-	key      string
-	user     string
-	throttle chan bool
+	host       string
+	key        string
+	user       string
+	throttle   chan bool
+	httpClient *http.Client
 
 	// configuration for TLS support
 	tlsClientConfig *tls.Config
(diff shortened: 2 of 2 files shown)


raghavendra-talur:
For comparision,

with only the first commit in the PR, heketi-cli used up 12 ports over one volume create operation as it pinged server once every second for the status.

with both the patches applied, heketi-cli used only 1 port for the volume create operation.

client/api/go-client/client.go
```diff
@@ -178,14 +184,7 @@ func (c *Client) doBasic(req *http.Request) (*http.Response, error) {
 		<-c.throttle
 	}()
 
-	httpClient := &http.Client{}
-	if c.tlsClientConfig != nil {
```
Madhu-1: @raghavendra-talur, any specific reason to remove ```TLSClientConfig: c.tlsClientConfig,``` TLS Client configuration?

apps/glusterfs/operations_manage.go
```diff
@@ -97,6 +98,7 @@ func AsyncHttpOperation(app *App,
 		// decrement the op counter once the operation is done
 		// either success or failure
 		defer app.opcounter.Dec()
+		time.Sleep(10 * time.Second)
```
Madhu-1: do we still need this Sleep?  I think it may add an extra delay.  if its required can it be reduced to a smaller value?

phlogistonjohn: I think that's just there to help prove the point and would not be part of a long term change.

humblec: may be I am missing something,  can you please explain  how this delay helps here ?

phlogistonjohn: @raghavendra-talur  correct me if I am wrong, but I think the point is that it makes the problem worse by prolonging the TCP connection in order to churn through the ports more quickly. Like I mentioned after @Madhu-1 commented I doubt this will be part of any PR that actually gets merged.

raghavendra-talur: @Madhu-1 @humblec 

@phlogistonjohn is right, the PR in current state is only to demonstrate the bug. By increasing the time server spends on servicing the request I was able to clearly see the increase in port consumption by the client. 

The fix would be to embed the http client object in the heketi client object. When done properly it will have the TLS cert config also considered.  The question is, do we all agree with the approach?

phlogistonjohn:
I think the general approach is OK, although it does make me wonder why it wasn't done that way initially. Was there intent or did it just happen, and if there was intent why?

We may also want to see if there are any server changes we can make to help here because it takes a long time for client library changes to propagate to all of the client apps.

raghavendra-talur:
@phlogistonjohn I have run out of ideas to fix this on server side. The only method that might work but is not recommended is to use SetLinger and to set the value to 0. It is not recommended as it might cause problems with connection states.

I would rather change this in client and let the consumers take it slowly in upcoming releases.

phlogistonjohn:
Understandable. Can we create an issue out this PR and later followup on that issue with changes to the client libs?

obnoxxx:
Any new thoughts on this?

PR #1299, opened 2018-08-02
apps: Reject parallel block hosting volume creation

phlogistonjohn:

When a block volume create operation is started check that there
are no pending block hosting volumes being created. If there are,
reject the request with the too-many-requests http code.
This avoids creating multiple block hosting volumes when only one
should be created (assuming it can fist the subsequent request).

### Does this PR fix issues?

Fixes [rhbz#1607821](https://bugzilla.redhat.com/show_bug.cgi?id=1607821)

### Notes for the reviewer

This patch is incorrect if the pending block hosting volumes are part of stale pending operations. However, we do not have a mechanism to determine if an operation is stale or not. Before this PR can be merged we need to add support for distinguishing between live and stale operations and allowing "parallel" block hosting volume completes when the operation is stale and the volume create would not be satisfied.

I'm filing the PR now for early feedback.


phlogistonjohn:
Rebased on master

obnoxxx:
retest this please

obnoxxx:
(that was an ssh connection error)

obnoxxx:
retest this please

phlogistonjohn:
> The only nit, or rather a remark, that I have is that NewApp does the DB upgrade and then already looks for pending operations. Wouldn't that have been a natural place to put the stale-marking? Or else maybe move the warning about stale operations to setupApp?

I did think about putting it there, but I started thinking about needing a split between "schema" upgrades & db consistency, vs. db content & functionality. Because this felt like a content change rather than a correctness or schema change I felt that we needed a new way to represent changes of this this type. Plus, db upgrade is part of db open currently and I really wanted this to be something distinct that is only run on server startup.

It might seem funny that in PR #1302 I am adding to the db upgrade code path. But, I'd argue, in this case we're changing the content of the db to be internally consistent rather than making a "state change".

PR #1300, opened 2018-08-03
function test: added test cases for block volumes

### LINKED ISSUE #1288: Add tests for block hosting raw size vs. real size
PR #1287 added a feature to calculate the free size of a hosting volume as less than the total volume size (raw size) due to filesystem overhead. This change fixed existing tests but needs new tests of their own.

Madhu-1:

This added a new functional test setup to test block volume operations(create,get and delete block volumes)
added test cases to test folloowing sceneraios
* creation of block volume
* get  and deletion of block volumes

### Does this PR fix issues?

Fixes #1288

### DIFF
--- a/tests/functional/TestSmokeTest/tests/block_volume_test.go
+++ b/tests/functional/TestSmokeTest/tests/block_volume_test.go
@@ -1,0 +1,73 @@
+// +build functional
+
+//
+// Copyright (c) 2018 The heketi Authors
+//
+// This file is licensed to you under your choice of the GNU Lesser
+// General Public License, version 3 or any later version (LGPLv3 or
+// later), or the GNU General Public License, version 2 (GPLv2), in all
+// cases as published by the Free Software Foundation.
+//
+package functional
+
+import (
+	"testing"
+
+	"github.com/heketi/heketi/pkg/glusterfs/api"
+	"github.com/heketi/tests"
+)
+
+func TestBlockVolumeOperation(t *testing.T) {
+
+	// Setup the VM storage topology
+	setupCluster(t, 3, 4)
+	defer teardownCluster(t)
+
+	defer teardownBlock(t)
+
+	req := &api.BlockVolumeCreateRequest{}
+	//check it is not possible to create block volume if  size is greated then block hosting volume
+	req.Size = 101
+
+	_, err := heketi.BlockVolumeCreate(req)
+	tests.Assert(t, err != nil, "expected err != nil, got:", err)
+
+	//check it is not possible to create block volume as same size of block hosting volume
+	req.Size = 100
+	_, err = heketi.BlockVolumeCreate(req)
+	tests.Assert(t, err != nil, "expected err != nil, got:", err)
+
+	// 1% is reserved in blockhosting volume, we should be able to create
+	//block volumes of total size 98 GB
+	req.Size = 2
+	for i := 1; i <= 49; i++ {
+		vol, err := heketi.BlockVolumeCreate(req)
+		tests.Assert(t, err == nil, "expected err == nil, got:", err)
+		tests.Assert(t, vol.Size == 2, "expected vol.Size == 2 got:", vol.Size)
+	}
+
+	volList, err := heketi.BlockVolumeList()
+	tests.Assert(t, err == nil, "expected err == nil, got:", err)
+	tests.Assert(t, len(volList.BlockVolumes) == 49, "expected len(volList.BlockVolumes) == 49 got:", len(volList.BlockVolumes))
+
+	for _, ID := range volList.BlockVolumes {
+		volInfo, err := heketi.BlockVolumeInfo(ID)
+		tests.Assert(t, err == nil, "expected err == nil, got:", err)
+		tests.Assert(t, volInfo.Size == 2, "expected volInfo.Size == 2 got:", volInfo.Size)
+
+		err = heketi.BlockVolumeDelete(ID)
+		tests.Assert(t, err == nil, "expected err == nil, got:", err)
+	}
+
+}
+
+func teardownBlock(t *testing.T) {
+
+	volList, err := heketi.BlockVolumeList()
+	tests.Assert(t, err == nil, "expected err == nil, got:", err)
+
+	for _, ID := range volList.BlockVolumes {
+		err = heketi.BlockVolumeDelete(ID)
+		tests.Assert(t, err == nil, "expected err == nil, got:", err)
+	}
+}
--- a/tests/functional/TestSmokeTest/vagrant/roles/gluster/tasks/main.yml
+++ b/tests/functional/TestSmokeTest/vagrant/roles/gluster/tasks/main.yml
@@ -1,8 +1,12 @@
-- name: install glusterfs
+- name: install glusterfs, gluster-block
   yum: name={{ item }} state=present
   with_items:
     - glusterfs-server
     - glusterfs-client
+    - gluster-block
 
-- name: start glusterd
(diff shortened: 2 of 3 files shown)


phlogistonjohn:
Last I discussed with @raghavendra-talur we were hoping to reduce the number of vagrant environments needed by the functional tests. So I'm not enthusiastic about seeing a new one added here.
It seems to me that we ought to be able to add gluster block packages and services to  the TestSmokeTest vagrant environment and avoid creating another one. Please let me know if you think that there additional reasons why you may not want to do that.

Madhu-1:
@phlogistonjohn I have added another vagrant environment to keep gluster block related functional tests separate, I don't want to disturb the current smoke test environment. if you guys agree I will add block related test cases in TestSmokeTest itself.

tests/functional/TestSmokeTest/tests/block_volume_test.go
```diff
@@ -0,0 +1,60 @@
...
+	setupCluster(t, 3, 4)
+
+	req := &api.BlockVolumeCreateRequest{}
+	//check is it possible to create block volume if  size is greated then block hosting volume
+	req.Size = 101
+
+	_, err := heketi.BlockVolumeCreate(req)
+	tests.Assert(t, err != nil, "expected err != nil, got:", err)
+
+	//check is it possible to create block volume as same size of block hosting volume
```
phlogistonjohn: Shouldn't this be "check it is NOT possible...." ?

tests/functional/TestSmokeTest/tests/block_volume_test.go
```diff
@@ -0,0 +1,60 @@
...
+	req.Size = 100
+	_, err = heketi.BlockVolumeCreate(req)
+	tests.Assert(t, err != nil, "expected err != nil, got:", err)
+
+	// 1% is reserved in blockhosting volume, we should be able to create
+	//block volumes of total size 98 GB
+	for i := 1; i <= 49; i++ {
+		req.Size = 2
```
phlogistonjohn: Why are you setting req.Size inside the loop?

Madhu-1: this can be moved outside

Madhu-1:
@phlogistonjohn I have run the complete smoke test locally, I have not seen any issue, I can update teardown.
do you want me to add block volume deletion in teardown function?

phlogistonjohn:
ok to test

phlogistonjohn:
> I have run the complete smoke test locally, I have not seen any issue, I can update teardown.
do you want me to add block volume deletion in teardown function?

My expectations don't necessarily match reality. I've enabled centos-ci for this PR. At the same time I'll re-read the cleanup code more carefully.

tests/functional/TestSmokeTest/tests/block_volume_test.go
Madhu-1: updated comment

tests/functional/TestSmokeTest/tests/block_volume_test.go
Madhu-1: moved outside of for loop

tests/functional/TestSmokeTest/vagrant/roles/gluster/tasks/main.yml
```diff
@@ -6,3 +6,11 @@
 
 - name: start glusterd
   service: name=glusterd state=started enabled=yes
+
+- name: install gluster block
+  yum: name={{ item }} state=present
+  with_items:
+    - gluster-block
```
phlogistonjohn: I suggest merging this with the previous block that installs the other gluster packages.

phlogistonjohn:
I swear I read this code something like three times over and every time missed that you are deleting the block volumes in the final loop.

This is OK to do, but we can't rely entirely on the test always passing. For example, imagine we tweak the size accounting and we only can fit 48 volumes. If the test is run without changes the 49th volume will fail the assertion and the final loop will not be run. Then the cleanup function will fail because the block volumes still exist.
I recommend either adding an additional blockTeardown function that will be called from a defer prior to the teardownCluster function OR updating teardownCluster to also remove block volumes.

Madhu-1:
@phlogistonjohn in case of failure we will face the issue. sure I will add a function to clear created block hosting volumes

Madhu-1:
@phlogistonjohn addressed review comments

tests/functional/TestSmokeTest/tests/block_volume_test.go
```diff
@@ -0,0 +1,73 @@
+// +build functional
+
+//
+// Copyright (c) 2015 The heketi Authors
```
phlogistonjohn: Need to update the date here as this is a new file.

Madhu-1: done

phlogistonjohn:
retest this please

tests/functional/TestSmokeTest/tests/block_volume_test.go
```diff
@@ -0,0 +1,73 @@
...
+
+	"github.com/heketi/heketi/pkg/glusterfs/api"
+	"github.com/heketi/tests"
+)
+
+func TestBlockVolumeOperation(t *testing.T) {
+
+	// Setup the VM storage topology
+	defer teardownCluster(t)
+	defer teardownBlock(t)
```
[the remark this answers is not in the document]
Madhu-1: I had 2 options to delete block volume as @phlogistonjohn suggested
* integrate with teardown function
* create own block teardown function

all block related test cases go into this file, so I that instead of disturbing the old implementation(which i may introduce new bugs) 
i have created a new function to delete block volumes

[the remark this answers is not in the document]
Madhu-1: defer should be before setupCluster.
cleanup functions should get called in defer before calling the other functions if functions call fails, defer will be executed to clean up resources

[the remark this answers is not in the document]
obnoxxx: @Madhu-1, normally the pattern that we use is this:

1. call function1
2. immediately afterwards: defer call to cleanup-function-for-function1

The main point here is that there is usually no point in calling the cleanup function for function1 before function1 was called. Also if we follow this pattern, then it is always easy to spot since we group the function and its cleanup function together. It would be ok to reverse the process, to first call defer cleanup_function1 and immediately afterwards call function1, but they are equivalent. 

That said, I don't fully understand @raghavendra-talur 's argument "defer should come after setupCluster, just in case setupCluster fails" -- the order (of function and its defer cleanup) is not that important, it is more important to group together, and we have a habit for the ordering.

But it is important that the cleanup for the later action (in our case block volume stuff) is defer-called later, because the deferred cleanups are called in reverse order.

So I would suggest to
* call `defer teardownCluster()` after `setupCluster()`
* call `defer teardownBlock()` after `setupCluster()` and after `defer teardownCluster()`.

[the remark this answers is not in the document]
Madhu-1: @obnoxxx am little confused here, why we should call `defer teardownCluster()` after `setupCluster()`
if `setupCluster()` fails how to defer will get called( which is after `setupCluster`)?

Madhu-1:
@obnoxxx @raghavendra-talur moved `teardownCluster` and `teardownBlock` after setupCluster

Madhu-1:
@raghavendra-talur @obnoxxx, if I have addressed your comments can we merge this one?

PR #1302, opened 2018-08-03
apps: return errors rather than panic when checking FreeSize bounds

phlogistonjohn:

Change the VolumeEntry's ModifyFreeSize and ModifyReservedSize
functions to return errors instead of panicking. Also plumb
the error handling up to the callers of these functions.
This will prevent wrong values from being saved to the db
without killing the server. The admin of the server can then
stop the server at a later time and repair the db as needed.

### DIFF
--- a/apps/glusterfs/volume_entry.go
+++ b/apps/glusterfs/volume_entry.go
@@ -139,7 +139,13 @@ func NewVolumeEntryFromRequest(req *api.VolumeCreateRequest) *VolumeEntry {
 	vol.GlusterVolumeOptions = req.GlusterVolumeOptions
 
 	if vol.Info.Block {
-		vol.SetRawCapacity(req.Size)
+		if err := vol.SetRawCapacity(req.Size); err != nil {
+			logger.Err(err)
+			// we can either panic here or return nil. We panic because
+			// returning nil is most likely going to lead to nil
+			// dereference panics anyway
+			panic(err)
+		}
 		// prepend the gluster-block group option,
 		// so that the user-specified options can take precedence
 		blockoptions := strings.Split(BlockHostingVolumeOptions, ",")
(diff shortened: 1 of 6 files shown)


phlogistonjohn:
Added patches to auto-correct free size value on db upgrade if possible.

apps/glusterfs/volume_entry.go
```diff
@@ -335,21 +355,71 @@ func ReduceRawSize(size int) int {
...
+	}
+	return nil
 }
 
 func (v *VolumeEntry) SetRawCapacity(delta int) {
 	v.Info.BlockInfo.FreeSize = 0
 	v.Info.BlockInfo.ReservedSize = 0
-	v.AddRawCapacity(delta)
+	if err := v.AddRawCapacity(delta); err != nil {
+		// if this fails, something is deeply wrong and panic is appropriate
```
obnoxxx: How is this any different from e.g. AddRawCapacity?
I.e., why do we need to treat this specially?

phlogistonjohn: SetRawCapacity should only be called when the volume is being created. It really should not be able to trigger the error conditions being checked because of the values being set to zero. Thus the "deeply wrong" comment.

If the asymmetry bugs you I can easily change it to return an error, the rarity/severity of it seemed like it was still more of a panic level thing.

raghavendra-talur:
@obnoxxx Merge if your question has been answered.

phlogistonjohn:
Rebased on master

apps/glusterfs/volume_entry.go
obnoxxx: It is imho irrelevant that `FreeSize` has been set to 0, because the resulting FreeSize is checked against `volume.Info.Size`. So if you call `SetRawCapacity()` with something larger than `volume.Info.Size` it will fail.

In that regard, `SetRawCapacity()` is not different than any other call of `ModifyFreeSize()` or `ModifyReservedSize()`. Hence I would say either panic all or panic none.

Making any sense?

phlogistonjohn: Panic none it is then. I will implement as soon as I can.

phlogistonjohn:
>    The introduction of the FreeSize correcting function is not actually part of the PR description (but, well, ok.. ;-) ).

The PR evolved a bit from when I first submitted. Do you want me to update it?

>    Is the correct location for doing such changes really the UpgradeDB function or rather (as you suggested yourself) the newly introduced ServerReset() function or similar. Since this not a schema update.

I worked on those changes in parallel and didn't have the opportunity to use that function here. I can of course do so. It wouldn't be a lot of extra work.

>    Finally this change makes me think whether we should not store the free size after all but only the reserved size and the block volume sizes and always calculate the remaining free size.... would at least remove one source of inconsistency...

Generally, I think this is the right direction to go it but I didn't want to make such a radical change yet. I think the counterargument to calculating the free size (and effectively the used size) is probably the extra cost of looking up the items in question in the db. However, I think that calculating the sizes (say, once per db-transaction) and then caching the result would be a good compromise between not having to deal with inconsistency in the db and performance.

Putting on "request changes" again to get these questions discussed before we merge. All that can done as later patches, imho, but asking for your opinion now.

PR #1303, opened 2018-08-03
middleware: ignore iat claim in jwt

raghavendra-talur:
This PR implements new claims type and validation function. We need it due to inability of jwt library to handle clock skew.

### Notes for the reviewer
1. required_claims is embedded in validation function and missing claims error is replaced with unauthorized
2. As claims type has changed from map to struct, claims are now accessed as members and not indexed.

### DIFF
--- a/middleware/jwt.go
+++ b/middleware/jwt.go
@@ -115,10 +115,10 @@ func (j *JwtAuth) ServeHTTP(w http.ResponseWriter, r *http.Request, next http.Ha
 		errmsg := fmt.Sprintf("Invalid JWT token: %s", err)
 		// annoying that the types don't actually match
 		if err.Error() == jwt.ErrSignatureInvalid.Error() {
-			errmsg += "\n(client and server secrets may not match)"
+			errmsg += " (client and server secrets may not match)"
 		}
 		if strings.Contains(err.Error(), "used before issued") {
-			errmsg += "\n(client and server clocks may differ)"
+			errmsg += " (client and server clocks may differ)"
 		}
 		http.Error(w, errmsg, http.StatusUnauthorized)
 		return
--- a/middleware/jwt.go
+++ b/middleware/jwt.go
@@ -15,22 +15,76 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"os"
+	"strconv"
 	"strings"
+	"time"
 
 	"github.com/auth0/go-jwt-middleware"
 	jwt "github.com/dgrijalva/jwt-go"
 	"github.com/gorilla/context"
+	"github.com/heketi/heketi/pkg/utils"
 )
+
+var (
+	logger          = utils.NewLogger("[jwt]", utils.LEVEL_DEBUG)
+	iatLeeway int64 = 120
+)
+
+func init() {
+	if leeway, exists := os.LookupEnv("HEKETI_JWT_IAT_LEEWAY_SECONDS"); exists {
+		value, err := strconv.ParseUint(leeway, 10, 64)
+		if err != nil {
+			logger.Info("HEKETI_JWT_IAT_LEEWAY_SECONDS not valid: %v, input string: %v, using default: %v", err, leeway, 120)
+		} else {
+			iatLeeway = int64(value)
+		}
+	}
+}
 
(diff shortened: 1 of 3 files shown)


phlogistonjohn:
Before I look at the code, a thought: Should we make this a configurable behavior? Would there be users of heketi out there who would prefer to suffer from the "token used before issued" errors at the expense of having potentially improved security?

middleware/jwt_test.go
```diff
@@ -394,6 +396,77 @@ func TestJwt(t *testing.T) {
 	tests.Assert(t, r.StatusCode == http.StatusOK)
 	tests.Assert(t, called == true)
 }
+func TestJwtIgnoreIAT(t *testing.T) {
```
phlogistonjohn: I guess gofmt doesn't care about this, but I do: there should be an empty line between the closing curly brace of the previous function and the start of the new one.

middleware/jwt.go
```diff
@@ -25,12 +26,44 @@ import (
...
+	vErr := new(jwt.ValidationError)
+	now := jwt.TimeFunc().Unix()
+
+	if c.VerifyExpiresAt(now, true) == false {
+		delta := time.Unix(now, 0).Sub(time.Unix(c.ExpiresAt, 0))
+		vErr.Inner = fmt.Errorf("Token is expired by %v", delta)
+		vErr.Errors |= jwt.ValidationErrorExpired
+	}
+
+	// We skip the check for "iat"
```
phlogistonjohn: We should either eliminate, modify, or conditionalize this block but we should not leave it just as a comment.

raghavendra-talur:
@phlogistonjohn Changes done as requested. Also, I found a cool way to test ENV based configuration :  https://medium.com/@povilasve/go-advanced-tips-tricks-a872503ac859. Check new tests added.

middleware/jwt.go
```diff
@@ -15,22 +15,76 @@ import (
...
+	"github.com/heketi/heketi/pkg/utils"
 )
 
+var (
+	logger          = utils.NewLogger("[jwt]", utils.LEVEL_DEBUG)
+	iatLeeway int64 = 120
+)
+
+func init() {
+	if leeway, exists := os.LookupEnv("HEKETI_JWT_IAT_LEEWAY_SECONDS"); exists {
```
Madhu-1: I have a question, what happens if we set ```HEKETI_JWT_IAT_LEEWAY_SECONDS``` as a negative value.

middleware/jwt.go
```diff
@@ -15,22 +15,76 @@ import (
...
+		vErr.Inner = fmt.Errorf("Token is expired by %v", delta)
+		vErr.Errors |= jwt.ValidationErrorExpired
+		logger.LogError("exp validation failed: %v", vErr.Error())
+	}
+
+	// "iat" check
+	if now < c.IssuedAt-iatLeeway {
+		vErr.Inner = fmt.Errorf("Token used before issued")
+		vErr.Errors |= jwt.ValidationErrorIssuedAt
+		logger.LogError("iat validation failed: %v, time now: %v, time issued: %v", vErr.Error(), now, c.IssuedAt)
```
Madhu-1: ```IssuedAt``` and ```now``` are both in Unix form.
can we convert back the Unix time to UTC time for more readability on logging

phlogistonjohn:
retest this please

middleware/jwt.go
raghavendra-talur: done

middleware/jwt.go
raghavendra-talur: Handled now and added a test.

middleware/jwt_test.go
```diff
@@ -391,10 +390,414 @@ func TestJwt(t *testing.T) {
 	req.Header.Set("Authorization", "bearer "+tokenString)
 	r, err := http.DefaultClient.Do(req)
 	tests.Assert(t, err == nil)
-	tests.Assert(t, r.StatusCode == http.StatusOK)
+	tests.Assert(t, r.StatusCode == http.StatusOK, r.StatusCode, r.Status)
+	tests.Assert(t, called == true)
+}
+func TestJwtLeewayIAT(t *testing.T) {
```
phlogistonjohn: Missing blank line between closing curly and new function?

PR #1306, opened 2018-08-08
Support formatting bricks in a brick set differently based on brick properties

### LINKED ISSUE #1296: Consider different xfs formatting options for arbiter brick
### Kind of issue

Feature request

### Description

Currently the xfs file system written for the arbiter brick of an arbiter volume uses the exact same options as a normal brick. As this volume is treated specially it may make sense to format it specially too. We would want to optimize the filesystem for creating directories and inodes at the expense of data (of which there should be little to none).

From a quick skim of the manpage it seems that the option `-i maxpct=0` reserves no space for data:
```
    Setting the value to 0 means that essentially all of  the
    filesystem  can  become  inode blocks, subject to inode32
    restrictions.
```
Which seems like it would make sense for an arbiter brick. We should investigate if there are any other options we should set for the brick fs as well.

phlogistonjohn:

Support formatting arbiter bricks differently than normal bricks (xfs options) with an eye towards making it possible to make other formatting differences in the future.

### Does this PR fix issues?

Fixes #1296 

### Notes for the reviewer

This is a collaboration of @raghavendra-talur and me, some of these patches should probably eventually get squashed, but I have not done so yet.


phlogistonjohn:
TODO: Add functional test to verify the formatting differences on the brick file systems.

apps/glusterfs/brick_allocate.go
```diff
@@ -521,6 +523,8 @@ func (bp *StandardBrickPlacer) Replace(
 	newDevices := make([]*DeviceEntry, bs.SetSize)
 	for i := 0; i < bs.SetSize; i++ {
 		if i == index {
+			if index == arbiter_index {
```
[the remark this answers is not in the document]
phlogistonjohn: np. Like I mentioned in the desc. this will need to be squshed and cleaned up a bit prior to merging.

raghavendra-talur:
Changes look good to me as per our discussion. @obnoxxx and @Madhu-1 have a look.

phlogistonjohn:
Centos-ci is failing in the same place. That's suspicious, I will investigate.

apps/glusterfs/brick_allocate.go
```diff
@@ -204,6 +204,8 @@ func tryAllocateBrickOnDevice(
 		logger.Debug(
 			"Unable to place a brick of size %v & factor %v on device %v",
 			brickSize, snapFactor, device.Info.Id)
+	} else {
+		brick.SubType = NormalSubType
```
obnoxxx: This seems to be an initialization to a default value. Why initialize here instead of in `brick_entry.go:NewBrickEntry()`?

phlogistonjohn: We could set the sub type in NewBrickEntry but the only code that "knows" the correct subtype is the placer. Normal is not the default, Unknown sub type is the default value. 

The lifecycle of the new field goes something like:
* All old entries pulled from the db will get an "unknown" sub type
* Placers that create a new brick entry set the correct sub type
* The new sub type values will get stored in the db
* The create brick code path uses the value to determine the formatting for new brick fs

This approach was designed not to need a specific upgrade function. However, adding a flag to indicate that this db can contain subtypes seems reasonable.

phlogistonjohn:
DB feature flag added.

raghavendra-talur:
@obnoxxx @phlogistonjohn 
I guess we could have skipped adding a DB feature flag here as the default value of 0 has no meaning. I don't mind having it though.

Changes look good to me.

phlogistonjohn:
@obnoxxx  have I sufficiently handled your requests?

apps/glusterfs/brick_allocate.go
obnoxxx: ok. well then!

phlogistonjohn:
retest this please

PR #1308, opened 2018-08-09
docker: use a non-blocking flock in heketi-start.sh

phlogistonjohn:

The startup script for the heketi container image used a blocking flock with a timeout. Normally, this would be the best way to do it but as heketi db is frequently deployed on a gluster volume and current versions of gluster may have issues with this use pattern we switch to polling the db with a non-blocking flock.

### Does this PR fix issues?

Works around issues in rhbz#1613260

### DIFF
--- a/extras/docker/fromsource/heketi-start.sh
+++ b/extras/docker/fromsource/heketi-start.sh
@@ -34,8 +34,18 @@ if [[ -d "${HEKETI_PATH}" ]]; then
     stat "${HEKETI_PATH}/heketi.db" | tee -a "${HEKETI_PATH}/container.log"
     # Workaround for scenario where a lock on the heketi.db has not been
     # released.
-    flock -w 60 "${HEKETI_PATH}/heketi.db" true
-    if [[ $? -ne 0 ]]; then
+    # This code uses a non-blocking flock in a loop rather than a blocking
+    # lock with timeout due to issues with current gluster and flock
+    # ( see rhbz#1613260 )
+    for _ in $(seq 1 60); do
+        flock --nonblock "${HEKETI_PATH}/heketi.db" true
+        flock_status=$?
+        if [[ $flock_status -eq 0 ]]; then
+            break
+        fi
+        sleep 1
+    done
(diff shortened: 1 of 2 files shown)


phlogistonjohn:
Update to make shellcheck happier.

phlogistonjohn:
FWIW the internal use of the db by bolt does appear to use flock, but its a non-blocking flock so we should not need to worry about the db use by the heketi binary itself.

```
openat(AT_FDCWD, "heketi.db", O_RDWR|O_CREAT|O_CLOEXEC, 0600) = 5
epoll_ctl(4, EPOLL_CTL_ADD, 5, {EPOLLIN|EPOLLOUT|EPOLLRDHUP|EPOLLET, {u32=485752688, u64=139883275616112}}) = -1 EPERM (Operation not permitted)
epoll_ctl(4, EPOLL_CTL_DEL, 5, 0xc42062b8dc) = -1 EPERM (Operation not permitted)
flock(5, LOCK_EX|LOCK_NB)               = 0
fstat(5, {st_mode=S_IFREG|0600, st_size=262144, ...}) = 0
```

extras/docker/fromsource/heketi-start.sh
```diff
@@ -34,8 +34,18 @@ if [[ -d "${HEKETI_PATH}" ]]; then
     stat "${HEKETI_PATH}/heketi.db" | tee -a "${HEKETI_PATH}/container.log"
     # Workaround for scenario where a lock on the heketi.db has not been
     # released.
-    flock -w 60 "${HEKETI_PATH}/heketi.db" true
-    if [[ $? -ne 0 ]]; then
+    # This code uses a non-blocking flock in a loop rather than a blocking
+    # lock with timeout due to issues with current gluster and flock
+    # ( see rhbz#1613260 )
```
obnoxxx: does this mean we are working around a (new?) bug in gluster here?
That's fine, but just want to make sure that this gets tracked as well....

phlogistonjohn: Yeah, we'll make sure a bug remains open for the gluster side of the issue.

PR #1311, opened 2018-08-10
Enable secure I/O communications (using TLS) on volume creation by default config or request field.

### LINKED ISSUE #934: Secure Heketi with TLS
Given growing concerns about wiretapping and multitenancy, many service providers expect to secure even internal networks. Could Heketi provide configuration to secure the connection with TLS? Otherwise it requires setting up a secure reverse proxy in front just for this purpose.

/cc @jarrpa

xakep666:
It allows to enable I/O path encryption for newly created volumes (gluster feature).
Actually main change is setting gluster volume options according to request/config.
Preparations needed before using this option: https://gluster.readthedocs.io/en/latest/Administrator%20Guide/SSL/.

### Notes for the reviewer
I didn`t write any tests.


phlogistonjohn:
PR is very stale. Closing.

PR #1314, opened 2018-08-11
docker: load an initial topology if HEKETI_TOPOLOGY_FILE is set

nixpanic:

By providing the environment variable HEKETI_TOPOLOGY_FILE with a
filename that is available in the container, the database gets populated
with this topology file.

In Kubernetes environments the HEKETI_TOPOLOGY_FILE can be provided as a
ConfigMap or Secret and mounted as a volume in the pods specification.
With this change it is now possible to deploy Heketi with a .yaml file
that contains the topology.json and the Heketi Deployment.

### Notes for the reviewer

This has been [tested within minikube](https://github.com/nixpanic/minikube/blob/addons/gluster/README.storage-provisioner-gluster.md) and a [temporary fork for the container build](https://quay.io/repository/nixpanic/minukube-heketi).

### DIFF
--- a/extras/docker/fromsource/heketi-start.sh
+++ b/extras/docker/fromsource/heketi-start.sh
@@ -1,4 +1,8 @@
 #!/bin/bash
+#
+# HEKETI_TOPOLOGY_FILE can be passed as an environment variable with the
+# filename of the initial topology.json. In case the heketi.db does not exist
+# yet, this file will be used to populate the database.
 
 : "${HEKETI_PATH:=/var/lib/heketi}"
 : "${BACKUPDB_PATH:=/backupdb}"
@@ -74,4 +78,33 @@ if [[ -d "${BACKUPDB_PATH}" ]]; then
     fi
 fi
 
-/usr/bin/heketi --config=/etc/heketi/heketi.json
+# if the heketi.db does not exist and HEKETI_TOPOLOGY_FILE is set, start the
+# heketi service in the background and load the topology. Once done, move the
+# heketi service back to the foreground again.
+if [[ "$(stat -c %s ${HEKETI_PATH}/heketi.db)" == 0 && -n "${HEKETI_TOPOLOGY_FILE}" ]]; then
+    # start hketi in the background
+    /usr/bin/heketi --config=/etc/heketi/heketi.json &
+
+    # wait until heketi replies
+    while ! curl http://localhost:8080/hello; do
+        sleep 0.5
+    done
+
+    # load the topology
+    if [[ -n "${HEKETI_ADMIN_KEY}" ]]; then
+        HEKETI_SECRET_ARG="--secret='${HEKETI_ADMIN_KEY}'"
+    fi
+    heketi-cli --user=admin "${HEKETI_SECRET_ARG}" topology load --json="${HEKETI_TOPOLOGY_FILE}"
+    if [[ $? -ne 0 ]]; then
+        # something failed, need to exit with an error
+        kill %1
+        echo "failed to load topology from ${HEKETI_TOPOLOGY_FILE}"
+        exit 1
+    fi
+
+    # bring heketi back to the foreground
+    fg %1
+else
+    # just start in the foreground
+    exec /usr/bin/heketi --config=/etc/heketi/heketi.json
+fi
--- a/extras/docker/rpi/heketi-start.sh
+++ b/extras/docker/rpi/heketi-start.sh
@@ -1,4 +1,8 @@
 #!/bin/bash
+#
+# HEKETI_TOPOLOGY_FILE can be passed as an environment variable with the
+# filename of the initial topology.json. In case the heketi.db does not exist
+# yet, this file will be used to populate the database.
 
 : "${HEKETI_PATH:=/var/lib/heketi}"
 : "${BACKUPDB_PATH:=/backupdb}"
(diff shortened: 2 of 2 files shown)


nixpanic:
[[epic-ocs-developer-environment] Automatically load topology when Heketi pod starts](https://trello.com/c/cYDzfItF/68-epic-ocs-developer-environment-automatically-load-topology-when-heketi-pod-starts)

extras/docker/fromsource/heketi-start.sh
```diff
@@ -28,6 +28,12 @@ if [[ -d "${HEKETI_PATH}" ]]; then
                 fi
                 ((check+=1))
             done
+        else
+            # database or heketidbstorage do not exist yet, maybe we need to
+            # populate it?
+            if [ -e "${HEKETI_TOPOLOGY_FILE}" ]; then
+                NEED_TOPOLOGY_LOAD=1
+            fi
```
phlogistonjohn: Why does this block appear in two places? Can we consolidate this in one place? (for DRY reasons)

extras/docker/fromsource/heketi-start.sh
```diff
@@ -74,4 +85,32 @@ if [[ -d "${BACKUPDB_PATH}" ]]; then
...
+# load the topology.json from the path in HEKETI_TOPOLOGY_FILE
+if [ -z "${NEED_TOPOLOGY_LOAD}" ]; then
+    # by default just start in the foreground
+    /usr/bin/heketi --config=/etc/heketi/heketi.json
+else
+    # start hketi in the background
+    /usr/bin/heketi --config=/etc/heketi/heketi.json &
+
+    # wait until heketi replies
+    while ! curl http://localhost:8080/hello; do
```
phlogistonjohn: Do we want to suppress curl's (standard) output here?

Also, heketi's port is configurable in theory. It's generally not changed much though. Any thoughts about that?

extras/docker/fromsource/heketi-start.sh
```diff
@@ -74,4 +85,32 @@ if [[ -d "${BACKUPDB_PATH}" ]]; then
     fi
 fi
 
-/usr/bin/heketi --config=/etc/heketi/heketi.json
+# if NEED_TOPOLOGY_LOAD is not empty, start the service in the background, and
+# load the topology.json from the path in HEKETI_TOPOLOGY_FILE
+if [ -z "${NEED_TOPOLOGY_LOAD}" ]; then
```
phlogistonjohn: We should use bash-style double brackets to be consistent with the rest of the script.

extras/docker/fromsource/heketi-start.sh
```diff
@@ -74,4 +85,32 @@ if [[ -d "${BACKUPDB_PATH}" ]]; then
...
+    /usr/bin/heketi --config=/etc/heketi/heketi.json &
+
+    # wait until heketi replies
+    while ! curl http://localhost:8080/hello; do
+        sleep 0.5
+    done
+
+    # load the topology
+    if [ -n "${HEKETI_ADMIN_KEY}" ]; then
+        HEKETI_SECRET_ARG="--secret='${HEKETI_ADMIN_KEY}'"
```
phlogistonjohn: Since this is bash, perhaps a bash array would be cleaner for the args.

extras/docker/fromsource/heketi-start.sh
```diff
@@ -74,4 +85,32 @@ if [[ -d "${BACKUPDB_PATH}" ]]; then
...
+    fi
+    heketi-cli --user=admin "${HEKETI_SECRET_ARG}" topology load --json="${HEKETI_TOPOLOGY_FILE}"
+    if [ $? -ne 0 ]; then
+        # something failed, need to exit with an error
+        kill %1
+	echo "failed to load topology from ${HEKETI_TOPOLOGY_FILE}"
+	exit 1
```
phlogistonjohn: These lines appear to be mixing spaces to indent with tabs. We should be consistent with the rest of the file (spaces).

extras/docker/fromsource/heketi-start.sh
nixpanic: This is where the directory for the db exists, the other one does not have the directory yet. I could probably move it out of this if/else completely and add a single "if heketi.db does not exist".

extras/docker/fromsource/heketi-start.sh
nixpanic: Ok, I can change that.

extras/docker/fromsource/heketi-start.sh
nixpanic: I thought about it, but it would be good to have some logging in case this decides to hang.

Is there a good way to find out what the port is configured to? Can heketi-cli pick that from the heketi.json maybe?

extras/docker/fromsource/heketi-start.sh
nixpanic: Uh, thats possible, but maybe a little less common to do in shell scripts.

extras/docker/fromsource/heketi-start.sh
nixpanic: Oh, of course!

extras/docker/fromsource/heketi-start.sh
phlogistonjohn: I think that would be cleaner.

extras/docker/fromsource/heketi-start.sh
phlogistonjohn: > I thought about it, but it would be good to have some logging in case this decides to hang.

OK, understood. It's not like this code path will run frequently.

> Is there a good way to find out what the port is configured to? Can heketi-cli pick that from the heketi.json maybe?

That's a great idea! :-)

extras/docker/fromsource/heketi-start.sh
phlogistonjohn: I think it would be cleaner, but I won't require it for merge.

extras/docker/fromsource/heketi-start.sh
raghavendra-talur: NEED_TOPOLOGY_LOAD needs to be in lower case and local to the script.

extras/docker/fromsource/heketi-start.sh
nixpanic: This has been done.

extras/docker/fromsource/heketi-start.sh
nixpanic: No change here. It would be part of a different Issue/PR. If you want this changed, maybe create an Issue for it and assign it to me :)

extras/docker/fromsource/heketi-start.sh
nixpanic: This should have been addressed now.

extras/docker/fromsource/heketi-start.sh
nixpanic: I did not change it to arrays, those are almost never used in bash scripts that I have seen.

extras/docker/fromsource/heketi-start.sh
nixpanic: Done!

extras/docker/fromsource/heketi-start.sh
phlogistonjohn: Certainly, I woudln't expect that to be resolved by this pr.

phlogistonjohn:
@nixpanic the current version of this script is failing ShellCheck. Can you please investigate?

nixpanic:
The `shellcheck` issue has been fixed now, there was an other error in the script that I have addressed as well. With the latest 'simplification', it was not taken into account that `flock` would create a file if it does not exist yet. Now not checking for the existence of the db, but rather verifying if the size of the db is 0.

extras/docker/fromsource/heketi-start.sh
obnoxxx: it seems you have not seen too many bash scripts ... ;-)

PR #1315, opened 2018-08-13
makefile: stop building darwin binaries as part of the release

phlogistonjohn:
It is highly unlikely that anyone is using the darwin builds
of heketi, especially that of the server. If there are any
these builds can be done from source by the interested parties.


Makefile
```diff
@@ -149,12 +149,12 @@ linux_arm_dist:
 linux_arm64_dist:
 	GOOS=linux GOARCH=arm64 $(MAKE) dist
 
-darwin_amd64_dist:
-	GOOS=darwin GOARCH=amd64 $(MAKE) dist
+# NOTE: if you need to build binaries for darwin or any other platform
+# golang supports just run: make dist GOOS=myos GOARCH=myarch
```
obnoxxx: This sentence does not seem to be quite right "... golang supports just run ...". do you mean "golang supports **to** just run" ?

obnoxxx: Oh! Ok, that makes sense. I just did not parse it that way. :-)
Will you push an update with the comma?
If you don't have time, I could also do it ;-)

phlogistonjohn: I reworded the whole thing to try and make it clearer.

PR #1316, opened 2018-08-13
Attempt to unify delivery of container files

phlogistonjohn:

This change restructures how some of the container related files are delivered. It adds a 'make install' target that installs the heketi and heketi-cli binaries along with the manpage and some supporting files. The fromsource container builds use these files rather than pull them directly through the Dockerfile.

The end goal is to unify how these heketi source files are delivered so that all sorts of container builds can use the same files, rather than making private copies that easily get out of sync.


extras/docker/rpi/Dockerfile
```diff
@@ -7,8 +7,8 @@ LABEL description="Heketi Container for Raspberry Pi"
 
 ADD ./heketi /usr/bin/heketi
 ADD ./heketi-cli /usr/bin/heketi-cli
-ADD ./heketi.json /etc/heketi/heketi.json
-ADD ./heketi-start.sh /usr/bin/heketi-start.sh
+ADD ../../container/heketi.json /etc/heketi/heketi.json
```
nixpanic: This only works when the build context is set to the `../..` directory. Maybe it would be good to add the buildah command as a comment in the header of this Dockerfile?

How does `./heketi` and `./heketi-cli` get in this directory?

This Dockerfile probably needs more corrections (maintainer/email), but that can be done in a different PR as well.

phlogistonjohn: > This only works when the build context is set to the ../.. directory. Maybe it would be good to add the buildah command as a comment in the header of this Dockerfile?

I think both buildah/docker build make the directiory the Dockerfile resides in the cwd automatically. I can try to verify, but if that didn't work other existing parts of the Dockerfile would have already been broken.

> How does ./heketi and ./heketi-cli get in this directory?

It's pretty hokey. If you look at the README file in the rpi/ directory it tells you to run the build-rpi-dockerfile.sh file. This script builds the heketi binaries outside the container and then runs docker build.
To be honest, I don't really care much about supporting heketi on rasberry pi and don't plan to expend any extra effort to improve this process. I do care about not having multiple copies of the same script in the source tree though. Otherwise, I would have happily ignored this subdir. :-)

> This Dockerfile probably needs more corrections (maintainer/email), but that can be done in a different PR as well.

Indeed. The regular fromsource one does as well.

phlogistonjohn:
retest this please

(centos ci test timed out again)

PR #1317, opened 2018-08-13
tests: create a testutils package

phlogistonjohn:

To avoid future copy and paste of test code this PR moves and refactors code that should be reusable across golang functional test packages.

Right now the packages provides two features, one per file:
* Heketi server management for tests that need to stop and start heketi themselves
* Cluster setup and teardown based on the "smoke test" directory

### Notes for the reviewer

For now I've let the some of the commits demonstrate what was changed from when the code was copied from the uniqe source file. This may need to be squashed.

### DIFF
--- a/pkg/testutils/cluster.go
+++ b/pkg/testutils/cluster.go
@@ -26,76 +26,61 @@ import (
 	"github.com/heketi/tests"
 )
 
-var (
-	// The heketi server must be running on the host
+type ClusterEnv struct {
+	Nodes   []string
+	SSHPort string
+	Disks   []string
 
-	// VMs
-	storage0    = "192.168.10.100"
-	storage1    = "192.168.10.101"
-	storage2    = "192.168.10.102"
-	storage3    = "192.168.10.103"
-	portNum     = "22"
-	storage0ssh = storage0 + ":" + portNum
-	storage1ssh = storage1 + ":" + portNum
-	storage2ssh = storage2 + ":" + portNum
-	storage3ssh = storage3 + ":" + portNum
+	HeketiUrl string
+}
 
-	// Heketi client
-	heketiUrl = "http://localhost:8080"
-	heketi    = client.NewClientNoAuth(heketiUrl)
-
-	// Storage systems
-	storagevms = []string{
-		storage0,
-		storage1,
-		storage2,
-		storage3,
-	}
-
-	// Disks on each system
-	disks = []string{
-		"/dev/vdb",
-		"/dev/vdc",
-		"/dev/vdd",
-		"/dev/vde",
-
-		"/dev/vdf",
-		"/dev/vdg",
-		"/dev/vdh",
-		"/dev/vdi",
-	}
-)
-
-func setupCluster(t *testing.T, numNodes int, numDisks int) {
-	tests.Assert(t, heketi != nil)
-
-	// Get ssh port first, we need it to create
-	// storageXssh variables
+func (ce *ClusterEnv) Update() {
+	// update the port from the env
 	env := os.Getenv("HEKETI_TEST_STORAGEPORT")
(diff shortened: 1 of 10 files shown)


phlogistonjohn:
Rebased & fixed Go 1.8 compile error

phlogistonjohn:
retest this please

phlogistonjohn:
@obnoxxx @raghavendra-talur PTAL, getting this reviewed will help both the openshift-ci effort and the effort towards #1401.

pkg/testutils/cluster.go
```diff
@@ -0,0 +1,202 @@
...
+	// Verify that there are no outstanding operations on the
+	// server. A test that needs to mess with the operations _must_
+	// clean up after itself.
+	oi, err := heketi.OperationsInfo()
+	tests.Assert(t, err == nil, "expected err == nil, got:", err)
+	tests.Assert(t, oi.Total == 0, "expected oi.Total == 0, got", oi.Total)
+	tests.Assert(t, oi.InFlight == 0, "expected oi.InFlight == 0, got", oi.Total)
+
+	// Create a cluster
+	cluster_req := &api.ClusterCreateRequest{
```
Madhu-1: to make `Setup` function more useful, how about making `Block` and `File` as input parameters?

pkg/testutils/serverctl.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+	go func() {
+		s.cmdErr = s.cmd.Wait()
+		s.cmdExited = true
+	}()
+	time.Sleep(300 * time.Millisecond)
+	if !s.IsAlive() {
+		return errors.New("server exited early")
+	}
+	// dump some logs if heketi fails to start?
```
Madhu-1: question: do we need this comment here?

pkg/testutils/serverctl.go
```diff
@@ -0,0 +1,179 @@
...
+	if err := s.cmd.Process.Signal(os.Interrupt); err != nil {
+		return err
+	}
+	time.Sleep(100 * time.Millisecond)
+	if !s.cmdExited {
+		if err := s.cmd.Process.Kill(); err != nil {
+			return err
+		}
+	}
+	s.logF.Close()
```
Madhu-1: question:
don't we need to call `Close()` in case of failure?

pkg/testutils/cluster.go
```diff
@@ -0,0 +1,202 @@
...
+			ce.Nodes[i] = value
+		}
+	}
+}
+
+func (ce *ClusterEnv) client() *client.Client {
+	return client.NewClientNoAuth(ce.HeketiUrl)
+}
+
+func (ce *ClusterEnv) SshHost(index int) string {
```
Madhu-1: As this is an Exposed function can we have a check for the slice length to avoid `index out of range error`?

Madhu-1:
@phlogistonjohn Centos CI is failing,

pkg/testutils/cluster.go
phlogistonjohn: Because this is meant for test code and I'm OK with panics in test code I think I'd like to leave it as-is for now to retain a simpler interface.

pkg/testutils/serverctl.go
phlogistonjohn: No. Will clean up.

phlogistonjohn:
updates pushed

pkg/testutils/cluster.go
obnoxxx: yeah, I think we can further generalize later..

pkg/testutils/cluster.go
obnoxxx: good point, @Madhu-1 , but I'm ok with John's justification.

PR #1323, opened 2018-08-17
apps:glusterfs fix logger format for go 1.11

sigma:

DeviceInfo.Id is actually a string, which makes the %d formatter an error with upcoming go 1.11.
Use %v for consistency.

### Does this PR fix issues?

Fixes #


phlogistonjohn:
ok to test

PR #1326, opened 2018-08-22
[RFC] stop testing go 1.7.5 as part of travis ci

phlogistonjohn:
Go version 1.7.5 is older than any version of go on platforms
we normally ship on and it should not be needed to regularly
test on this version any more.

### Notes for the reviewer

Please do not merge until we've determined the right time to drop support for this version of Go.


obnoxxx:
I have reverted this now. Is there a way to re-open this PR in github or do we need to create a new one? ....

obnoxxx:
revived as #1337

PR #1334, opened 2018-08-27
Add a Trace() function to the utils package.

ansiwen:

The Trace() function returns the name of the calling function, the
name of the calling file and the line number of the calling line.
The convenience function TraceFunc() only returns the function name.

* The logging now uses this function for getting filename and line number.
* The long logging format now logs the function name additionally to the file name and line number.
* In order to improve readability, the filenames are shortened by omitting the redundant "/src/github.com/heketi/" prefix.


phlogistonjohn:
I'm not sure how this is meant to be used. How is it better than the file:lineno logging that the system already does?

ansiwen:
It returns the name of the function, not only the filename and line number?

I'm using it with `logger.Debug(utils.TraceFunc())` like here:
https://github.com/ansiwen/heketi/commit/a8fbc9324816869da3dd244ac870da03cc404ec8

It results in output like: 
```
[glusterd] DEBUG 2018/08/30 14:20:18 /src/github.com/heketi/heketi/executors/glusterd/device.go:48: github.com/heketi/heketi/executors/glusterd.(*executor).DeviceSetup
```

We could think about adding the function name to the log output in general.

pkg/utils/log.go
```diff
@@ -51,10 +51,13 @@ func logWithLongFile(l *log.Logger, format string, v ...interface{}) {
 	// From
 	// /builddir/build/BUILD/heketi-3f4a5b1b6edff87232e8b24533c53b4151ebd9c7/src/github.com/heketi/heketi/apps/glusterfs/volume_entry.go
 	// to
-	// src/github.com/heketi/heketi/apps/glusterfs/volume_entry.go
-	i := strings.Index(file, "/src/")
+	// apps/glusterfs/volume_entry.go
+	const basePath = "github.com/heketi/"
```
phlogistonjohn: Does this need to deal with paths outside of "github.com/heketi"? Or is that just not an issue for this code?

ansiwen: Are you aware of any place outside of the heketi codebase, where pkg/utils/log is being used? that would surprise me. And even if, it would just not shorten the path end print the full path. I was just annoyed by the fact that I hade lot's of useless src/github.com/heketi/heketi/ in my terminal window.

phlogistonjohn: I'm +1 to making the heketi paths shorter. I don't think this version of the heketi utils logger will be used outside of heketi code, so we're probably safe. I just wanted to see if you had thought about the issue. The slightly longer story is that there's a copy of heketi utils in its own standalone repo that could in fact be used by other stuff... but since I started working on heketi we've never tried to keep that code in sync, so I am not really worried about it.

phlogistonjohn:
@obnoxxx @raghavendra-talur any thoughts? This seems like a reasonable refactor, but I don't want to be the only review here. :-)

pkg/utils/trace.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"path/filepath"
+	"runtime"
+)
+
+// TraceSkip is like Trace, but additionally skips stack frames. So TaceSkip(0)
+// behaves like Trace(), TraceSkip(1) refers to the call of the current
+// function, TraceSkip(2) refers to the call of the calling function, and so on.
+func TraceSkip(skip int) (string, string, int) {
+	pc := make([]uintptr, 15)
```
raghavendra-talur: Is there a check required here? skip is less than or equal to 15?

pkg/utils/log.go
raghavendra-talur: IMO, the path had src/github.com/heketi/heketi because it is possible that heketi is built from a different Source control repo. That possibility is small, I think this is fine.

pkg/utils/trace.go
raghavendra-talur: I would like an explanation for this question before the patch is merged.

obnoxxx: @raghavendra-talur  if you want to block this PR until this is solved, you should raise a review change request. :-)

raghavendra-talur: @obnoxXx @ansiwen thanks!

PR #1335, opened 2018-08-29
Restrict usage of some block volumes: by admin or after upgrade

phlogistonjohn:

In order to protect against creating block volumes on existing block-hosting-volumes that may be in a bad state or unwanted for new allocations this PR defines a new concept: block restriction. The block restriction is a new api type on BHVs that can be set to locked or cleared by admins. The system may also set a special locked state that indicates the BHV was not cleanly upgraded from previous versions.
If the admin tries to clear one of these restrictions the system checks that the BHV is in an eligible  state.

Tests for the lockout condition, setting the restriction on BHVs, the new client api call, and upgrade conditions are included.

### Does this PR fix issues?

Fixes [rhbz#1622645](https://bugzilla.redhat.com/show_bug.cgi?id=1622645)

### Notes for the reviewer

While this PR does not directly fix issues with gluster-block on already filled hosting volumes it does prevent the problem from getting worse while providing a stepping stone for more automated actions around the health and management of block hosting volumes.


apps/glusterfs/operations_block_restriction.go
```diff
@@ -0,0 +1,164 @@
...
+func (ro *VolumeSetBlockRestrictionOperation) Rollback(executor executors.Executor) error {
+	return nil
+}
+
+func (ro *VolumeSetBlockRestrictionOperation) checkCanUnlock(
+	v *VolumeEntry, r api.BlockRestriction) error {
+
+	switch {
+	case v.Info.BlockInfo.Restriction == r:
+		// no change is always OK
```
raghavendra-talur: I like reading switch states where cases of are of same type. Hence, if we could have all current states handles first it would be nice. Like:
a. current state is locked and request is for...
b. current state is "" and request is for...
c. current state is "locked-by-update" and request is for...

Whatever checks remain apart from the above ones can come in later in the validation.

apps/glusterfs/dbcommon.go
```diff
@@ -141,6 +143,13 @@ func UpgradeDB(tx *bolt.Tx) error {
 		return err
 	}
 
+	err = fixBlockHostingReservedSize(tx)
```
raghavendra-talur: I think we missed it in previous patches but would like to have updates with respect a particular entry within the function. Could we move this under VolumeEntryUpgrade() please.

apps/glusterfs/operations_block_restriction.go
phlogistonjohn: Updated.

apps/glusterfs/dbcommon.go
raghavendra-talur: @phlogistonjohn and I discussed about this. It is actually better this way. We should not add any upgrade functions in the middle. Some of the later upgrade functions might have a prior upgrade function as a prerequisite. Keep it this way.

PR #1338, opened 2018-09-02
Add volume protection option

rsevilla87:
This PR gives the option to protect a volume from an accidental deletion. If for some reason an admin user doesn't want a certain volume to be deleted by Heketi, this option adds an additional confirmation layer that can be very useful to avoid unexpected data loss of critical volumes.


phlogistonjohn:
@rsevilla87 interesting stuff! Recently we added a restrictions feature to block hosting volumes. Could you please look at that code and see if these is a way we could harmonize what you're doing here and that code? I ask because I am hoping to avoid having many different approaches to similar things in the code base.

Thank you!

PR #1339, opened 2018-09-03
Added a '--cluster' boolean flag to the device resync command to have the ability to resync all devices owned by a cluster

mcaimi:
Hello,
I added a "topology resync" command to the heketi-cli binary to easily let users resync all devices inside the cluster.


phlogistonjohn:
So, first off apologies for overlooking your contribution for so long.

Looking over the changes there are a few high-level items I tihnk we need to sort out:
* First, the PR needs a rebase against master
* The name is a bit confusing. Because it's still actually re-syncing the devices, just the devices in the cluster
* Maybe we should enhance the device resync command to take --cluster or --node options that tell it to resync all the devices under the given cluster id or node id, rather then just create a new command?

mcaimi:
No worries, thank you.
That makes a lot of sense, I'll do a rebase and look into it

mcaimi:
I did a rebase and moved the feature to the 'device resync' command.
I added a boolean flag '--cluster' to the command to instruct heketi-cli to use the first argument as the cluster id to sync rather than the id of a specific device.

The default behaviour is still the same. By default a single device is resynced:

$ heketi-cli device resync <device_id> # only this device gets resynced

but when '--cluster' is specified on the command line, the device_id is actually the cluster_id:

$ heketi-cli device resync --cluster <cluster_id> # every device in every node owned by this cluster_id gets resynced

Is this solution acceptable? Also, do I need to add a '--node' flag too?

client/cli/go/cmds/device.go
```diff
@@ -322,30 +324,83 @@ var deviceDisableCommand = &cobra.Command{
...
 			return err
 		}
 
-		//set url
-		err = heketi.DeviceResync(deviceId)
-		if err == nil {
-			fmt.Fprintf(stdout, "Device %v updated\n", deviceId)
+		// a single device Id was specified and no cluster resync was requested
+		if resyncCluster == false {
+			fmt.Println("No cluster-wide resync operation requested. Using object_id as device_id.")
```
phlogistonjohn: Because this is the default behavior I think we should drop this line.

client/cli/go/cmds/device.go
```diff
@@ -322,30 +324,83 @@ var deviceDisableCommand = &cobra.Command{
...
+			//set url
```
phlogistonjohn: This looks like a stray comment.

client/cli/go/cmds/device.go
```diff
@@ -322,30 +324,83 @@ var deviceDisableCommand = &cobra.Command{
...
+			err = heketi.DeviceResync(objectId)
+			if err == nil {
+				fmt.Fprintf(stdout, "Device %v updated\n", objectId)
+			}
+		} else { // requested cluster resync..
```
phlogistonjohn: Because this case becomes deeply nested I would like this to be split into a helper function.

client/cli/go/cmds/device.go
```diff
@@ -322,30 +324,83 @@ var deviceDisableCommand = &cobra.Command{
...
+			// RESYNC ALL
+			fmt.Println("Cluster-wide resync operation requested. Using object_id as cluster_id.")
```
phlogistonjohn: I like the first sentence, nice and clear. I don't the 2nd half is needed.

client/cli/go/cmds/device.go
```diff
@@ -322,30 +324,83 @@ var deviceDisableCommand = &cobra.Command{
...
+			for cluster, _ := range topoinfo.ClusterList {
+				thisCluster := topoinfo.ClusterList[cluster]
+				// check cluster id.
+				if thisCluster.Id == objectId {
+					// iterate over Nodes
+					for node, _ := range thisCluster.Nodes {
+						thisNode := thisCluster.Nodes[node]
+						// resync every Device in this Node
+						for _, device := range thisNode.DevicesInfo {
+							fmt.Fprintf(stdout, "\t\tDevice Id:%-35v"+
```
phlogistonjohn: I'm not sold on printing all of this state out before the device resync is called. Why not just print something like ("Resyncing device %v...", device.Id)

client/cli/go/cmds/device.go
mcaimi: Agree.

client/cli/go/cmds/device.go
mcaimi: Yes, it's definitely a stray comment, I'll remove that

client/cli/go/cmds/device.go
mcaimi: Yeah, I kind of liked some verbose printout but I see that it clobbers the output.
I will change that

client/cli/go/cmds/device.go
mcaimi: Another stray comment...
Definitely redundant, I'll remove the second line.

client/cli/go/cmds/device.go
mcaimi: I'll move all the 'cluster resync' logic in another function.
This function will take an heketi Client object and a cluster_id string as input parameters and then just resync all devices owned by that cluster_id.

mcaimi:
I addressed all changes you requested. Have a look!

Thanks!

client/cli/go/cmds/device.go
phlogistonjohn: Unfortunately, it still looks like it exists.

phlogistonjohn:
We're really really close to mergable state. Would you please squash your commits into a single commit? We don't want the full history of the PR in master and your changes are well sized for a single commit.

@raghavendra-talur PTAL

client/cli/go/cmds/device.go
mcaimi: I can't see it anymore, I checked in my fork too, it's not there anymore

phlogistonjohn: Strange. I see it when I view "all files" on the PR; I also see it when I pull the PR locally and look at the diff:
```
@@ -342,10 +384,18 @@ var deviceResyncCommand = &cobra.Command{
                        return err
                }
 
-               //set url
-               err = heketi.DeviceResync(deviceId)
-               if err == nil {
-                       fmt.Fprintf(stdout, "Device %v updated\n", deviceId)
+               if resyncCluster == false {
+                       //set url                                    <-------- here
+                       err = heketi.DeviceResync(objectId)
+                       if err == nil {
+                               fmt.Fprintf(stdout, "Device %v updated\n", objectId)
+                       }
```

Maybe you need to (force) push?

mcaimi: yeah maybe. I just squashed all commits into one and force-push it again.
let's see :)

mcaimi:
All commits squashed into one.

client/cli/go/cmds/device.go
phlogistonjohn: I still see it even in the squashed pr! It's right under the `if resyncCluster == false {` line

phlogistonjohn:
I'm not sure what the leading "[+]" characters in the commit message means. Heketi convention is "topic: description" where topic roughtly maps to directory. So I'd recommend something like: "cli: add a '--cluster' flag to the device resync command" with a short description of the --cluster flag below.

mcaimi:
Ok, re-submitted the PR with -hopefully- all requested changes baked in.
Also, changed the commit message to accomodate the Heketi convention.

Thank you!

client/cli/go/cmds/device.go
mcaimi: I re-submitted the PR. Do you see that still?

mcaimi:
Great thank you!

PR #1343, opened 2018-09-05
Fix  volume info to list only completed (not pending) block volumes

### LINKED ISSUE #1342: volume info lists pending block volumes
### Kind of issue

Bug

### Observed behavior

If there are pending block volume entries, then volume info on the block hosting volume still lists the IDs of those pending block volumes.

### Expected/desired behavior
volume info should only list completed block volumes.

* Create a situation where there is a pending block volume entry (slightly difficult, since this might be a race condition).
* Do `heketi cli volume info <ID>` on the ID of the block hosting volume.

- Heketi version used (e.g. v6.0.0 or master):  master
- Operating system used: linux / fedora 28
- Heketi compiled from sources, as a package (rpm/deb), or container:  sources, git

obnoxxx:

Fix the result of VolumInfo for block hosting volumes to spare pending block volumes.

### Does this PR fix issues?

Fixes #1342

### DIFF
--- a/apps/glusterfs/app_volume.go
+++ b/apps/glusterfs/app_volume.go
@@ -207,6 +207,11 @@ func (a *App) VolumeInfo(w http.ResponseWriter, r *http.Request) {
 			http.Error(w, err.Error(), http.StatusInternalServerError)
 			return err
 		}
+		err = UpdateVolumeInfoComplete(tx, info)
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return err
+		}
 
 		return nil
 	})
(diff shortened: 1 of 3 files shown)


phlogistonjohn:
@obnoxxx your test is intermittently failing because there are two volumes getting created but you only select volumes using `NewVolumeEntryFromId(tx, vids[0])` but ordering of vids is not deterministic. We should either stop creating more than one volume or make the volume selection more robust by picking the ID of the volume inside the Update callback.

obnoxxx:
@phlogistonjohn I found the same. The volume creation was unnecessary and just a carry-over from the test which inspired this one. I dropped it, now I only have the auto-created block-hosting volume and all is fine.

PR #1346, opened 2018-09-06
Fix error checking when deleting a deleted block volume

phlogistonjohn:

PR #1340 was incomplete and did not correctly handle the various error reporting modes of gluster-block. It turns out gluster block always prints its status on stdout but does not always exit with a non-zero code when an error is encountered. This PR aims to fix those oversights and add functional tests for these situations.

### DIFF
--- a/tests/functional/TestSmokeTest/tests/block_volume_test.go
+++ b/tests/functional/TestSmokeTest/tests/block_volume_test.go
@@ -13,6 +13,7 @@ package functional
 import (
 	"fmt"
 	"testing"
+	"time"
 
 	"github.com/heketi/heketi/pkg/glusterfs/api"
 	"github.com/heketi/heketi/pkg/utils/ssh"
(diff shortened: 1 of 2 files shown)


tests/functional/TestSmokeTest/tests/block_volume_test.go
```diff
@@ -133,3 +136,120 @@ func TestBlockVolumeCreateManyAtOnce(t *testing.T) {
...
+		err   error
+	}
+
+	bvol, err := heketi.BlockVolumeCreate(blockReq)
+	tests.Assert(t, err == nil, "expected err == nil, got:", err)
+
+	_, err = heketi.BlockVolumeInfo(bvol.Id)
+	tests.Assert(t, err == nil, "expected err == nil, got:", err)
+
+	//targetHost := storage0ssh
```
obnoxxx: remove comment?

tests/functional/TestSmokeTest/tests/block_volume_test.go
```diff
@@ -133,3 +136,120 @@ func TestBlockVolumeCreateManyAtOnce(t *testing.T) {
...
+
+	defer startServices()
+
+	l, err := heketi.BlockVolumeList()
+	tests.Assert(t, err == nil, "expected err == nil, got:", err)
+	tests.Assert(t, len(l.BlockVolumes) == 1,
+		"expected len(l.BlockVolumes) == 1, got:", len(l.BlockVolumes))
+
+	stopServices(false)
+	for i := 0; i < 10; i++ {
```
obnoxxx: Could you add a comment explaining the reason for doing a loop for 10 attempts here?

tests/functional/TestSmokeTest/tests/block_volume_test.go
```diff
@@ -133,3 +136,120 @@ func TestBlockVolumeCreateManyAtOnce(t *testing.T) {
...
+		err = heketi.BlockVolumeDelete(bvol.Id)
+		tests.Assert(t, err != nil, "expected err != nil, got:", err)
+	}
+
+	stopServices(true)
+	for i := 0; i < 10; i++ {
```
obnoxxx: and here?

PR #1348, opened 2018-09-07
Replace errors package with github.com/pkg/errors

ansiwen:

The errors we return and handle are forwarded through many levels of the function call stack, so often -- especially while debugging -- it can be very hard to see, where an error really is originating from. This is especially true if errors are forwarded through channels to different goroutines or asynchronously from 3rd party libraries.

In order to improve this situation the advice of [this](https://dave.cheney.net/2016/06/12/stack-traces-and-the-errors-package) blog article has been followed. That is the [github.com/pkg/errors](https://godoc.org/github.com/pkg/errors) package is used instead of the standard errors package throughout the code base, which annotates all errors with a stack trace, and also allows to wrap errors with new stack traces, in case of "stack-switches", like when they are transferred asynchronously over channels or come in from 3rd party libraries. The only "old" errors that are kept are the sentinel errors defined in apps/glusterfs/errors.go.

### DIFF
--- a/apps/glusterfs/app.go
+++ b/apps/glusterfs/app.go
@@ -16,13 +16,14 @@ import (
 
 	"github.com/boltdb/bolt"
 	"github.com/gorilla/mux"
+	"github.com/heketi/rest"
+	"github.com/pkg/errors"
+
 	"github.com/heketi/heketi/executors"
 	"github.com/heketi/heketi/executors/kubeexec"
 	"github.com/heketi/heketi/executors/mockexec"
 	"github.com/heketi/heketi/executors/sshexec"
 	"github.com/heketi/heketi/pkg/logging"
-	"github.com/heketi/rest"
-	"github.com/pkg/errors"
 )
 
 const (
--- a/apps/glusterfs/app.go
+++ b/apps/glusterfs/app.go
@@ -10,8 +10,6 @@
 package glusterfs
 
 import (
-	"errors"
-	"fmt"
 	"net/http"
 	"os"
 	"strconv"
@@ -24,6 +22,7 @@ import (
 	"github.com/heketi/heketi/executors/sshexec"
 	"github.com/heketi/heketi/pkg/logging"
 	"github.com/heketi/rest"
+	"github.com/pkg/errors"
 )
 
 const (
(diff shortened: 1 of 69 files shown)


apps/glusterfs/errors.go
```diff
@@ -10,26 +10,26 @@
 package glusterfs
 
 import (
-	"errors"
+	stdErrors "errors"
```
phlogistonjohn: I don't understand this change. There's no import of the enhanced errors package here.

phlogistonjohn:
Looks pretty good so far. A couple of points:

First, this is somewhat of a nit and I don't think I'd block the merge over this but I like a style of imports like:
```golang
import (
// std library imports go first
   "fmt"
   "io"
   "strings"

// non-local library imports go next
   "github.com/something/amazing"
   "github.com/cool/stuff/pkg/foobar"

// local (within the current repo last)
    "github.com/heketi/heketi/pkg/womble"
)
```

Some code in heketi follows this while some code doesn't. 

Second, like any change that touches glide I think we need to argue about the delta in glide lock. Can I assume that most of these changes to all the other imports were done automatically by glide?

apps/glusterfs/errors.go
ansiwen: I like to make it explicit here, that this is the only place where still the "old" errors package is used. Also it avoids, that when code is added to the file (like I did in follow up PRs) doesn't accidentally use the standard errors package.

ansiwen:
@phlogistonjohn: Isn't the import style enforced by the format tool? I didn't do anything on purpose here and just let the tool do it's work.

phlogistonjohn:
> Isn't the import style enforced by the format tool? I didn't do anything on purpose here and just let the tool do its work.

The tool only enforces that each group of import lines is in sorted order. Logically grouping them by separating them by empty lines is a convention I have seen elsewhere and liked.

phlogistonjohn:

ansiwen:
@phlogistonjohn I rebased, removed the stdErrors stuff and reorganized the import lines. And yes, the glide.lock changes were done automatically by `glide update -v`

phlogistonjohn:

PR #1349, opened 2018-09-10
Fix error message in glusterfs.DbCreate()

ansiwen:
The created error message in https://github.com/heketi/heketi/blob/master/apps/glusterfs/db_operations.go#L302 incorporates `err`, which refers to an `err` variable that is always `nil` by https://github.com/heketi/heketi/blob/master/apps/glusterfs/db_operations.go#L272-L275 .

I assume this is a copy and paste mistake and replaced it with the value of the `durability` variable, which this switch is about and hence makes more sense to report.


phlogistonjohn:
retest this please

PR #1350, opened 2018-09-10
Add more comprehensive results for command execution

phlogistonjohn:

When working on block volume delete it was discovered (or at least brought to the forefront) that Heketi's remote command execution functions were discarding stdout on command error and discarding stderr on command success. This isn't very unixy and quite unhelpful for handling the outputs of gluster-block. This PR reworks adds new types, "Result" and "Results" such that the code can return Results that have a lot more information about each command's result.

These new command running functions are plumbed through the various layers and used to clean up the block volume delete executor function (and demonstrate how they should be used).

### Notes for the reviewer

I originally tried to refactor the functions more but it turned into a lot of churn for little benefit. This PR is much more targeted so that it more immediately testable but also exists to demonstrate the direction I think we should aim for the executor and it's supporting packages.

### DIFF
--- a/executors/cmdexec/block_volume.go
+++ b/executors/cmdexec/block_volume.go
@@ -88,11 +88,23 @@ func (s *CmdExecutor) BlockVolumeDestroy(host string, blockHostingVolumeName str
 	godbc.Require(blockVolumeName != "")
 
 	commands := []string{
-		// this ugly hack exists so that heketi can extract the error message
-		// from stderr if the command exits non-zero but can scrape the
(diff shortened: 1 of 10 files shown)


phlogistonjohn:
retest this please

phlogistonjohn:
retest this please

phlogistonjohn:
Please note that the openshift ci failure happened because the test environment for openshift ci is not fully set up yet, this was just a convenient pr to check if the bot had access. Please ignore that test failure wrt to actually reviewing this pr. Work on the openshift ci will continue elsewhere.

PR #1354, opened 2018-09-14
block-volume-optin: add a new network option to block volume

amarts:
As identified in [glusterfs patch](https://review.gluster.org/#/c/glusterfs/+/21170/),
we need option 'server.user-tcp-timeout' to be set on gluster-block volume to mitigate
some of the 'hung status' like behavior when a node goes down. Adding this option on
top of the volume would solve the issue for all the released version of glusterfs,
where this patch is not present, and even in future, if the option is present,
it would serve as no-op, so no need to bother about this.

Signed-off-by: Amar Tumballi <amarts@redhat.com>

### DIFF
--- a/apps/glusterfs/block_settings.go
+++ b/apps/glusterfs/block_settings.go
@@ -14,5 +14,5 @@ var (
 	CreateBlockHostingVolumes = false
 	// Default 1 TB
 	BlockHostingVolumeSize    = 1024
-	BlockHostingVolumeOptions = "group gluster-block"
+	BlockHostingVolumeOptions = "group gluster-block, server.tcp-user-timeout 42"
 )


amarts:
> If I understand correctly, heketi should be setting the option because not all versions of gluster that heketi manages will have that option baked into the the "gluster-block" group, correct?

Yes, that is the exact reason for this patch! While upstream patch for glusterfs to make this option's value 42 is merged, it will land in a release v5.0 upstream, but there are so many existing gluster deployments which will be benefit from this option, specially in container workloads (as the option is useful only in n/w disconnections which is common in k8s setup than in normal on-premise glusterfs setups).

Hence this patch! Also note that, it won't regress even if this option is made default in glusterfs, it would be just a 'no-op' in this case.

phlogistonjohn:
@obnoxxx I'm trying to understand your suggestion better. Are you suggesting that the overall block hosting be set something like (pseduocode):
```
  AdditionalBlockHostingVolumeOptions = config_or_env(...).AdditionalBlockHostingVolumeOptions
  VolOptions = BlockHostingVolumeOption + AdditionalBlockHostingVolumeOptions
```
And that the server.tcp-user-timeout value get shipped in our default configuration?

obnoxxx:
> @obnoxxx  I'm trying to understand your suggestion better. Are you suggesting  that the overall block hosting be set something like (pseduocode): 
> ```
> AdditionalBlockHostingVolumeOptions = config_or_env(...).AdditionalBlockHostingVolumeOptions
> VolOptions = BlockHostingVolumeOption + AdditionalBlockHostingVolumeOptions
> ```
> And that the server.tcp-user-timeout value get shipped in our default configuration?

Yes.

obnoxxx:
@pkalever writes:
> IMO, tcp timeout should be set for all type of volumes irrespective of their purpose.

Sounds ok to me, even though this was brought up in the block-hosting volume context, it is applied to all kinds of volumes in gluster...

> @amarts @obnoxxx @phlogistonjohn how about this change ?
> 
> 0001-VolumeEntry-default-tcp-timeout-to-42sec-for-all-vol.patch.txt

First of all: what is this? How could you upload a patch file to the heketi org/repository???

Then, the same concern applies for me as with @amarts 's patch:

I would **not** like to hard-code such config options which might change later in gluster. I'd like to either have a config option for additional default volume options in heketi's config, or else have the server so clever that it can detect whether the option is set on the volume from gluster, and only if it is *not* set, then set this additional default....

phlogistonjohn:
> First of all: what is this? How could you upload a patch file to the heketi org/repository???

@obnoxxx fwiw when I initially read this it almost sounded like you were offended. Thinking about it for a moment I realized you were confused (as am I) as to what mechanism you used to create a patch file like that on a github issue.

> Then, the same concern applies for me as with @amarts 's patch:
> 
> I would not like to hard-code such config options which might change later in gluster. I'd like to either have a config option for additional default volume options in heketi's config, or else have the server so clever that it can detect whether the option is set on the volume from gluster, and only if it is not set, then set this additional default....

I agree with the general concept. Currently we create a volume by doing steps like:
* gluster volume create
* gluster volume add-brick (multiple, If needed)
* gluster volume set (multiple, if needed)
* gluster volume start

If we could be more clever wrt the volume set step, perhaps by checking on gluster's defaults first, and then conditionally calling another set command for the server.tcp-user-timeout option. Is this a valid approach?

obnoxxx:
@phlogistonjohn writes:
> I agree with the general concept. Currently we create a volume by doing steps like:
> 
> * gluster volume create
> * gluster volume add-brick (multiple, If needed)
> * gluster volume set (multiple, if needed)
> * gluster volume start
>
> If we could be more clever wrt the volume set step, perhaps by checking on gluster's defaults first, and then conditionally calling another set command for the server.tcp-user-timeout option. Is this a valid approach?

This would be perfect, IMO. Exactly what I had in mind.

phlogistonjohn:
Good, one remaining question I have then is how should we check for the condition? Can we just do an options list/get and set it if nothing else is specified?
This is where my gluster knowledge is a bit weak and I'd like to validate the approach with @amarts and @pkalever. If not, is there something else we can key off of?

obnoxxx:
@phlogistonjohn right, this is all dependent on the question of whether it can be done. :-) If not, we might still have to do the more simple overriding with a config variable.

raghavendra-talur:
@phlogistonjohn @obnoxxx @amarts @pkalever 

I see that the requirement is to add 'server.user-tcp-timeout' option to
1. existing and new volumes both
2. block hosting and file volumes both

Also, we have to restart bricks once the volume option is set on existing volumes.


Apart from the design reviews on existing PR and patch, my concern is that it does not apply to existing volumes.

I propose that we solve this issue in the gluster container. 

For existing volumes:
Gluster container should list all volumes and set the timeout for each volume and restart bricks.

For new volumes:
Gluster container should add a hook script that sets the option in either "post-create" or "pre-start" dir of hooks.

Ahhh, scratch the solution for new volumes, I forgot that we should not call gluster commands in hook scripts. I wish we had default volume options in Glusterd. I still think Heketi is not the right place to have default volume option, it is logically too far. 

If we don't get any other solution, we will have to merge some form this PR and also add a script in gluster container that applies the option for existing volumes.

phlogistonjohn:
@raghavendra-talur have you made any progress with your design or are you just declaring the "do it all on the gluster container" approach a dead end?

He's a thought:
What if the container provided a script for heketi to call. For example: `/usr/bin/gluster-vol-defaults.sh VOLNAME` Heketi could run it if the script and allow the container (or other node) to set the default options appropriate for that version of gluster. Perhaps the script would even do nothing. If the script doesn't exist heketi could just skip over it.

Any opinions on this idea?

raghavendra-talur:
> @raghavendra-talur have you made any progress with your design or are you just declaring the "do it all on the gluster container" approach a dead end?
> 
> He's a thought:
> What if the container provided a script for heketi to call. For example: `/usr/bin/gluster-vol-defaults.sh VOLNAME` Heketi could run it if the script and allow the container (or other node) to set the default options appropriate for that version of gluster. Perhaps the script would even do nothing. If the script doesn't exist heketi could just skip over it.
> 
> Any opinions on this idea?

I was thinking of a solution which is similar to this but realized that container based solutions won't work in setups where Gluster is in independent cluster outside kubernetes and openshift.

@amarts How difficult is it to backport the fix in Gluster?

amarts:
> @amarts How difficult is it to backport the fix in Gluster?

It is already present in master branch of glusterfs. This was the proposal to make sure even for the versions which are already present in deployments, by getting this option in heketi, we can make n/w failure cases better.

If it cant make it in heketi, then we have to make sure when the corresponding release of glusterfs happens, heketi (aka, container users) should upgrade to such release. As clearly stated the patch is 'temporary' fix, if it is delayed, then there is no need for it in future.

obnoxxx:
> @phlogistonjohn writes:
> > @raghavendra-talur  have you made any progress with your design or are you just declaring  the "do it all on the gluster container" approach a dead end?
> > 
> >  He's a thought: What if the container provided a script for heketi to call. For example: `/usr/bin/gluster-vol-defaults.sh VOLNAME`  Heketi could run it if the script and allow the container (or other  node) to set the default options appropriate for that version of  gluster. Perhaps the script would even do nothing. If the script doesn't  exist heketi could just skip over it. 
> > Any opinions on this idea?
> 
> I was thinking of a solution which is similar to this but realized  that container based solutions won't work in setups where Gluster is in  independent cluster outside kubernetes and openshift.

Not sure what's the inhibiting factor here: If heketi runs in a container, then it can still act on an independent gluster outside of kube. I think what you mean is that your originally envisioned solution where we have a start script in the *gluster* container to fix existing volumes does not work in the independent mode, right?

>  @amarts How difficult is it to backport the fix in Gluster?

I think the main point originally was not that it's difficult to backport (its not!) but that it takes more time to fully qualify in gluster for a real downstream release than we have. And as @amarts says, to have a workaround for working with "older" flavors of gluster that don't have the default yet.

raghavendra-talur:
@amarts @obnoxxx thanks for the explanation.

We now have a generic feature to provide default options in #1382 and #1383 sets the tcp-user-timeout in the container. Please have a look and close this PR is all concerns are addressed.

amarts:
Solution done right now is better for sure, and does solve the issue!

PR #1356, opened 2018-09-15
Small cleanups for heketi-start.sh script

phlogistonjohn:

Small improvements to the readability of the start-heketi.sh script.
Remove an old obsolete version of the script from the tree.


extras/docker/fromsource/heketi-start.sh
```diff
@@ -64,17 +73,17 @@ if [[ -d "${BACKUPDB_PATH}" ]]; then
...
     fi
     if [[ -f "${BACKUPDB_PATH}/heketi.db" ]] ; then
         cp "${BACKUPDB_PATH}/heketi.db" "${HEKETI_PATH}/heketi.db"
         if [[ $? -ne 0 ]]; then
-            echo "Unable to copy backup database" | tee -a "${HEKETI_PATH}/container.log"
+            error "Unable to copy backup database"
             exit 1
         fi
-        echo "Copied backup db to ${HEKETI_PATH}/heketi.db"
+        error "Copied backup db to ${HEKETI_PATH}/heketi.db"
```
obnoxxx: info?

extras/docker/rpi/heketi-start.sh
```diff
@@ -64,17 +73,17 @@ if [[ -d "${BACKUPDB_PATH}" ]]; then
...
     fi
     if [[ -f "${BACKUPDB_PATH}/heketi.db" ]] ; then
         cp "${BACKUPDB_PATH}/heketi.db" "${HEKETI_PATH}/heketi.db"
         if [[ $? -ne 0 ]]; then
-            echo "Unable to copy backup database" | tee -a "${HEKETI_PATH}/container.log"
+            error "Unable to copy backup database"
             exit 1
         fi
-        echo "Copied backup db to ${HEKETI_PATH}/heketi.db"
+        error "Copied backup db to ${HEKETI_PATH}/heketi.db"
```
obnoxxx: info?

extras/docker/fromsource/heketi-start.sh
```diff
@@ -50,12 +59,12 @@ if [[ -d "${HEKETI_PATH}" ]]; then
         sleep 1
     done
     if [[ $flock_status -ne 0 ]]; then
-        echo "Database file is read-only" | tee -a "${HEKETI_PATH}/container.log"
+        error "Database file is read-only"
```
obnoxxx: Is there actually an `exit` missing here?

If it is missing, then we should add it (and could actually make the `exit 1` part of the error function... Otherwise, let' maybe put `info` here?

extras/docker/fromsource/heketi-start.sh
```diff
@@ -6,36 +6,45 @@
...
+                   error "Database file did not appear, exiting."
                    exit 1
                 fi
                 ((check+=1))
             done
         fi
     fi
 
-    stat "${HEKETI_PATH}/heketi.db" | tee -a "${HEKETI_PATH}/container.log"
+    stat "${HEKETI_PATH}/heketi.db" | tee -a "${LOG}"
```
obnoxxx: Could we do something like this?
```
info "$(stat "${HEKETI_PATH}/heketi.db"})"
```

obnoxxx:
Note the comments were on the first patch, so they are hidden in the main view since the contexts have changed.
i'll re-comment on the ultimate diff

extras/docker/fromsource/heketi-start.sh
```diff
@@ -6,75 +6,83 @@
...
+        done
     fi
-else
-    mkdir -p "${HEKETI_PATH}"
-    if [[ $? -ne 0 ]]; then
-        echo "Failed to create ${HEKETI_PATH}"
-        exit 1
+fi
+
+stat "${HEKETI_PATH}/heketi.db" | tee -a "${LOG}"
```
obnoxxx: Can we do something like this?
```
info "$(stat "${HEKETI_PATH}/heketi.db")"
```

extras/docker/fromsource/heketi-start.sh
```diff
@@ -6,75 +6,83 @@
...
-if [[ -d "${HEKETI_PATH}" ]]; then
-    # Test that our volume is writable.
-    touch "${HEKETI_PATH}/test" && rm "${HEKETI_PATH}/test"
-    if [ $? -ne 0 ]; then
-        echo "${HEKETI_PATH} is read-only"
-        exit 1
-    fi
+error() {
+    echo "error: $*" | tee -a "$LOG" >&2
+}
```
obnoxxx: can we add the exit 1 here to reduce more boiler-plate?

extras/docker/fromsource/heketi-start.sh
phlogistonjohn: I want to re-do this logic a bit but I figured I would do that in a different pr, in my mind, this one was just for cleaning up the existing code to make it more readable/convenient w/o much changes to the logic or behavior.

phlogistonjohn:
PR updated.

extras/docker/fromsource/heketi-start.sh
obnoxxx: Well, this would not change anything really, it would result in the stat output being put into the logs, as before, right? that's fine, we can do it later...

extras/docker/rpi/heketi-start.sh
```diff
@@ -6,75 +6,83 @@
...
-        fi
-        sleep 1
-    done
-    if [[ $flock_status -ne 0 ]]; then
-        echo "Database file is read-only" | tee -a "${HEKETI_PATH}/container.log"
+# Ensure the data dir exists
+mkdir -p "${HEKETI_PATH}" 2>/dev/null
+if [[ $? -ne 0 && ! -d "${HEKETI_PATH}" ]]; then
+    error "Failed to create ${HEKETI_PATH}"
+    exit 1
```
obnoxxx: no fail function here

extras/docker/rpi/heketi-start.sh
```diff
@@ -6,75 +6,83 @@
...
+for _ in $(seq 1 60); do
+    flock --nonblock "${HEKETI_PATH}/heketi.db" true
+    flock_status=$?
+    if [[ $flock_status -eq 0 ]]; then
+        break
     fi
+    sleep 1
+done
+if [[ $flock_status -ne 0 ]]; then
+    error "Database file is read-only"
```
obnoxxx: exit here

extras/docker/rpi/heketi-start.sh
obnoxxx: info!

phlogistonjohn:
Ugh. Forgot to keep the rpi script in sync with the "real one". Someday I want to remove the copy, but that's also not a job for this pr :-).

PR #1360, opened 2018-09-18
Fail device remove if vgremove hits an error

### LINKED ISSUE #1359: Heketi deletes a device even if VG & PV fail to be removed
### Kind of issue
Bug

### Observed behavior

If, for whatever reason, Heketi is unable to run vgremove on the device vg it still considers the device "deleted" and removes it from the db. It seems like this device could then be reused for other purposes but it is not.

### Expected/desired behavior

Heketi should report an error at the API level and leave the device in the db. Then once the underlying problem is resolved, the admin can retry the device delete again.

Before removing the device add a new "dummy" LV to the vg such that attempting to remove the VG will produce an error (or prompt in non-heketi lvm use). Heketi will consider the device deleted from the db but the device can not be re-added to heketi afterward.

- Heketi version used (e.g. v6.0.0 or master): master
- Operating system used: linux 
- Heketi compiled from sources, as a package (rpm/deb), or container: sources

### Other useful information

See also: [rhbz#1629889](https://bugzilla.redhat.com/show_bug.cgi?id=1629889)

phlogistonjohn:

When deleting a device do not ignore failures to remove vg/pv.

### Does this PR fix issues?

Fixes #1359

### Notes for the reviewer

Functional test case added.

### DIFF
--- a/executors/cmdexec/device.go
+++ b/executors/cmdexec/device.go
@@ -80,7 +80,8 @@ func (s *CmdExecutor) DeviceTeardown(host, device, vgid string) error {
 	// Execute command
 	_, err := s.RemoteExecutor.RemoteCommandExecute(host, commands, 5)
 	if err != nil {
-		logger.LogError("Error while deleting device %v with id %v on host %v: %v",
+		return logger.LogError(
+			"Failed to delete device %v with id %v on host %v: %v",
 			device, vgid, host, err)
 	}
 
--- a/executors/cmdexec/device.go
+++ b/executors/cmdexec/device.go
@@ -85,6 +85,8 @@ func (s *CmdExecutor) DeviceTeardown(host, device, vgid string) error {
 			device, vgid, host, err)
 	}
 
+	// TODO: remove this LBYL check and replace it with the rmdir
+	// followed by error condition check that handles ENOENT
 	pdir := utils.BrickMountPointParent(vgid)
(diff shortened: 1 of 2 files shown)


executors/cmdexec/device.go
```diff
@@ -80,10 +80,13 @@ func (s *CmdExecutor) DeviceTeardown(host, device, vgid string) error {
 	// Execute command
 	_, err := s.RemoteExecutor.RemoteCommandExecute(host, commands, 5)
 	if err != nil {
-		logger.LogError("Error while deleting device %v with id %v on host %v: %v",
+		return logger.LogError(
```
nixpanic: What does `logger.LogError()` return? Should this not be an instance of `error` instead?

nixpanic: Ok, so this wrapper does not return the success/fail of the logging, but it returns the formatted string. Who would have thought that!?

executors/cmdexec/device.go
```diff
@@ -85,6 +85,8 @@ func (s *CmdExecutor) DeviceTeardown(host, device, vgid string) error {
 			device, vgid, host, err)
 	}
 
+	// TODO: remove this LBYL check and replace it with the rmdir
```
obnoxxx: Heh, I had to look up LYBL .... :-)

PR #1364, opened 2018-09-20
Convert executors to use new style command execution functions

phlogistonjohn:

PR #1350 added a new approach to getting results from commands run remotely. This PR adds utility functions to reduce boilerplate code for working with the new types and then remove the old approach from the executor code.

### Notes for the reviewer

This PR depends on PR #1350, please review and merge that series first.

While the old approach is totally gone from the kube executor code the ssh.go file still exposes the old style function. This function is still used all over the functional tests and I did not want to tackle converting all those usages to the new function. I do plan on doing it eventually though.


phlogistonjohn:
retest this please

pkg/remoteexec/result.go
```diff
@@ -49,3 +49,49 @@ func (rs Results) SquashErrors() ([]string, error) {
...
+			return false
+		}
+	}
+	return true
+}
+
+// FirstErrorIndexed returns the first error found in the Results
+// along with its index. If no error is found -1 and nil are
+// returned.
+func (rs Results) FirstErrorIndexed() (int, error) {
```
Madhu-1: question:  how are we planning to make use of returned index value?

PR #1373, opened 2018-09-26
Split kube cmd execution logic into a pkg

phlogistonjohn:

Split the logic of running commands on a container via the k8s api into it's own package, rather than having all the logic smooshed into the executors/kubeexec package.
This will allow for easier future improvements of the pkg as well as making it easier to use kube commands like we do ssh commands in the functional tests someday.

This series depends on the changes in PR #1364.

### Notes for the reviewer

This series depends on the changes in PR #1364, please review that series first.

This series will conflict with the changes in PR #1272, however I think that this provides a cleaner & more testable foundation for the kind of changes being made in that PR. This change does not change the apis currently used by the kubeexec version. If we merge this I think we have a better opportunity for going forward with #1272.

Speaking of tests, the "unit test" for the exec function is a bit hokey but does make it possible to try out just the execution bits without the rest of heketi along for the ride. In the future, we can add to the test setup code so that it could run outside of a (heketi) pod with something like minikube for example.


pkg/remoteexec/kube/target.go
```diff
@@ -0,0 +1,201 @@
...
+	if numPods == 0 {
+		// No pods found with that label
+		err := fmt.Errorf("No pods with the label '%v=%v' were found",
+			t.Key, t.Value)
+		k.logger.Critical(err.Error())
+		return tp, err
+
+	} else if numPods > 1 {
+		// There are more than one pod with the same label
+		err := fmt.Errorf("Found %v pods with the sharing the same label '%v=%v'",
```
Madhu-1: if we log the pod names will it be helpful?

phlogistonjohn: done in followup commit

pkg/remoteexec/kube/conn_test.go
```diff
@@ -0,0 +1,65 @@
...
+	"testing"
+
+	restclient "k8s.io/client-go/rest"
+
+	"github.com/heketi/tests"
+)
+
+type dummyLogger struct{}
+
+func (d *dummyLogger) LogError(s string, v ...interface{}) error {
```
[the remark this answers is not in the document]
phlogistonjohn: Nice suggestion!

obnoxxx:
@phlogistonjohn is the "do not merge" label still correct here?

phlogistonjohn:
Label removed

obnoxxx:
@phlogistonjohn i rebased it for you after merging the other PRs...

PR #1374, opened 2018-09-27
remove waffle io badge & link from main page

phlogistonjohn:

None of the current heketi developers use the waffle.io system previously set up for heketi. Since I joined the project the only times I ever viewed it was when I was wondering, "what is this?"
I've also disabled the associated webhook which was apparently failing for months.

This change removes clutter and possible confusion from the project's README


nixpanic:
On Thu, Sep 27, 2018 at 11:10:18AM -0700, Coveralls wrote:
> 
> [Coverage Status](https://coveralls.io/builds/19230793)
> 
> Coverage remained the same at 66.951% when pulling **ba4598b0a4ee51c1e64895bd6c41c1d0782ebf91 on jjm-no-waffle-badge** into **001943a0d6e2dfd84a66f36e7799cd6d9953aba6 on master**.

I'm like "what is this?" with coveralls too :D

phlogistonjohn:
This is the first time I've seen coveralls comment on a PR. That's rather odd.

@obnoxxx and I briefly discussed the state of waffle a few days ago. We could try reviving waffle OR just drop it entirely. I lean toward the latter. However, I don't think we should have it linger in a half-forgotten state (with failing web hooks, etc).

PR #1379, opened 2018-10-01
Check retry errors for behavior instead of type

ansiwen:
This PR is based on top of #1348, please merge that first.

PR #1348 replaces the errors package with the github.com/pkg/errors package, which allows to annotate errors with stack traces and wrap them into errors with additional messages. The original causing error is recovered by the errors.Cause() function. In order to make the OperationRetryError compatible with that scheme, this PR refactors it by following the "Assert errors for behaviour, not type" principle from [this](https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully) article and wrap the checks into functions that take care of calling errors.Cause() before before checking for behavior.

### DIFF
--- a/apps/glusterfs/app.go
+++ b/apps/glusterfs/app.go
@@ -16,13 +16,14 @@ import (
 
 	"github.com/boltdb/bolt"
 	"github.com/gorilla/mux"
+	"github.com/heketi/rest"
+	"github.com/pkg/errors"
+
 	"github.com/heketi/heketi/executors"
 	"github.com/heketi/heketi/executors/kubeexec"
 	"github.com/heketi/heketi/executors/mockexec"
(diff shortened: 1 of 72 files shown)


ansiwen:
PS: I would have preferred to call the _behavior_ method of a retry error `Temporary()`instead of `Retry()` for errors that indicate the operation might be retried, in accordance with for example the standard [net.Error](https://golang.org/pkg/net/#Error). There might be a slight semantic difference between `IsTemporary()` and `IsRetry()`, but I was not able to see it yet.(For me if the _error_ is _temporary_ the _operation_ might be _retried_.)

phlogistonjohn:

ansiwen:
@phlogistonjohn rebased

phlogistonjohn:

PR #1380, opened 2018-10-01
Replace sentinel errors with comparable ErrTag type

ansiwen:
This PR is based on top of #1348, please merge that first.

PR #1348 replaces the errors package with the github.com/pkg/errors package, which allows to annotate errors with stack traces and wrap them into errors with additional messages. The original causing error is recovered by the errors.Cause() function. Since sentinel errors are not compatible with these stack-annotated errors (since all errors are created beforehand, and the stack would refer to the point of error creation, not the point of error emission, which is useless), a new error type ErrTag is introduced, that can produce errors with Err() that contain stack traces and are still identifiable with the In() method.

### DIFF
--- a/apps/glusterfs/device_entry.go
+++ b/apps/glusterfs/device_entry.go
@@ -22,6 +22,7 @@ import (
 	"github.com/heketi/heketi/pkg/idgen"
 	"github.com/heketi/heketi/pkg/sortedstrings"
 	"github.com/lpabon/godbc"
+	"github.com/pkg/errors"
 )
 
 const (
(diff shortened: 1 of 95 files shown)


phlogistonjohn:

apps/glusterfs/device_entry.go
```diff
@@ -485,7 +485,7 @@ func (d *DeviceEntry) removeBricksFromDevice(db wdb.DB,
 		logger.Info("Replacing brick %v on device %v on node %v", brickEntry.Id(), d.Id(), d.NodeId)
 		err = volumeEntry.replaceBrickInVolume(db, executor, brickEntry.Id())
 		if err != nil {
-			return logger.Err(errors.Errorf("Failed to remove device, error: %v", err))
+			return logger.Err(fmt.Errorf("Failed to remove device, error: %v", err))
```
ansiwen: Note to myself: this is wrong

phlogistonjohn:

PR #1381, opened 2018-10-01
Split the utils package into more distinct packages

phlogistonjohn:

The goal of this PR is to split some of the major components of the utils package into packages of there own when the component  is used by the heketi server alone (not used by the client api pkg).
This will create a clear(er) cut distinction for what the client pkg dependencies are.


obnoxxx:
@phlogistonjohn , is the following statement from the original PR message still true?

> ### Notes for the reviewer
> This PR still needs some cleanups and tweaking and as well as some patch squashing before it can be merged. But it's in a good enough state to see the general direction of things.

The patches look clean to me as they are!

phlogistonjohn:
Comment about needed to clean and/or squash removed.

obnoxxx:
@phlogistonjohn writes
> Comment about needed to clean and/or squash removed.

Ok, thanks! So good to merge once CI passes (again, after the last rebasing on top of master).

phlogistonjohn:
retest this please

PR #1382, opened 2018-10-02
Default volume options 

raghavendra-talur:
    Volume create requests have provision to provide volume options.
    However, a heketi admin does not have any means to set volume options
    by default. This feature allows admin to set default volume options.

    It is possible that heketi admin and heketi volume creator want to
    set different values for the same volume option. To address such use
    cases, admin is provided with two default option lists, as below:

    1. HEKETI_PRE_REQUEST_VOLUME_OPTIONS env variable
       pre_request_volume_options key in config's glusterfs section

    Volume options that will be applied for all volumes created. Can be
    overridden by volume options in volume create request.

    2. HEKETI_POST_REQUEST_VOLUME_OPTIONS env variable
       post_request_volume_options key in config's glusterfs section

    Volume options that will be applied for all volumes created. Will
    override volume options in volume create request.

### Does this PR fix issues?

Fixes #

### DIFF
--- a/apps/glusterfs/app.go
+++ b/apps/glusterfs/app.go
@@ -54,6 +54,14 @@ var (
 	// avoids having to update config files to enable the feature
 	// while avoiding having to touch all of the unit tests.
 	MonitorGlusterNodes = false
+
+	// global var that contains list of volume options that are set *before*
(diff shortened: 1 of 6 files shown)


phlogistonjohn:
retest this please

PR #1383, opened 2018-10-03
container: add tcp-user-timeout in default option

raghavendra-talur:

    container: add tcp-user-timeout in default option
    
    As identified in glusterfs patch[1], we need option
    'server.user-tcp-timeout' to be set on gluster-block volume to mitigate
    some of the 'hung status' like behavior when a node goes down. It is
    also required for file volumes but the impact is not as critical. Adding
    the option in heketi's default pre_request_volume_options will apply it
    for all new volumes created by heketi.
    
    For existing volumes, users can query the volume list and perform a
    gluster volume set.
    
    [1] https://review.gluster.org/#/c/glusterfs/+/21170/

### Does this PR fix issues?

Fixes #

### Notes for the reviewer
Obsoletes #1354

### DIFF
--- a/extras/docker/fromsource/heketi-start.sh
+++ b/extras/docker/fromsource/heketi-start.sh
@@ -86,6 +86,23 @@ if [[ -d "${BACKUPDB_PATH}" ]]; then
     fi
 fi
 
+info "HEKETI_PRE_REQUEST_VOLUME_OPTIONS is ${HEKETI_PRE_REQUEST_VOLUME_OPTIONS}"
+if [[ -z "$HEKETI_PRE_REQUEST_VOLUME_OPTIONS" ]]
+then
+        # set the env variable as it is not set
+        export HEKETI_PRE_REQUEST_VOLUME_OPTIONS="server.tcp-user-timeout 42"
+else
+        if [[ ! -z "${HEKETI_PRE_REQUEST_VOLUME_OPTIONS##*server.tcp-user-timeout*}" ]]
+        then
+                # variable exists but does not have tcp-user-timeout set, append it
+                echo "did not find server.tcp-user-timeout in options" >> ${HEKETI_PATH}/container.log
+                export HEKETI_PRE_REQUEST_VOLUME_OPTIONS="${HEKETI_PRE_REQUEST_VOLUME_OPTIONS},server.tcp-user-timeout 42"
+        fi
+        # else case would be where tcp-user-timeout is set with some value. don't override it.
+fi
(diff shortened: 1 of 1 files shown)


extras/docker/fromsource/heketi-start.sh
```diff
@@ -86,6 +86,23 @@ if [[ -d "${BACKUPDB_PATH}" ]]; then
...
+info "HEKETI_PRE_REQUEST_VOLUME_OPTIONS is ${HEKETI_PRE_REQUEST_VOLUME_OPTIONS}"
+if [[ -z "$HEKETI_PRE_REQUEST_VOLUME_OPTIONS" ]]
+then
+        # set the env variable as it is not set
+        export HEKETI_PRE_REQUEST_VOLUME_OPTIONS="server.tcp-user-timeout 42"
+else
+        if [[ ! -z "${HEKETI_PRE_REQUEST_VOLUME_OPTIONS##*server.tcp-user-timeout*}" ]]
+        then
+                # variable exists but does not have tcp-user-timeout set, append it
+                echo "did not find server.tcp-user-timeout in options" >> ${HEKETI_PATH}/container.log
```
phlogistonjohn: Use `info` here instead of echo?

extras/docker/fromsource/heketi-start.sh
```diff
@@ -86,6 +86,23 @@ if [[ -d "${BACKUPDB_PATH}" ]]; then
...
+                export HEKETI_PRE_REQUEST_VOLUME_OPTIONS="${HEKETI_PRE_REQUEST_VOLUME_OPTIONS},server.tcp-user-timeout 42"
+        fi
+        # else case would be where tcp-user-timeout is set with some value. don't override it.
+fi
+info "modified HEKETI_PRE_REQUEST_VOLUME_OPTIONS is $HEKETI_PRE_REQUEST_VOLUME_OPTIONS"
```
phlogistonjohn: I woudln't use the word modified here, just skip it or use "current" or something like that.

nixpanic:
Why only `server.tcp-user-timeout` and not `client.tcp-user-timeout`?

phlogistonjohn:
@raghavendra-talur can you answer @nixpanic's ealier question?

Because PR #1316 has not been merged you also need to make matching edits for the other copies of the start script elsewhere in the tree.

phlogistonjohn:
@raghavendra-talur Can you revisit this?

raghavendra-talur:
This PR was required only for those gluster versions where the default tcp-user-timeout was wrongly configured to be 0.

The bug has been fixed in all the latest releases of Gluster.

Symptom: If writes hang at Gluster client when 1/3 bricks are disconnected from client.
Fix: Set the tcp-user-timeout option to 42 OR upgrade.

PR #1385, opened 2018-10-03
extras: add a subdir to support integration with openshift ci

phlogistonjohn:

Add a dockerfile and a readme file that will be used to
help bootstrap testing heketi in openshift-ci.


raghavendra-talur:
@centos-ci retest this please

PR #1386, opened 2018-10-04
makefile: stop using -i in go build options

phlogistonjohn:

After reading up a bit on go modules (vgo) it appears
that -i is no longer needed or recommended for Go [1].
It may have improved performance on some versions prior to
1.10 but is not needed for correctness and may not
have been much use for containerized builds anyway.

1: golang/go#27285 (comment)

### Notes for the reviewer

Merging this will help enable more experimentation with go module support.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,11 +14,6 @@ GOOS := $(shell $(GO) env GOOS)
 GOHOSTARCH := $(shell $(GO) env GOHOSTARCH)
 GOHOSTOS := $(shell $(GO) env GOHOSTOS)
 GOBUILDFLAGS :=
-ifeq ($(GOOS),$(GOHOSTOS))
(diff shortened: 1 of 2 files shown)


nixpanic:
On Mon, Oct 08, 2018 at 09:20:56AM -0700, Michael Adam wrote:
> the centos-ci build was aborted because it reached the 2hr mark...

We can change that to a longer timeout. What would be appropriate?

The configuration is at
https://github.com/gluster/centosci/blob/master/heketi-functional.yml#L68

phlogistonjohn:
Yeah, IIRC there are two timeouts that would need to be adjusted one within heketi's scripts and one at centos level. I'd be happy bumping it up by another hour, but eventually we need to start thinking about accelerating the tests rather than just bumping the timeout up.

PR #1390, opened 2018-10-09
container: use current working repository for "fromsource" image

nixpanic:

This builds makes it possible to build container images from the currently checked-out sources. The `extras/docker/fromsource/Dockerfile` configuration has a hardcoded git repository and branch. This is not very flexible.

The following command will build the image, run it from the root of the repository:

```shell
$ buildah bud --network=host --file=extras/docker/fromsource/Dockerfile .
```

### Notes for the reviewer

With this change it becomes much easier to do builds when a GitHub PR is sent. My [personal heketi repository on Quay](https://quay.io/repository/nixpanic/heketi) builds for the pushes I do to any of my 'wip/...' or 'pr/...' branches.

### DIFF
--- a/extras/docker/fromsource/Dockerfile
+++ b/extras/docker/fromsource/Dockerfile
@@ -1,7 +1,20 @@
-# set author and base
+# Dockerfile to build a container image from the current working tree. This
+# includes any non-committed changes.
+#
+# Build from the root of the repository with:
+#
+#  $ buildah bud --file=extras/docker/fromsource/Dockerfile .
+#
+#
+# Alternatively, the legacy way of building the image based on the master
+# branch of the heketi github repository is still functional:
+#
+#  $ buildah bud --file=extras/docker/fromsource
+#
+
 # need to use fedora 27 until glide is fixed in fedora 28
 # (or we get an alternative glide)
-FROM fedora:27
+FROM fedora:27 AS build
 MAINTAINER Heketi Developers <heketi-devel@gluster.org>
 
 LABEL version="1.3.1"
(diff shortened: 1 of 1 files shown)


extras/docker/fromsource/Dockerfile
```diff
@@ -11,29 +19,42 @@ LABEL description="Development build"
...
-    cp client/cli/go/heketi-cli /usr/bin/heketi-cli && \
-    glide cc && \
-    cd && rm -rf $BUILD_HOME && \
-    dnf -y remove git glide golang mercurial && \
-    dnf -y autoremove && \
-    dnf -y clean all
+    cp client/cli/go/heketi-cli /usr/bin/heketi-cli
+
+
+# the final container can be based on the latest Fedoda version
```
phlogistonjohn: "Fedoda"

nixpanic: *bleah*

nixpanic:
On Tue, Oct 09, 2018 at 11:36:07AM -0700, John Mulligan wrote:
> First, our primary image host is dockerhub so we need to verify that we can take these changes and expect dockerhub to continue working.

Yes, if this change would get merged now, it is likely that the
configuration on Docker Hub needs adjusting.

> Another issue I have with this is that the Dockerfile is no longer self-sufficient. While this was never 100% true you could pretty naively run `docker build extras/docker/fromsource` without reading any additional instructions and things would work. This is no longer the case with this change.

Right, and I guess that is basically what Docker Hub currently runs as
well. The 'build context' or 'working directory' needs to be set to the
root of the git repository with this PR.

We can make it easier by documenting better. And even a 'make container'
or similar target in the Makefile (which would allow a lot of other
options and flexibility as well).

> Finally, I don't think this change leaves us with a clean vendor tree which is something I want from containerized builds, even if I'm building from my desktop.

I am not sure what you mean with 'clean vendor tree'. Because the vendor
tree is not checked in, building a container image without (re)running
glide is not an option. If this is what you prefer, we could change the
RUN command in the Dockerfile to not run any glide stuff in case the
vendor directory exists.

nixpanic:
On Tue, Oct 09, 2018 at 09:35:06PM +0200, Niels de Vos wrote:
..
> > Finally, I don't think this change leaves us with a clean vendor tree which is something I want from containerized builds, even if I'm building from my desktop.
> 
> I am not sure what you mean with 'clean vendor tree'. Because the vendor
> tree is not checked in, building a container image without (re)running
> glide is not an option. If this is what you prefer, we could change the
> RUN command in the Dockerfile to not run any glide stuff in case the
> vendor directory exists.

This is probably a good thing to have anyway. The downloading of
everythign in vendor/ takes most of the time while building. It should
probably be skipped for local container builds.

nixpanic:
@phlogistonjohn I think this is more in the direction what you want. With this (hacky way) it is possible to run `docker build extras/docker/fromsource` **and** `buildah bud --file=extras/docker/fromsource/Dockerfile .`. The way of building depending on the contents that `COPY` placed in the build directory.

It also prevents running `glide install` in the case where `vendor/` already exists (`COPY`'d from the working repository).

This is a full/clean build in quay: https://quay.io/repository/nixpanic/heketi/build/bc24c544-d07f-4124-9e54-45d0765681f1. You can see that the 1st `RUN` command did not do anything useful and finished in a second. The 2nd one actually did build the binaries.

phlogistonjohn:
While I like using the build container, I strongly disagree with allowing the default method for building the container be "tainted" by the state of a developers working directory. I'm not against having a "fast" mode but I do think it should be clearly marked as such and we should have no way of accidentally building a "fromsource" images with bits from the developers tree. Since the discussion also previously petered out a little under a year ago I will be closing this PR. Reopen or file a new PR if you feel that is appropriate, thanks.

PR #1393, opened 2018-10-10
makefile: make adjustments to openshift-ci behavior

phlogistonjohn:

When the tests are run within a container in openshift-ci certain less typical (compared to desktop/server linux) environmental conditions are present. In the case of our tests what caused problems for our tests was the lack of a normal home directory and a UID that was generated and not present in /etc/passwd. These two conditions are worked around in order to prevent poor behaviors from glide and tox.
In addition, one of the kubeexec unit tests was failing because it assumed that it was never running inside a pod. This is no longer true. The test was disabled, rather than removed, in order to make me revisit it when I next rebase the patches in PR #1373.

### Does this PR fix issues?

One this change is commited to heketi, matching changes need to be commited to the openshift-ci release repo. Once that is done we ought to be able to start regularly running some tests in openshift-ci!

### Notes for the reviewer

Please expedite looking at this patch series so we can get centos-ci up and running sooner.

Note that while I tested this patch defaulted to CI=openshift in the Makefile, this has been removed as its not a good default outside of openshift ci. Thus we can't run this can in openshift ci again until the CI config can be changed to match.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,8 @@ BRANCH := $(subst /,-,$(shell git rev-parse --abbrev-ref HEAD))
 VER := $(shell git describe --match='v[0-9].[0-9].[0-9]')
 TAG := $(shell git tag --points-at HEAD 'v[0-9].[0-9].[0-9]' | tail -n1)
 GO:=go
+GLIDE:=glide
+TESTBIN:=./test.sh
 GOARCH := $(shell $(GO) env GOARCH)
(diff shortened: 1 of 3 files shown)


phlogistonjohn:
retest this please

phlogistonjohn:
@obnoxxx  @raghavendra-talur  when you have a moment PTAL. This will enable the use of openshift-ci on other PRs.

Issue #1401, opened 2018-10-22
Automatically clean up after stale and failed operations


phlogistonjohn:
### Kind of issue

Feature request

### Observed behavior

When the server is terminated during the execution of an operation or an operation fails rollback the operation lingers in the heketi db until manually cleaned up. The reason heketi retains this information is so that the underlying storage systems can be checked that all potential items created by heketi are removed in a repeatable manner. It's time to actually use this information the way it was meant to be used and allow heketi to clean up most of the operations itself, rather than need human intervention.

### Expected/desired behavior

Heketi will support:
* A core function for cleaning up / reverting partial operations.
* Automatic clean up of operation after heketi is restarted (to catch most stale operations)
* Periodic clean up of operations in a long running heketi (to catch failed operations)
* Admin command to trigger clean ups:
  * General clean up command
  * Clean a specific operation (by id)
* CLI command to list operations & state by id
* CLI command (new or enhance existing) to get status of cleanups

### Other useful information

This is a "meta issue" filed to use as a reference for this general effort. Individual issues & PRs, etc. may be filed to cover specific aspects of the overall task.

phlogistonjohn:
Needs to be able to handle "orphaned" pending operations from previous versions. Orphaned operations are operations where the IDs in the actions don't point to any current obejcts in the db.

phlogistonjohn:
Closing issue. Any bugs or enhancements needed for this feature set can be filed separately.

PR #1404, opened 2018-10-23
Break up the rather large operations.go into multiple files

phlogistonjohn:

The file operations.go has grown to the point where it is too large. This PR splits up the file such that operations.go only retains the most generic functions & types. New files for volume operations, block volume operations, and device operations are added. Similarly, the operations_test.go is broken up in a similar manner. The mutli cluster error is put in a new file as well.

This should make working with the operations & their tests simpler in the future.

### Notes for the reviewer

No substantive changes were made in these changesets, save to get the imports correct for the new files. Every other line in the new files should be exactly as it was in operations.go.

This is preparatory work for #1401. This series is based on top of #1400 please review and merge that first.


apps/glusterfs/operations_block_volume.go
```diff
@@ -0,0 +1,353 @@
...
+
+package glusterfs
+
+import (
+	"fmt"
+
+	"github.com/heketi/heketi/executors"
+	wdb "github.com/heketi/heketi/pkg/db"
+
+	"github.com/boltdb/bolt"
```
Madhu-1: this import order should be changed?

importing third party lib should be the second set of import

apps/glusterfs/operations_block_volume.go
```diff
@@ -0,0 +1,353 @@
...
+)
+
+// BlockVolumeCreateOperation  implements the operation functions used to
+// create a new volume.
+type BlockVolumeCreateOperation struct {
+	OperationManager
+	noRetriesOperation
+	bvol *BlockVolumeEntry
+	//vol *VolumeEntry
```
Madhu-1: commented code

apps/glusterfs/operations_device.go
```diff
@@ -0,0 +1,143 @@
...
+// later), or the GNU General Public License, version 2 (GPLv2), in all
+// cases as published by the Free Software Foundation.
+//
+
+package glusterfs
+
+import (
+	"fmt"
+
+	"github.com/heketi/heketi/executors"
```
Madhu-1: which import format we need to follow
```
import (
	"errors"
	"fmt"
	"os"
	"strconv"

	"github.com/lpabon/godbc"

	"github.com/heketi/heketi/executors/cmdexec"
)
```
and
```
import (
	"bytes"
	"encoding/gob"
	"encoding/json"
	"net/http"
	"time"

	"github.com/heketi/heketi/pkg/glusterfs/api"

	"github.com/boltdb/bolt"
	"github.com/lpabon/godbc"
)
```

apps/glusterfs/operations_block_volume.go
phlogistonjohn: Sure, but lets do that in a follow up commit. I've been trying to use that approach personally but right now heketi as a project does not enforce that as a general rule (yet).

apps/glusterfs/operations_block_volume.go
phlogistonjohn: This commit was a pure move and the comment existed previously. Will clean it up at a later time.

obnoxxx:
@phlogistonjohn I did the rebase on #1400 for you ;-)

PR #1407, opened 2018-10-29
Avoid manual addition of heketi.json

### LINKED ISSUE #1406: heketi.json is added manually in the Dockerfile
### Kind of issue

 Bug 

I am referring the below Dockerfile as the heketi upstream Dockerfile.
https://github.com/heketi/heketi/blob/master/extras/docker/centos/Dockerfile#L13

As you can see heketi.json is added manually here. 
When heketi is installed it will automatically brings in heketi.json - and it is the recent one from the rpm.

Any specific reason to overwrite the heketi.json ?  Keeping two different heketi.json is error prone and recent changes in heketi.json(from rpm) may be missed.

SaravanaStorageNetwork:
heketi.json is provided during heketi installation.
Avoid manual addition of heketi.json

Signed-off-by: Saravanakumar Arumugam <sarumuga@redhat.com>

heketi.json is provided during heketi installation.
Avoid manual addition of heketi.json

### Does this PR fix issues?

Fixes #1406

### DIFF
--- a/extras/docker/centos/Dockerfile
+++ b/extras/docker/centos/Dockerfile
@@ -10,7 +10,6 @@ RUN yum -y install centos-release-gluster && \
     yum -y clean all
 
 # post install config and volume setup
-ADD heketi.json /etc/heketi/heketi.json
 ADD heketi-start.sh /usr/bin/heketi-start.sh
 
 VOLUME /etc/heketi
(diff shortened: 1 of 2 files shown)


obnoxxx:
This needs rebase with conflict resolution..

SaravanaStorageNetwork:
@obnoxxx PTAL

phlogistonjohn:
I think we will decline to take this PR. Little discussion has happened since January and I personally don't think this is the right approach. Please feel free to reopen if you wish to discuss further.

PR #1408, opened 2018-10-30
client: for completeness support volume options on heketidbstorage

phlogistonjohn:

Allow passing explicit volume options to the command that creates
heketidbstorage volume. This will be useful if the deployment tools being
used to setup the volume need to pass an option that is different
from the defaults.


client/cli/go/cmds/heketi_storage.go
```diff
@@ -68,6 +70,11 @@ func init() {
 		3,
 		"\n\tOptional: Replica value for durability type 'replicate'."+
 			"\n\tDefault is 3")
+	setupHeketiStorageCommand.Flags().StringVar(&heketiStorageOptions,
```
Madhu-1: can we use `StringSliceVar` instead of `StringVar`?

phlogistonjohn: We could, but then it would work differently from the option on `volume create`. If we were doing it from the start I would have suggested that as well. In this case I think that consistency beats cleanliness.

PR #1411, opened 2018-10-31
Convert topology info to a template and add tags to template

phlogistonjohn:

These changes switch topology info from using a bunch of printf-style calls to a "text/template" based templated output. The template ought to be easier to maintain that the printfs. After converting to a template, lines for tags on nodes and devices are added.

### Does this PR fix issues?

Enhances output of topology info, fixing [rhbz#1572579](https://bugzilla.redhat.com/show_bug.cgi?id=1572579)


obnoxxx:
@phlogistonjohn rebase needs conflict resolution..

PR #1417, opened 2018-11-08
apps:healthz implement healthz URL

### LINKED ISSUE #1412: Add a more complete /healthz endpoint to heketi
### Kind of issue

Feature request

### Observed behavior

The /hello endpoint is simple and works well for a quick curl test but says nothing else about the internal state of the server.

### Expected/desired behavior

Provide a /healthz endpoint that follows common conventions and is able to report on more about the internal state of the server.

Madhu-1:

`healthz` checks the following status in heketi.

* check DB is readable.
* check DB is writable.
* check the gluster nodes are reachable.

### Does this PR fix issues?
Fixes #1412

### Notes for the reviewer
TODO:
* split the commit in to multiple commits
* add test case

### DIFF
--- a/apps/glusterfs/health_cache.go
+++ b/apps/glusterfs/health_cache.go
@@ -174,6 +174,9 @@ func (hc *NodeHealthCache) toProbe() ([]*NodeHealthStatus, error) {
 
 func (s *NodeHealthStatus) update(e executors.Executor) {
 	// TODO: add ability to skip check if node was already recently checked
+	if !s.Up {
+		// skip  glusterd status check if node  is not  marked as  online
+	}
 	err := e.GlusterdCheck(s.Host)
 	s.Up = (err == nil)
 	s.LastUpdate = healthNow()
--- a/apps/glusterfs/health_cache.go
+++ b/apps/glusterfs/health_cache.go
@@ -154,14 +154,17 @@ func (hc *NodeHealthCache) toProbe() ([]*NodeHealthStatus, error) {
 			if err != nil {
 				return err
 			}
-			// Ignore if the node is not online
-			if !node.isOnline() {
-				continue
-			}
+
 			nhs := &NodeHealthStatus{
 				NodeId: nodeId,
 				Host:   node.Info.Hostnames.Manage[0],
 			}
+			// Ignore if the node is not online
+			if !node.isOnline() {
+				continue
+			}
+			nhs.Up = true
+
 			probeNodes = append(probeNodes, nhs)
 		}
 		return nil
--- a/apps/glusterfs/health_cache.go
+++ b/apps/glusterfs/health_cache.go
@@ -83,11 +83,8 @@ func (hc *NodeHealthCache) Refresh() error {
 func (hc *NodeHealthCache) updateNode(s *NodeHealthStatus) {
 	hc.lock.Lock()
 	defer hc.lock.Unlock()
-	if prev, found := hc.nodes[s.NodeId]; found {
-		s = prev
-	} else {
-		hc.nodes[s.NodeId] = s
-	}
+
+	hc.nodes[s.NodeId] = s
 	s.update(hc.exec)
 }
 
--- a/apps/glusterfs/healthz.go
+++ b/apps/glusterfs/healthz.go
@@ -1,0 +1,90 @@
+//
+// Copyright (c) 2019 The heketi Authors
+//
+// This file is licensed to you under your choice of the GNU Lesser
+// General Public License, version 3 or any later version (LGPLv3 or
+// later), or the GNU General Public License, version 2 (GPLv2), in all
+// cases as published by the Free Software Foundation.
+//
+
+package glusterfs
+
+import (
+	"encoding/json"
+	"net/http"
+
+	"github.com/heketi/heketi/pkg/glusterfs/api"
+
+	"github.com/boltdb/bolt"
+)
+
+var (
+	clusterHealth = "ClusterHealth"
+)
+
+func checkDBState(db *bolt.DB, info *api.Healthz) {
+
+	var faulty = "faulty"
+	//db Read check
+	err := db.View(func(tx *bolt.Tx) error {
+		_, err := NewDbAttributeEntryFromKey(tx, clusterHealth)
+		return err
+	})
+
+	if err != nil && err != ErrNotFound {
+		info.ClusterHeath = faulty
+		info.DbState = faulty
+		return
+	}
+	info.DbState = "read"
+
+	//db write check
+	err = db.Update(func(tx *bolt.Tx) error {
+		h := NewDbAttributeEntry()
+		h.Key = clusterHealth
+		return h.Save(tx)
+	})
+	if err == nil {
+		info.DbState = "write"
+	}
+	return
+}
+
+func checkClusterState(health *NodeHealthCache, info *api.Healthz) {
+
+	if len(health.nodes) == 0 {
+		return
+	}
+	totalUnhealthyNodes := 0
+	for _, n := range health.nodes {
+		if !n.Up {
+			totalUnhealthyNodes++
+		}
+		info.NodesStatus = append(info.NodesStatus, (api.NodeHealthStatus)(*n))
+	}
+	if totalUnhealthyNodes == len(info.NodesStatus) {
+		info.ClusterHeath = "faulty"
+	} else {
+		info.ClusterHeath = "healthy"
+	}
+	return
+}
+
+func (a *App) Healthz(w http.ResponseWriter, r *http.Request) {
+	info := api.Healthz{}
+	info.NodesStatus = make([]api.NodeHealthStatus, 0)
+
+	w.Header().Set("Content-Type", "application/json; charset=UTF-8")
+
+	// db status check
+	checkDBState(a.db, &info)
+
+	if MonitorGlusterNodes {
+		// cluster status check
+		checkClusterState(a.nhealth, &info)
+	}
+
+	if err := json.NewEncoder(w).Encode(info); err != nil {
+		panic(err)
+	}
+}
--- a/pkg/glusterfs/api/types.go
+++ b/pkg/glusterfs/api/types.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"regexp"
 	"sort"
+	"time"
 
 	"github.com/go-ozzo/ozzo-validation"
 	"github.com/go-ozzo/ozzo-validation/is"
@@ -677,4 +678,17 @@ func ValidateIds(v interface{}) error {
 		}
 	}
 	return nil
+}
+
+type NodeHealthStatus struct {
+	NodeId     string    `json:"node"`
+	Host       string    `json:"host"`
+	Up         bool      `json:"online"`
+	LastUpdate time.Time `json:"updated"`
+}
+
+type Healthz struct {
+	DbState      string             `json:"db-state"`
+	NodesStatus  []NodeHealthStatus `json:"nodes"`
+	ClusterHeath string             `json:"cluster-state,omitempty"`
 }
--- a/apps/glusterfs/app.go
+++ b/apps/glusterfs/app.go
@@ -650,6 +650,13 @@ func (a *App) SetRoutes(router *mux.Router) error {
 			Method:      "GET",
 			Pattern:     "/internal/state/examine/gluster",
 			HandlerFunc: a.ExamineGluster},
+
+		// Healthz
+		rest.Route{
+			Name:        "Healthz",
+			Method:      "GET",
+			Pattern:     "/healthz",
+			HandlerFunc: a.Healthz},
 	}
 
 	// Register all routes from the App


apps/glusterfs/health.go
```diff
@@ -0,0 +1,123 @@
+package glusterfs
```
phlogistonjohn: This file looks to be missing the normal preamble we have in the source files in heketi.

apps/glusterfs/health.go
```diff
@@ -0,0 +1,123 @@
...
+	"github.com/lpabon/godbc"
+)
+
+var clusterHealth = "ClusterHealth"
+
+type dbchecker struct {
+	CurrentTime time.Time
+}
+
+func NewHealthz() *api.Healthz {
```
phlogistonjohn: This function seems to be in an odd location between the type definition and that type's functions. Let's move it so it doesn't disrupt the flow when reading the code as much.

apps/glusterfs/health.go
```diff
@@ -0,0 +1,123 @@
...
+	return buffer.Bytes(), err
+}
+
+func (d *dbchecker) Unmarshal(buffer []byte) error {
+	dec := gob.NewDecoder(bytes.NewReader(buffer))
+	return dec.Decode(d)
+}
+
+func (*dbchecker) BucketName() string {
+	return BOLTDB_BUCKET_DBATTRIBUTE
```
phlogistonjohn: If this is using the BOLTDB_BUCKET_DBATTRIBUTE bucket why isn't it using the DbAttributeEntry type? I really don't like the idea of multiple types sharing the same bucket. It will just lead to confusion IMO.

apps/glusterfs/health.go
```diff
@@ -0,0 +1,123 @@
...
+	return EntryLoad(tx, d, clusterHealth)
+}
+
+func (a *App) Healthz(w http.ResponseWriter, r *http.Request) {
+	info := NewHealthz()
+
+	w.Header().Set("Content-Type", "application/json; charset=UTF-8")
+
+	//db Read check
+	err := a.db.View(func(tx *bolt.Tx) error {
```
phlogistonjohn: All of the db checking logic should be encapsulated in it's own function call so that the actual rest endpoint function call is simply formatting and reporting on the status of the various checks but not doing all of them directly.

apps/glusterfs/health.go
```diff
@@ -0,0 +1,123 @@
...
+	a.db.Update(func(tx *bolt.Tx) error {
+		h := dbchecker{CurrentTime: time.Now()}
+		return h.Write(tx)
+	})
+	if err == nil {
+		info.DbState = api.DbWrite.String()
+	}
+	//cluster status check
+	err = a.db.View(func(tx *bolt.Tx) error {
+		health := NewNodeHealthCache(0, 0, a.db, a.executor)
```
phlogistonjohn: Don't create a new health status cache. We should re-use the one previously set up when the app is set up. If the cache is disabled, simply report an empty but non-failure result.

apps/glusterfs/health_cache.go
```diff
@@ -171,6 +172,10 @@ func (hc *NodeHealthCache) toProbe() ([]*NodeHealthStatus, error) {
 
 func (s *NodeHealthStatus) update(e executors.Executor) {
 	// TODO: add ability to skip check if node was already recently checked
+
+	if !s.Up {
+		return
+	}
```
phlogistonjohn: This doesn't make any sense to me. If the node is up and then goes down later how will ever detect that?

apps/glusterfs/health_cache.go
```diff
@@ -154,14 +154,15 @@ func (hc *NodeHealthCache) toProbe() ([]*NodeHealthStatus, error) {
...
-			}
 			nhs := &NodeHealthStatus{
 				NodeId: nodeId,
 				Host:   node.Info.Hostnames.Manage[0],
 			}
+			// Ignore if the node is not online
+			if !node.isOnline() {
+				continue
+			}
+			nhs.Up = true
```
phlogistonjohn: I don't understand this change. What was wrong with the previous behavior? How does this change improve it? Because this is a subtle thing this is the kind of change I'd prefer to see in a separate commit with it's own commit message explaining the change (assuming it is still needed).

pkg/glusterfs/api/types.go
```diff
@@ -693,3 +694,53 @@ func ValidateIds(v interface{}) error {
...
 }
+
+type NodeHealthStatus struct {
+	NodeId     string    `json:"node"`
+	Host       string    `json:"host"`
+	Up         bool      `json:"online"`
+	LastUpdate time.Time `json:"updated"`
+}
+
+type DbState uint8
```
phlogistonjohn: I don't think this kind of detail is desirable in the api. Just have an enumerated string type with the ok/faulty info needed. That seems where you were going with the Healthz type, so I suggest simply removing this type from the api, and if you still think you want these types, locate them inside the app/glusterfs (for example).
The important part is not to put anything extra in the API.

pkg/glusterfs/api/types.go
```diff
@@ -693,3 +694,53 @@ func ValidateIds(v interface{}) error {
...
+	switch d {
+	case DbRead:
+		state = "read"
+	case DbWrite:
+		state = "write"
+	}
+	return state
+}
+
+type ClusterState uint8
```
phlogistonjohn: See above.

apps/glusterfs/health.go
Madhu-1: add license  path

apps/glusterfs/health.go
Madhu-1: removed

apps/glusterfs/health.go
Madhu-1: create new db attribute for healthz

apps/glusterfs/health.go
Madhu-1: done

apps/glusterfs/health.go
Madhu-1: sorry I missed that we  are storing  the heathcache in app, updated code

apps/glusterfs/health_cache.go
Madhu-1: i have added  it to todo list to make seperate commit,

currently, we  are saving only online node details  in the slice, for the health
implementation, we need  all nodes  status  irrespective of its state to decide whether
the cluster state is healthy or not.

apps/glusterfs/health_cache.go
Madhu-1: updated code,  only if the node is marked as up, then only checking the glusterd status or else just  updating the state

pkg/glusterfs/api/types.go
Madhu-1: removed

apps/glusterfs/health_cache.go
phlogistonjohn: Ok, makes sense. I do wonder if it would be beneficial to keeping two slices, one for up/healthy nodes and one for down/unhealthy nodes. That way when one needs only the list of healthy nodes code only needs to consult one slice w/o needing any additional conditionals.

apps/glusterfs/health_cache.go
phlogistonjohn: I'm still confused. In addition this appears to be a conditional that does nothing. Is the current version complete?

apps/glusterfs/healthz.go
```diff
@@ -0,0 +1,130 @@
...
+	return buffer.Bytes(), err
+}
+
+func (d *dbchecker) Unmarshal(buffer []byte) error {
+	dec := gob.NewDecoder(bytes.NewReader(buffer))
+	return dec.Decode(d)
+}
+
+func (*dbchecker) BucketName() string {
+	return BOLTDB_BUCKET_DBHEALTHZ
```
phlogistonjohn: To be honest I'm really not sure we want a special bucket just for tihs, especially since we're saving a single value. Wouldn't it be simpler and less code to simply set/get an new key under dbattributes?

@raghavendra-talur : I'm curious if you have an opinion about this. The current approach is functional but feels a little like overkill to me. Could I get a 2nd opinion from you?

apps/glusterfs/healthz.go
```diff
@@ -0,0 +1,130 @@
...
+
+	if len(health.nodes) == 0 {
+		return
+	}
+	totalUnhealthyNodes := 0
+	for _, n := range health.nodes {
+		if !n.Up {
+			totalUnhealthyNodes++
+		}
+		info.NodesStatus = append(info.NodesStatus, (api.NodeHealthStatus)(*n))
```
phlogistonjohn: Are the parens around api.NodeHeathStatus needed? It makes it look like a C type cast but it doesn't look like idomatic go to me.

apps/glusterfs/healthz.go
```diff
@@ -0,0 +1,130 @@
...
+
+	if !MonitorGlusterNodes {
+		if err := json.NewEncoder(w).Encode(info); err != nil {
+			panic(err)
+		}
+		return
+	}
+
+	// cluster status check
+	checkClusterState(a.nhealth, &info)
```
phlogistonjohn: I think this would read more logically to do the serialization once (at the bottom of the function) and move this line up into the if-staement (with an inverted condition) like so:
```golang
if MonitorGlusterNodes {
    checkClusterState(a.nhealth, &info)
}
// ... serialize response ...
```

apps/glusterfs/healthz.go
Madhu-1: yes this is needed for type casting

apps/glusterfs/healthz.go
Madhu-1: @raghavendra-talur let me  know your thoughts on this

phlogistonjohn:
@Madhu-1 do you plan to continue working on this PR?

Madhu-1:
@phlogistonjohn I was waiting for @raghavendra-talur comments on nhttps://github.com/heketi/heketi/pull/1417#discussion_r258905434

raghavendra-talur:
@Madhu-1 I am not 100% clear about "check the gluster nodes are reachable" check.

- What is the need for the check
- Why are we looking at nodes without giving importance to cluster they belong?

raghavendra-talur:
> @Madhu-1 I am not 100% clear about "check the gluster nodes are reachable" check.
> 
> * What is the need for the check

I got the answer to this in the issue, the requirement is "Check that executor has been able to communicate with nodes in some timeframe"
> * Why are we looking at nodes without giving importance to cluster they belong?


I am not sure of the approach where:
1. none of the nodes are reachable == faulty
2. at least one of the nodes is reachable == healthy

Probably a timestamp of last successful reply(not at /healthz endpoint) would be a better candidate.

Madhu-1:
@raghavendra-talur addressed review comments
 below is the response from curl request

```
curl http://127.0.0.01:8080/healthz |python -m json.tool
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   574  100   574    0     0  21259      0 --:--:-- --:--:-- --:--:-- 21259
{
    "cluster-state": "healthy",
    "db-state": "write",
    "nodes": [
        {
            "host": "192.168.10.101",
            "node": "252ced46d6cd6effb07a4688e72c829a",
            "online": true,
            "updated": "2019-03-27T15:59:15.063340783+05:30"
        },
        {
            "host": "192.168.10.102",
            "node": "ad4b1df157ac35018c6e9c21cb77ed5b",
            "online": true,
            "updated": "2019-03-27T15:59:15.4671574+05:30"
        },
        {
            "host": "192.168.10.100",
            "node": "b744aca95e4cd3b21ab6a92d497378f9",
            "online": false,
            "updated": "2019-03-27T15:59:18.559924198+05:30"
        },
        {
            "host": "192.168.10.103",
            "node": "d885c1bce818b58af247a7476b952209",
            "online": true,
            "updated": "2019-03-27T15:59:19.0096784+05:30"
        }
    ]
}
```

Madhu-1:
> > @Madhu-1 I am not 100% clear about "check the gluster nodes are reachable" check.
> > 
> > * What is the need for the check
> 
> I got the answer to this in the issue, the requirement is "Check that executor has been able to communicate with nodes in some timeframe"
> 
> > * Why are we looking at nodes without giving importance to cluster they belong?
> 
> I am not sure of the approach where:
> 
> 1. none of the nodes are reachable == faulty

earlier approach was if a number of not reachable nodes are greater than the total nodes, I was marking cluster as faulty, now changed to code to make cluster state as faulty if all nodes are not reachable
> 2. at least one of the nodes is reachable == healthy
> 
> Probably a timestamp of last successful reply(not at /healthz endpoint) would be a better candidate.

raghavendra-talur:
@Madhu-1 @phlogistonjohn I am happy with the PR as it stands right now. I am however, not clear on what I want from /healthz in particular.

Should healthz be local to heketi's status(that means only the db read/write status)?
Or should healthz of heketi represent the status of the storage in whole?

apps/glusterfs/healthz.go
phlogistonjohn: It's not. I just checked.

phlogistonjohn:
I've looked over much of the code again and my comments and I stand by pretty much all of what I requested previously from my "requested changes". This is also going to need tests.

Madhu-1:
@phlogistonjohn @raghavendra-talur currently I don't have enough bandwidth to work on this one, please feel free to take my patches if they are useful. Thank you :)

phlogistonjohn:
Effort is stale/abandoned. Closing.

PR #1427, opened 2018-11-15
apps: simplify logic a bit in BlockVolumeDestroy()

obnoxxx:
No semantic change. Just slight code improvement.


obnoxxx:
@phlogistonjohn good point, but I'd rather avoid comments and put the message in log instead, e.g. like so:

```golang
if e := json.Unmarshal([]byte(errOutput), &blockVolumeDelete); e != nil {
    if r.Err != nil {
       return logger.LogErr("Failed to unmarshal response from block volume delete for volume %v: %v", blockVolumeName, r.Err)
    }

    return logger.LogError("Failed to unmarshal response from blockvolume delete for volume %v: Unable to parse result: %v", blockVolumeName, e)
}
```

What do you think?

phlogistonjohn:
> 
> What do you think?

Usually, I'd be OK with this but in this case I'm wary of changing the error type on the return statement. Change it back to `return err` and log above and I'll be OK with it.

obnoxxx:
@phlogistonjohn I changed it as suggested. PTAL (again)

obnoxxx:
I kept them separate so we could have that discussion.
I would keep the first patch separate, because that's just the logic streamlining, and squash the other two, which add debug messages. Sounds OK?

phlogistonjohn:
Yes

obnoxxx:
@phlogistonjohn squashed as discussed and rebased

phlogistonjohn:
retest this please

obnoxxx:
retest this please

obnoxxx:
yay, all passed.
@phlogistonjohn can we merge?

phlogistonjohn:
Yes. :-)

PR #1428, opened 2018-11-16
metrics: add device metrics in bytes format

Madhu-1:
updated heketi to add support for displaying device metrics in bytes format.
device total space, device free space and device used space will be available in bytes
format for consumers.

Fixes: [#BZ1640027](https://bugzilla.redhat.com/show_bug.cgi?id=1640027)


obnoxxx:
ok to test

phlogistonjohn:
Can we please add a comment to the "old" metrics to say that they are deprecated and there primarily for backwards compat. with the external Heketi Metrics Exporter?

phlogistonjohn:
Thanks. I would have preferred a comment in the code but this is sufficient.

phlogistonjohn:
Dropping this link here in case I lose it again. https://prometheus.io/docs/practices/naming/#base-units
I checked the naming of the new params and these appear to match the doc. Thanks for the patch!

PR #1430, opened 2018-11-17
convert OpCounter to an OpTracker 

phlogistonjohn:

The OpTracker records (in memory) what operations are in flight by
unique ID (or temporary token). Currently this is a lot of code change
for little benefit but in the future this can be used to:
* Report on the "liveness" of individual ops
* Prioritize some ops based on tracker class
* Maybe things I haven't dreamed of yet

### Notes for the reviewer

These are prerequisite changes for the auto-cleanup effort, issue #1401


phlogistonjohn:
retest this please

phlogistonjohn:
@obnoxxx @raghavendra-talur @Madhu-1 @nixpanic @ansiwen PTAL if possible, thanks.

apps/glusterfs/operations_manage.go
```diff
@@ -14,60 +14,96 @@ import (
...
+	TrackToken
 )
 
-// OpCounter is used to track and manage how many operations are being
+// OpTracker is used to track and manage how many operations are being
 // processed by the server.
-type OpCounter struct {
+type OpTracker struct {
 	// configuration
 	Limit uint64
```
Madhu-1: can this be converted to lower case?

apps/glusterfs/operations_manage.go
```diff
@@ -14,60 +14,96 @@ import (
...
+// the operation and returns false. ThrottleOrAdd exists to perform
 // the check and set atomically.
-func (oc *OpCounter) ThrottleOrInc() bool {
-	oc.lock.Lock()
-	defer oc.lock.Unlock()
-	if oc.value >= oc.Limit {
+func (ot *OpTracker) ThrottleOrAdd(id string, c OpClass) bool {
+	ot.lock.Lock()
+	defer ot.lock.Unlock()
+	n := len(ot.normalOps)
```
Madhu-1: cant we make use of `ot.Get()` here to avoid write lock in the whole function execution?

apps/glusterfs/operations_manage.go
```diff
@@ -14,60 +14,96 @@ import (
...
+	if uint64(n) >= ot.Limit {
 		logger.Warning(
-			"operations in-flight (%v) exceeds limit (%v)", oc.value, oc.Limit)
+			"operations in-flight (%v) exceeds limit (%v)", n, ot.Limit)
 		return true
 	}
-	oc.value++
+	ot.insert(id, c)
```
Madhu-1: same here can we make use of `ot.Add(id,c)`?

apps/glusterfs/operations_manage.go
phlogistonjohn: No, the action of throttle or add must be atomic and so the check must be under the same lock as any changes we make.

apps/glusterfs/operations_manage.go
phlogistonjohn: Possibly, but I'm not sure what the value of making it private would be. Do you see the limit value as an implementation detail that should not be exposed?

apps/glusterfs/operations_manage.go
phlogistonjohn: No, see above.

obnoxxx:
centos-ci failed due to some problem brining up vms. It had passed before, and afterwards, only a doc patch has been merged. ignoring centos-ci failure.

phlogistonjohn:
>  I have one question but that can be clarified later: What is the point of OpClass? (Two values TrackNormal and TrackToken.) This is passed in, eventually to insert, and there it is ignored.. Maybe for a future patch to do something with it?

Exactly, the idea behind the OpClass was to able, but not require, different service levels for the tracked items. In the future an OpClass for background clean ups will be added. This class never "blocks" normal operations but if the system is busy with normal operations this background cleanup class is throttled.

Issue #1432, opened 2018-11-20
Builds failing at prebuild step for registry.centos.org/gluster/storagesig-heketi:testing


bamachrn:
### Kind of issue

Bug

### Observed behavior

[centos] Running prebuild shell script
+ sh ./prepare-build.sh
cp: cannot stat './../fromsource/heketi.json': No such file or directory

Source repo is not getting ready to build, as a result builds are failing.

### Expected/desired behavior

prebuild script should pass to get the repo ready to build the image.

### Details on how to reproduce (minimal and precise)

this builds are running from the container-index entry:
https://github.com/CentOS/container-index/blob/master/index.d/gluster.yaml#L13-L25

```
 - id: 2
    app-id: gluster
    job-id: storagesig-heketi
    git-url: https://github.com/heketi/heketi
    git-branch: master
    git-path: extras/docker/centos
    target-file: Dockerfile.testing
    desired-tag: testing
    notify-email: heketi-devel@gluster.org
    build-context: ./
    prebuild-script: ./prepare-build.sh
    prebuild-context: extras/docker/centos
    depends-on: centos/centos:latest
```

### Information about the environment:

building the image in CentOS Community Container Pipeline service

phlogistonjohn:
Stale issue. Closing.

PR #1434, opened 2018-11-20
docker/centos: container files get installed by 'make install'

nixpanic:

There is no need to manually provide the `heketi.json` configuration and
`heketi-start.sh` script. These are installed with `make install` since
commit 0c12c6abf3.

The container image builds for the CentOS Container Registry call
`prepare-build.sh` which used to copy the files to the container build
root. Now these files do not exist anymore, the preparation fails and
the (:testing) container does not get updated anymore.

### Does this PR fix issues?

Fixes: #1432

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -167,6 +167,8 @@ bindir:=$(prefix)/bin
 datarootdir:=$(prefix)/share
 hekdir:=$(datarootdir)/heketi
 mandir:=$(datarootdir)/man
+heketi_start_path:=$(hekdir)/container/heketi-start.sh
+heketi_config_path:=$(hekdir)/container/heketi.json
 
 INSTALL:=install -D -p
 INSTALL_PROGRAM:=$(INSTALL) -m 0755
@@ -179,9 +181,9 @@ install:
 	$(INSTALL_DATA) docs/man/heketi-cli.8 \
 		$(DESTDIR)$(mandir)/man8/heketi-cli.8
 	$(INSTALL_PROGRAM) extras/container/heketi-start.sh \
-		$(DESTDIR)$(hekdir)/container/heketi-start.sh
+		$(DESTDIR)$(heketi_start_path)
 	$(INSTALL_DATA) extras/container/heketi.json \
-		$(DESTDIR)$(hekdir)/container/heketi.json
+		$(DESTDIR)$(heketi_config_path)
 
 
 .PHONY: server client test clean name run version release \
--- a/extras/docker/centos/prepare-build.sh
+++ b/extras/docker/centos/prepare-build.sh
@@ -1,15 +1,11 @@
 #!/bin/sh
 #
-# Preparation for building container images.
+# This script is now deprecated and should get removed once the CentOS
+# Community Container Pipeline does not point to it anymore.
 #
-# - artifacts need to be in this WORKDIR, symlinks won't work
-#
+# Once the :testing container builds successfully again, the prebuild-script
+# can be dropped from
+# https://github.com/CentOS/container-index/blob/master/index.d/gluster.yaml
+# and after that has been merged, this script can be deleted.
 
-# error out if anything fails
-set -e
-
-# the local directory where the Dockerfile is located
-WORKDIR=$(dirname "${0}")
-
-cp -f "${WORKDIR}"/../fromsource/heketi.json "${WORKDIR}"
-cp -f "${WORKDIR}"/../fromsource/heketi-start.sh "${WORKDIR}"
+true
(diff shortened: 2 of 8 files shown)


extras/docker/centos/prepare-build.sh
```diff
@@ -11,5 +11,6 @@ set -e
 # the local directory where the Dockerfile is located
 WORKDIR=$(dirname "${0}")
 
-cp -f "${WORKDIR}"/../fromsource/heketi.json "${WORKDIR}"
-cp -f "${WORKDIR}"/../fromsource/heketi-start.sh "${WORKDIR}"
+# We used to copy files into the working directory. These files are now
+# included when 'make install' is run.
+true
```
Madhu-1: now this file does nothing, do we need this file anymore?

nixpanic: We can drop it, but the CentOS Container Pipeline still calls out to it. Only once the configuration has been adjusted there, this file can be removed.

nixpanic:
@centos-ci retest this please

extras/docker/centos/prepare-build.sh
phlogistonjohn: Understandable. Could we edit the comment to mention that the script is deprecated and/or obsolete and only exists for compatibility with the CentOS Container Pipeline? Perhaps, also mention that the script can be removed once the container pipeline is updated?

Makefile
```diff
@@ -181,7 +181,7 @@ install:
 	$(INSTALL_PROGRAM) extras/container/heketi-start.sh \
 		$(DESTDIR)$(hekdir)/container/heketi-start.sh
 	$(INSTALL_DATA) extras/container/heketi.json \
-		$(DESTDIR)$(hekdir)/container/heketi.json
+		$(DESTDIR)/etc/heketi/heketi.json
```
phlogistonjohn: I'm surprised to see this change and it appears to be undoing a pattern I was trying to establish. The pattern was twofold: 1) that paths are override-able using variables  2) the files "installed" to the system are "examples" that can then be copied or linked by an RPM spec file or Dockerfile or what not.
I'd like to see this bit dropped but add a line to the dockerfile to copy the "/usr/share/heketi/container/heketi.json" to /etc/heketi/heketi.json and similarly copy the heket-start.sh to /usr/bin/

extras/docker/rpi/Dockerfile
```diff
@@ -8,7 +8,7 @@ LABEL description="Heketi Container for Raspberry Pi"
 ADD ./heketi /usr/bin/heketi
 ADD ./heketi-cli /usr/bin/heketi-cli
 ADD ../../container/heketi.json /etc/heketi/heketi.json
-ADD ../../conatiner/heketi-start.sh /usr/bin/heketi-start.sh
```
phlogistonjohn: Whoops! :-)

extras/docker/centos/prepare-build.sh
nixpanic: I'll replace the comments in the script with a clearer explanation.

Makefile
nixpanic: Packaging is one of the most annoying tasks (IMHO). And having to do the same steps in many different locations is error-prone. I'll do the copy for `/etc/heketi/heketi.json`, but do not see a reason to copy `heketi-start.sh` instead of just using it from the existing location.

nixpanic:
All corrections have been made, please have a look again. Thanks!

Makefile
phlogistonjohn: That's a fine & good point. Would you be open to doing something more like:
```Makefile
heketi_start_path:=$(DESTDIR)$(hekdir)/container/heketi-start.sh
heketi_config_path:=$(DESTDIR)$(hekdir)/container/heketi.json
# ...
	$(INSTALL_PROGRAM) extras/container/heketi-start.sh $(heketi_start_path)		
	$(INSTALL_DATA) extras/container/heketi.json $(heketi_config_path)

```
And then you'd be able to set the _path variables as needed in "make install"?

nixpanic:
@phlogistonjohn you have reviewed this before, care to check it again?

phlogistonjohn:
> @phlogistonjohn you have reviewed this before, care to check it again?

Please correct me if I am wrong but I don't think you addressed my comment from Nov 23, 2018 on the Makefile.

nixpanic:
@phlogistonjohn now those changes are there too!

obnoxxx:
How does this related to PR #1407? It certainly supersedes that one.

phlogistonjohn:
Who wants to remove this Dockerfile from the heketi repo?
* Me
* @obnoxxx ?

I'm open to creating another repo under the heketi org for this centos stuff but I no longer think repo is the right place for this.

Regardless, I'd really like to see some motion on this PR.

phlogistonjohn:
No response to my change request for around five months. Since I think this is not the right approach and the heketi maintainers don't want to be responsible for builds we have no control over I will probably my own pr to remove these extra builds.

PR #1435, opened 2018-11-21
remoteexec: use error output for error contents

phlogistonjohn:

Historically, heketi used to convert the stderr of commands to go error
type and discard the originating error. The remoteexec library however
goes to lengths to retain all the information about a command including
the error output (stderr) and originating error. But most of the
preexisting heketi code wants to check the error output not the
originating error.

This change allows treating a Result type as an error, looking for error
output if it exists and then falling back to the content of the
originating error. When the adapter functions FirstErrorIndexed,
FirstError, and AnyError were added the important ErrorOutput was
forgotten and the originating error was used. Now it returns the
result-as-error. This change fixes at least one regression where the
error response of "gluster volume delete" was being scraped.

### Notes for the reviewer

Currently, there's no test for this because the error handling that was failing happens in the executor layers which don't get executed with the unit tests in app/glusterfs, so there's no value in trying to write a unit test for this. As for functional tests, .... we do have tests for the volume-already-deleted case but this goes through a check for snapshots first, and this code already uses the "new api" rather than the AnyError adapter function. The problem shows up in the cleanup path which skips checking for snapshots and immediately tries to destroy the volume. This was revealed when adding new tests with the new support for error injection. I wanted to get the PR out there to cut down on the chances for regressions on master sooner than those PRs will be ready.

### DIFF
--- a/pkg/remoteexec/result.go
+++ b/pkg/remoteexec/result.go
@@ -29,6 +29,21 @@ func (r Result) Ok() bool {
 	return r.Completed && r.Err == nil && r.ExitStatus == 0
 }
 
+// Error allows an error result to be treated like an error type.
+// Do not assume all results are true errors.
+func (r Result) Error() string {
+	// to be more compatible with older versions of heketi we use
+	// the error output (stderr) of commands unless that
+	// string is empty
+	if r.ErrOutput != "" {
+		return r.ErrOutput
+	}
+	if r.Err != nil {
+		return r.Err.Error()
+	}
+	return ""
+}
+
 // Results is used for a grouping of Result items.
 type Results []Result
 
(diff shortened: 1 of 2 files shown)


phlogistonjohn:
Needed for Issue #1401

pkg/remoteexec/result.go
```diff
@@ -73,7 +88,7 @@ func (rs Results) FirstErrorIndexed() (int, error) {
 			continue
 		}
 		if r.Err != nil {
```
obnoxxx: Should this be checking r.Error() instead? Like:
```golang
if r.Error() != "" {
    return i, r
}
```

obnoxxx:
retest this please

PR #1436, opened 2018-11-21
apps: support exporting (dumping) db to stdout

phlogistonjohn:

This change allows the DbDump function to support the common cli
convention of treating a single dash "-" as a signal to write
to stdout rather than a file name. This allows us to run the
command `heketi db export --dbfile heketi.json --jsonfile -`
and read the json content on stdout.

### Notes for the reviewer

Can be useful for debugging real systems as well as potentially useful in tests.


obnoxxx:
again the weird flaky block failing stuff...

PR #1441, opened 2018-11-27
apps: on startup, do not try to "fix" pending BHVs

phlogistonjohn:

The db update code triggered during start up was incorrectly trying
to fix the size accounting of a pending block hosting volume. This
change disables the ability of the fix up code path to change
pending volumes and fixes the cause of the incorrect calculation.

### Does this PR fix issues?

Because this issue only impacts pending operations it currently doesn't do much because pending operations must be cleaned up manually.
However, this PR is a prerequisite of  #1401 in order to properly set the size of the BHV during cleanup.

### Notes for the reviewer

This issue was found while preparing tests for auto-cleanup code. Because the test suite can not currently test for error conditions, no tests have been added yet for this change.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: go
+dist: xenial
 sudo: true
 before_install:
     - sudo apt-get update -qq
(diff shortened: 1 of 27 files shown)


obnoxxx:
centos ci failed brining up the vms...

obnoxxx:
retest this please

obnoxxx:
@centos-ci retest this please

obnoxxx:
retest this please

PR #1444, opened 2018-11-28
systemd: provide a heketi.env file with possible variables

nixpanic:

The environment variables that can be set to influence the behaviour of
Heketi are not well documented. This is an attempt to capture the
variables that can be changed for the benefit of the heketi.service
systemd unit.

The systemd unit already tries to read variables from
/etc/heketi/heketi.env, but this file is not provided anywhere. If users
have a need to change the environment, they will need to create the file
and not mistype the name of the variable. Providing a file with the
defaults makes it less error prone.

### Notes for the reviewer

Some of the variables have been left out intentionally:

- admin and user keys: these should be set in the heketi.json config file. Passing them as an environment variable to non-containerized processes seems insecure (thanks @raghavendra-talur)
- heketi as a service started by systemd does not interact with kubernetes, these variables are not used

### DIFF
--- a/extras/systemd/heketi.env
+++ b/extras/systemd/heketi.env
@@ -1,0 +1,60 @@
+#
+# These environment variables will be loaded by systemd in case heketi is
+# started as a service (as opposed to running in a container).
+#
+# See the actual /etc/heketi/heketi.json configuration file for more options.
+#
+# The variables commented out show the default values. Variables not set here
+# might be set in the /etc/heketi/heketi.json configuration file.
+#
+
+# Database file name
+#HEKETI_DB_PATH=/var/lib/heketi/heketi.db
+
+# fstab file on nodes
+#HEKETI_FSTAB=/etc/fstab
+
+# Set log level. Choices are:
+#   none, critical, error, warning, info, debug
+#HEKETI_GLUSTERAPP_LOGLEVEL=warning
+
+# Heketi Server Port Number
+#HEKETI_HTTP_PORT=8080
+
+# Allow starting heketi when there are stale pending ops being in the db.
+#HEKETI_IGNORE_STALE_OPERATIONS=0
+
+# The maximum limit of snapshots. If no limit has been specified, then Heketi
+# will not set a limit and GlusterFS will use its default value.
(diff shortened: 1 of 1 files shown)


extras/systemd/heketi.env
```diff
@@ -0,0 +1,70 @@
...
+# options can be added next to it separated by a comma.
+#HEKETI_BLOCK_HOSTING_VOLUME_OPTIONS="group gluster-block"
+
+# Execute plugin. Possible choices:
+# mock: This setting is used for testing and development. It will not send
+#       commands to any node.
+# ssh:  This setting will notify Heketi to ssh to the nodes. It will need the
+#       values in sshexec to be configured.
+# kubernetes: Communicate with GlusterFS containers over Kubernetes exec api.
+#HEKETI_EXECUTOR=mock
```
phlogistonjohn: We should either drop this or not document mock here. I'd go for dropping it because this is really something someone should configure in the config file, if using heketi via systemd. And in that case there's almost only one choice: ssh.

extras/systemd/heketi.env
```diff
@@ -0,0 +1,70 @@
...
+#HEKETI_SSH_KEYFILE="path/to/private_key"
+#HEKETI_SSH_PORT=22
+#HEKETI_SSH_USER=sshuser
+
+# Rebalance on expansion is done by default, but it can be disabled if needed.
+#HEKETI_GLUSTERAPP_REBALANCE_ON_EXPANSION=1
+
+# The node health monitor is enabled by default. It can be disabled for
+# debugging.
+#HEKETI_DISABLE_HEALTH_MONITOR=0
```
phlogistonjohn: Lets drop this. User's shouldn't really be messing with this as its mostly for running heketi under test/deep debugging.

extras/systemd/heketi.env
nixpanic: Agreed!

extras/systemd/heketi.env
nixpanic: Ok, will do.

nixpanic:
This was updated according to the comments from @phlogistonjohn. @heketi/maintainers please have a look again :)

PR #1446, opened 2018-11-28
auto cleanup pt2

phlogistonjohn:

This PR implements a good chunk of the code needed to actually make operations cleanable. This is where the bulk of the refactorings to app/glusterfs lie as far as operations are concerned. The general gist is that you'll see:
* Various refactorings to make clean up code reasonable: including splitting out db access from executor stuff
* Making various operations loadable
* Making various operations cleanable
* Unit tests for the above
* An offline mode for cleaning: `heketi offline cleanup-operations ...` which can be run when heketi server is not running.

The latter is mainly included in this PR (as opposed to a latter PR) as a demonstration of things end-to-end.

Does this seem like a lot? There's more to come!

### Does this PR fix issues?

Part of issue #1401

### Notes for the reviewer

This looks like a lot of commits but hopefully these are small and reviewable commits that won't make your brain explode. :-)

This PR builds on top of PR #1445 please review and merge that series first.

### DIFF
--- a/apps/glusterfs/operations_block_volume.go
+++ b/apps/glusterfs/operations_block_volume.go
@@ -24,7 +24,6 @@ type BlockVolumeCreateOperation struct {
 	OperationManager
 	noRetriesOperation
 	bvol *BlockVolumeEntry
-	//vol *VolumeEntry
 }
 
 // NewBlockVolumeCreateOperation  returns a new BlockVolumeCreateOperation  populated
@@ -242,7 +241,15 @@ func (bvc *BlockVolumeCreateOperation) Rollback(executor executors.Executor) err
 	if err != nil {
 		return err
 	}
-	if e := bvc.bvol.cleanupBlockVolumeCreate(bvc.db, executor); e != nil {
+	hvname, err := bvc.bvol.blockHostingVolumeName(bvc.db)
+	if err != nil {
+		return err
+	}
+	err = bvc.bvol.deleteBlockVolumeExec(bvc.db, hvname, executor)
+	if err != nil {
+		return err
+	}
+	if e := bvc.bvol.removeComponents(bvc.db, false); e != nil {
 		return e
 	}
 	if vol != nil {
--- a/apps/glusterfs/operations_block_volume.go
+++ b/apps/glusterfs/operations_block_volume.go
@@ -43,6 +43,30 @@ func NewBlockVolumeCreateOperation(
 	}
 }
 
+// loadBlockVolumeCreateOperation returns a BlockVolumeCreateOperation populated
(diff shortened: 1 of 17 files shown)


phlogistonjohn:
retest this please

phlogistonjohn:
retest this please

obnoxxx:
retest this please

obnoxxx:
retest this please

apps/glusterfs/operations_cleanup_test.go
```diff
@@ -70,3 +70,80 @@ func TestBasicOperationsCleanup(t *testing.T) {
...
+	req.Durability.Type = api.DurabilityReplicate
+	req.Durability.Replicate.Replica = 3
+
+	// create a volume we can delete later
+	vol := NewVolumeEntryFromRequest(req)
+	vc := NewVolumeCreateOperation(vol, app.db)
+	e := RunOperation(vc, app.executor)
+	tests.Assert(t, e == nil, "expected e == nil, got", e)
+
+	app.db.Update(func(tx *bolt.Tx) error {
```
obnoxxx: This should probably be `View()` instead of `Update()`, but it does not harm.

apps/glusterfs/operations_cleanup_test.go
```diff
@@ -147,3 +147,80 @@ func TestOperationsCleanupThreeOps(t *testing.T) {
...
+	req.Durability.Replicate.Replica = 3
+
+	// create a volume we can delete later
+	vol := NewVolumeEntryFromRequest(req)
+	vc := NewVolumeCreateOperation(vol, app.db)
+	e := RunOperation(vc, app.executor)
+	tests.Assert(t, e == nil, "expected e == nil, got", e)
+
+	var deviceId string
+	app.db.Update(func(tx *bolt.Tx) error {
```
obnoxxx: `View()` should be sufficient?

obnoxxx:
retest this please

apps/glusterfs/operations_block_volume.go
```diff
@@ -431,7 +459,20 @@ func (vdel *BlockVolumeDeleteOperation) Finalize() error {
 			return e
 		}
 
-		vdel.op.Delete(tx)
-		return nil
+		return vdel.op.Delete(tx)
```
obnoxxx: So this is a difference in behavior of the `Finalize()` function in that now error of the `Delete` op will result in failure of `Finalize`. Is this intended? Was that an oversight before? What has changed?

phlogistonjohn: It was an oversight before.

PR #1448, opened 2018-12-03
Add a configurable timeout to gluster cli, defaulting to 10 minutes

### LINKED ISSUE #1426: Investigate changing gluster cli timeout value
### Kind of issue

Feature request

### Expected/desired behavior

Because heketi is ansync by nature and commands are not being monitored by humans the timeout used for gluster (cli) commands to talk wiith glusterd by default may be more useful with a larger value when run by heketi.

obnoxxx:

This PR adds a configurable timeout to invocations of the Gluster CLI, set to 10 minutes by default. The reason for this is that in a very busy system, with many volumes, some gluster cli commands can take longer than the default timeout of 2 minutes.

This adds a config setting of `gluster_cli_timeout` that can be overridden by the env variable `HEKETI_GLUSTER_CLI_TIMEOUT` and defaults to 10 minutes. This seemed to have been a good default value from the observations. This value is passed to the gluster cli commandline with the `--timeout` parameter.

### Does this PR fix issues?

Fixes #1426 

RHBZ: https://bugzilla.redhat.com/show_bug.cgi?id=1651041

### Notes for the reviewer

In order to not run into the `ExecCommands` timeout which is currently set to 10 minutes in the invocations, the timeout for the ExecCommands invocations is also made dynamic in that it is set to the maximum of 10 and the configured cli timeout (in minutes) +1, so that we will always wait for the gluster cli to time out rather than timing out the ssh connection first.

Only a very minimal amount of refactoring / reconsiliation has been done to the executor code as part of this PR, to avoid code duplication. More refactoring could be done as follow-up PRs.

### DIFF
--- a/executors/cmdexec/cmdexec.go
+++ b/executors/cmdexec/cmdexec.go
@@ -35,6 +35,10 @@ type CmdExecutor struct {
 	BackupLVM      bool
 }
 
+func (c *CmdExecutor) Init(config *CmdConfig) {
+	c.Throttlemap = make(map[string]chan bool)
+}
+
 func (s *CmdExecutor) AccessConnection(host string) {
 	var (
 		c  chan bool
--- a/executors/cmdexec/cmdexec.go
+++ b/executors/cmdexec/cmdexec.go
@@ -36,9 +36,14 @@ type CmdExecutor struct {
 	BackupLVM      bool
 }
 
+func setWithEnvVariables(config *CmdConfig) {
+}
+
 func (c *CmdExecutor) Init(config *CmdConfig) {
 	c.Throttlemap = make(map[string]chan bool)
 	c.config = config
+
+	setWithEnvVariables(config)
 }
 
 func (s *CmdExecutor) AccessConnection(host string) {
--- a/executors/cmdexec/cmdexec.go
+++ b/executors/cmdexec/cmdexec.go
@@ -10,6 +10,8 @@
 package cmdexec
 
 import (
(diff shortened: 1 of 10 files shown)


executors/cmdexec/cmdexec.go
```diff
@@ -35,6 +40,31 @@ type CmdExecutor struct {
...
+func (c *CmdExecutor) glusterCommand() string {
+	return fmt.Sprintf("gluster --mode=script --timeout=%v", c.GlusterCliTimeout())
+}
+
+func setWithEnvVariables(config *CmdConfig) {
+	var env string
+
+	env = os.Getenv("HEKETI_GLUSTER_CLI_TIMEOUT")
+	if env != "" {
+		value, err := strconv.ParseInt(env, 10, 32)
```
Madhu-1: can we use `ParseUint` instead of `ParseInt`?

obnoxxx: Good point. Will update the PR shortly.

Alternatively, we could discuss whether we want to just use `int` instead of `uint32` ...

obnoxxx:
What's going on with travis - is that a travis issue or something in this PR?

obnoxxx:
(restarted travis build)

phlogistonjohn:
I've done a quick scan and nothing jumps out at me as must-fix. Once we resolve the travis issues I'll give it a 2nd pass and ought to be able to approve.

PR #1449, opened 2018-12-03
travis: try building on xenial image, to resolve hg tls issues

phlogistonjohn:

On December 1st, [BitBucket disabled TLS 1.0 support](https://bitbucket.org/blog/deprecating-tlsv1-tlsv1-1-2018-12-01) which is causing problems for the old version of mercurial/python provided by the travis ci we are using.

This change bumps the distro version we request from travis to "xenial" where TLS > 1.0 does work. This also includes fixes for fallout from the version bump.

### Notes for the reviewer

All other PRs will continue to fail unless this PR is merged.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: go
+dist: xenial
 sudo: true
 before_install:
     - sudo apt-get update -qq
--- a/tests/200-py-unittest.sh
+++ b/tests/200-py-unittest.sh
@@ -39,9 +39,12 @@ fi
 
 TOXENVS="py27"
 if command -v python3.5 &>/dev/null; then
-	TOXENVS="${TOXENVS},py35"
 	if command -v pyenv &>/dev/null; then
-		pyenv global system 3.5
+		if pyenv global system 3.5; then
+			TOXENVS="${TOXENVS},py35"
+		fi
+	else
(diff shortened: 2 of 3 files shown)


tests/200-py-unittest.sh
```diff
@@ -39,9 +39,12 @@ fi
 
 TOXENVS="py27"
 if command -v python3.5 &>/dev/null; then
-	TOXENVS="${TOXENVS},py35"
 	if command -v pyenv &>/dev/null; then
```
raghavendra-talur: As someone who does not understand the python environment well, I would say it would be nice to document what is being checked here.

obnoxxx: I agree that an explaining comment would be nice. But would not even block on this.

.travis.yml
```diff
@@ -1,4 +1,5 @@
 language: go
+dist: xenial
```
obnoxxx: I wasn't aware that xenial is available! Is that recent?...

phlogistonjohn: No, it is around 2 years old (http://releases.ubuntu.com/xenial/)

tests/200-py-unittest.sh
phlogistonjohn: I'd like to do that in a followup pr tho.

PR #1450, opened 2018-12-04
apps: add cluster level error reporting to cluster selection

phlogistonjohn:

Use the previously created multi-cluster error type to gather
all the reasons for finding a cluster ineligible for volume
placement and return that error if no eligible clusters
were found. This should help users immediately diagnose some
common high-level error conditions related to volume names
or file and block flags vs. the cryptic "No space" error.

### Does this PR fix issues?

Fixes rhbz#1652876


apps/glusterfs/block_volume_entry.go
```diff
@@ -168,12 +168,15 @@ func (v *BlockVolumeEntry) eligibleClustersAndVolumes(db wdb.RODB) (
...
+	var cerr error
 	cr := ClusterReq{Block: true}
-	possibleClusters, e = eligibleClusters(db, cr, possibleClusters)
+	possibleClusters, cerr, e = eligibleClusters(db, cr, possibleClusters)
 	if e != nil {
 		return
-	}
-	if len(possibleClusters) == 0 {
+	} else if len(possibleClusters) == 0 && cerr != nil {
+		e = cerr
```
obnoxxx: Callers seem to converge the two errors like this. Does the `eligibleClusters` function need to provide both errors, or can that function actually to the convergence without loss of information, and thereby reducing boilerplate code in the callers?

phlogistonjohn: Thanks for the nudge, I've pushed the common error handling into the eligibleClusters function in a followup patch. PTAL.

obnoxxx: Thanks, this looks good, but wouldn't it be more natural to squash these?
(Otherwise the first patch changes the signature of `eligibleClusters` and the second patch changes it back to the original signature...

here is the diffstat of the two separate patches:
```
d85a5129 2019-01-18 15:38:28 -0500 John Mulligan  (origin/pr/1450) apps: further simplify error handling of eligibleClusters                                                                                                                  
 apps/glusterfs/block_volume_entry.go | 10 +---------
 apps/glusterfs/volume_entry.go       | 21 ++++++++++++---------
 2 files changed, 13 insertions(+), 18 deletions(-)

03d80d71 2018-11-30 11:22:03 -0500 John Mulligan  apps: add cluster level error reporting to cluster selection
 apps/glusterfs/block_volume_entry.go      |  9 +++++---
 apps/glusterfs/block_volume_entry_test.go |  6 ++++--
 apps/glusterfs/volume_entry.go            | 34 ++++++++++++++++++++++---------
 apps/glusterfs/volume_entry_test.go       | 17 ++++++++++------
 4 files changed, 45 insertions(+), 21 deletions(-)
```

and here's the diffstat of the compined diff:
```
 apps/glusterfs/block_volume_entry.go      |  5 -----
 apps/glusterfs/block_volume_entry_test.go |  6 ++++--
 apps/glusterfs/volume_entry.go            | 33 +++++++++++++++++++++++--------
 apps/glusterfs/volume_entry_test.go       | 17 ++++++++++------
 4 files changed, 40 insertions(+), 21 deletions(-)
```

phlogistonjohn:
> Please squash the two patches (as explained in my previous comment) or explain why they should stay separate. :-) Otherwise: LGTM

Heh. I only kept them separate so you could see the diff in response to your feedback. Squashed and pushed.

PR #1455, opened 2018-12-12
Implement sentinel errors as real constants

ansiwen:

The errors created with errors.New() can't be used in constant
values.  Hence the sentinel errors could be overwritten by accident.
However, pure strings can be stored as constant values.  This change
replaces the sentinel errors with an error type that is isomorph to a
pure string and consequently defines them as constant values.

Based on [this article](https://dave.cheney.net/2016/04/07/constant-errors).

Mutually exclusive with #1380 

### DIFF
--- a/apps/glusterfs/errors.go
+++ b/apps/glusterfs/errors.go
@@ -9,27 +9,29 @@
 
 package glusterfs
 
-import (
-	"errors"
-)
+type constErr string
 
-var (
-	ErrNoSpace          = errors.New("No space")
-	ErrFound            = errors.New("Id already exists")
-	ErrNotFound         = errors.New("Id not found")
-	ErrConflict         = errors.New("The target exists, contains other items, or is in use.")
-	ErrMaxBricks        = errors.New("Maximum number of bricks reached.")
-	ErrMinimumBrickSize = errors.New("Minimum brick size limit reached.  Out of space.")
-	ErrDbAccess         = errors.New("Unable to access db")
-	ErrAccessList       = errors.New("Unable to access list")
-	ErrKeyExists        = errors.New("Key already exists in the database")
-	ErrNoReplacement    = errors.New("No Replacement was found for resource requested to be removed")
-	ErrCloneBlockVol    = errors.New("Cloning of block hosting volumes is not supported")
+func (e constErr) Error() string {
+	return string(e)
+}
+
+const (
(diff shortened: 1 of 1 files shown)


phlogistonjohn:
I'm struggling to see what issues this PR solves. Overwriting the errors seems fairly unlikely to me.

ansiwen:
This is just spin-off code that I found from the times when I wrote the other error PRs and I thought I share it instead of throwing it away. It's certainly low prio. On the other hand, does it hurt? If we will go with the solution from #1380 you can just ignore this of course.

phlogistonjohn:
After giving this some more thought and re-reading the associated blog post I'm comfortable with this in the abstract. Here's what I'd like to see happen though: let's make the "ConstError" a new sub-package under pkg so that other parts of heketi (I'm specifically thinking of executors) can use this new type.
Then we can document the new type with some doc comments and a in-code link to the blog post. Having done that I'd think I'd be comfortable merging it.

phlogistonjohn:
@ansiwen ping. Please see me last comment.

ansiwen:
@phlogistonjohn sure, we can do that. but before putting effort into that, we should first decide, if we don't prefer replacing them with something that has back-traces to the point of where the error happened, like I proposed in #1380. I would prefer the latter, because it proved to be very helpful for me during debugging before. This PR is like a second choice for me, if we decide to not use the back-traceable errors package.

phlogistonjohn:
I honestly don't see the two as mutually exclusive, especially if we put the new type in it's own package and can thus the implementation can be changed more independently.

ansiwen:
I think I'm confused.

#1380 would replace every line of this PR, so how is it _not_ mutually exclusive? Maybe I don't understand correctly what you mean. So let me rephrase:

You are proposing to put the
```
type constErr string

func (e constErr) Error() string {
	return string(e)
}
```
into a file in pkg, but the actual instances of the errors from this PR still stay in the glusterfs/errors.go, right? (You couldn't put more into the pkg, like a New() function or so, since that doesn't work with constants. It would have to be created directly casting constant string into that type, like in this PR.)

So we would maybe later replace that with the errtag Instances, making the consterr pkg unused, no?

Or do you maybe suggest to implement the errtag in a way that they wrap arbitrary other errors, and we could use the consterr objects there?

phlogistonjohn:

PR #1460, opened 2018-12-17
Perform consistency checks on db

raghavendra-talur:
This PR implements consistency check for heketi db. The check can be performed in both offline and online mode. In online mode, a cli command or a REST call can be made to a heketi server. In offline version, `db check` subcommand is to be invoked for heketi binary with path to heketi.db. 

Consistency check only validates the information provided in the database and does not validate it against *real* state on the Gluster nodes and/or Kubernetes/Openshift.

### Notes for the reviewer
I did not find any value in making a version of this which takes a JSON file representing heketi database as input. Such scenarios can be handled by using the `db import` command first and then invoking a `db check` command later.

### DIFF
--- a/apps/glusterfs/dbcommon.go
+++ b/apps/glusterfs/dbcommon.go
@@ -33,6 +33,30 @@ type Db struct {
 	PendingOperations map[string]PendingOperationEntry `json:"pendingoperations"`
 }
 
+//DbBucketCheckResponse ... is summary of check on a db bucket.
+type DbBucketCheckResponse struct {
+	Total           int      `json:"total"`
+	Pending         int      `json:"pending"`
+	Ok              int      `json:"ok"`
+	NotOk           int      `json:"notok"`
+	Inconsistencies []string `json:"inconsistencies"`
+}
+
+//DbCheckResponse ... is the output of db check. It lists a summary of db state
+//and inconsistencies related to each bucket, if any.
+type DbCheckResponse struct {
+	Clusters             DbBucketCheckResponse `json:"clusters"`
+	Volumes              DbBucketCheckResponse `json:"volumes"`
+	Bricks               DbBucketCheckResponse `json:"bricks"`
+	Nodes                DbBucketCheckResponse `json:"nodes"`
+	Devices              DbBucketCheckResponse `json:"devices"`
+	BlockVolumes         DbBucketCheckResponse `json:"blockvolumes"`
+	DbAttributes         DbBucketCheckResponse `json:"dbattributes"`
+	PendingOperations    DbBucketCheckResponse `json:"pendingoperations"`
+	TotalInconsistencies int                   `json:"totalinconsistencies"`
+	Inconsistent         bool                  `json:"inconsistent"`
+}
+
 func initializeBuckets(tx *bolt.Tx) error {
 	// Create Cluster Bucket
 	_, err := tx.CreateBucketIfNotExists([]byte(BOLTDB_BUCKET_CLUSTER))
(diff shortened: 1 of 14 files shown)


raghavendra-talur:
@Madhu-1 @obnoxxx @phlogistonjohn @nixpanic I am still working on tests and upload in the same PR as an update. However, I would like to get your opinion on the output format and basic approach.

apps/glusterfs/dbcommon.go
```diff
@@ -33,6 +33,30 @@ type Db struct {
...
+	Clusters             DbBucketCheckResponse `json:"clusters"`
+	Volumes              DbBucketCheckResponse `json:"volumes"`
+	Bricks               DbBucketCheckResponse `json:"bricks"`
+	Nodes                DbBucketCheckResponse `json:"nodes"`
+	Devices              DbBucketCheckResponse `json:"devices"`
+	BlockVolumes         DbBucketCheckResponse `json:"blockvolumes"`
+	DbAttributes         DbBucketCheckResponse `json:"dbattributes"`
+	PendingOperations    DbBucketCheckResponse `json:"pendingoperations"`
+	TotalInconsistencies int                   `json:"totalinconsistencies"`
+	Inconsistent         bool                  `json:"inconsistent"`
```
phlogistonjohn: This parameter seems redundant, given that we can just check if TotalInconsistencies is non-zero.

raghavendra-talur:
@phlogistonjohn I am slowly understanding your approach to interfaces and I think I could have implemented a consistencyChecker interface for every *_entry . I have not done that here. If you think such refactoring can be done later, I would like to have a go at it. If you think, it is necessary to do it in current PR, let me know.

apps/glusterfs/pendingop_entry.go
```diff
@@ -380,3 +381,49 @@ func PendingOperationEntrySelection(
...
+		case 2:
+			fallthrough
+		case 4:
+			fallthrough
+		case 5:
+			fallthrough
+		case 9:
+			fallthrough
+		case 10:
+			fallthrough
```
phlogistonjohn: can't we write this as `case 2,4,5,...:` without fallthrough?
In addition I'd really rather we use the type names and not the raw values. So `OpAddBrick` rather than 1.

apps/glusterfs/volume_entry.go
```diff
@@ -1010,3 +1010,85 @@ func (v *VolumeEntry) cloneVolumeRequest(db wdb.RODB, clonename string) (*execut
 
 	return vcr, sshhost, nil
 }
+
+// consistencyCheck ... verifies that a volumeEntry is consistent with rest of the database.
+// It is a method on volumeEntry and needs rest of the database as its input.
+func (v *VolumeEntry) consistencyCheck(db Db) (consistent bool, inconsistencies []string) {
```
phlogistonjohn: I would have perferred that the consistencyCheck functions be functions rather than methods, but it's not a big deal.
What I don't understand is the need for a seperate bool to indicate an inconsistency when we can check check if the slice is zero length. Doing it with two params feels a bit non-DRY to me.

phlogistonjohn:
> I am slowly understanding your approach to interfaces and I think I could have implemented a consistencyChecker interface for every *_entry . I have not done that here. If you think such refactoring can be done later, I would like to have a go at it. If you think, it is necessary to do it in current PR, let me know.

I don't think that's needed for this PR. It might makes things look a bit more uniform, but it is not needed in the short term.

apps/glusterfs/dbcommon.go
raghavendra-talur: done

apps/glusterfs/pendingop_entry.go
raghavendra-talur: done

apps/glusterfs/volume_entry.go
raghavendra-talur: Removed the bool but we still have them as method.

phlogistonjohn:
retest this please

PR #1459, opened 2018-12-16
Fix lint issues

Madhu-1:
### What does this PR achieve? Why do we need it?
This PR fixes issues found in `gofmt` and correct misspelled word. 
### Does this PR fix issues?

Fixes #

PR #1461, opened 2018-12-18
apps: fix gofmt error found by go 1.11

obnoxxx:
Go 1.11 (on fedora 29) found this formatting error.

### DIFF
--- a/apps/glusterfs/app_test.go
+++ b/apps/glusterfs/app_test.go
@@ -183,6 +183,7 @@ func TestAppBlockSettings(t *testing.T) {
 		Executor:  "crazyexec",
 		Allocator: "simple",
 		DBfile:    "/path/to/nonexistent/heketi.db",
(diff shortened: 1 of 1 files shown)


Madhu-1:
@obnoxxxI have a similar patch which fixes formatting issue https://github.com/heketi/heketi/pull/1459

phlogistonjohn:
Yeah. Both of these PRs fail in travis because this is a change in gofmt. Perhaps it would be simpler to just add a comment on the line between DBfile and CreateBlockHostingVolumes, as I think that would make both versions of the formatter happy (I think, not tested).

obnoxxx:
@Madhu-1 oh, right, you have the same hunk in your PR. so we can dump this one.

@phlogistonjohn I tend to agree. But it seems like a bug in gofmt version older than 1.11 ... Silly to have to work around this...

obnoxxx:
Testing with a comment line inserted.
Works for me locally for go 1.11 ...

PR #1462, opened 2018-12-18
funtional tests: run TestErrorHandling by default

obnoxxx:
I found that the error handling tests were not run at all in the functional test runs... Fix this!


obnoxxx:
Yay, now the tests passed again, and the new tests ran! :-)

PR #1463, opened 2018-12-18
IO Alignment improvements

rsevilla87:
IO alignment should be configurable for RAID based architectures as stated by GlusterFS docs. This PR is intended to add a set of configurable flags relative to the LVM stack and XFS which allow to configure IO alignment properly. 
All of these flags are not required so current deployments would not be affected.

### DIFF
--- a/executors/cmdexec/device.go
+++ b/executors/cmdexec/device.go
@@ -43,7 +43,7 @@ func (s *CmdExecutor) DeviceSetup(host, device, vgid string, destroy bool) (d *e
 		logger.Info("Data on device %v (host %v) will be destroyed", device, host)
 		commands = append(commands, fmt.Sprintf("wipefs --all %v", device))
 	}
-	commands = append(commands, fmt.Sprintf("pvcreate -qq --metadatasize=128M --dataalignment=256K '%v'", device))
+	commands = append(commands, fmt.Sprintf("pvcreate -qq --metadatasize=128M --dataalignment=%v '%v'", s.config.PVDataAlignment, device))
 	commands = append(commands, fmt.Sprintf("vgcreate -qq --autobackup=%v %v %v", conv.BoolToYN(s.BackupLVM), paths.VgIdToName(vgid), device))
 
 	// Execute command
--- a/executors/cmdexec/device.go
+++ b/executors/cmdexec/device.go
@@ -43,8 +43,18 @@ func (s *CmdExecutor) DeviceSetup(host, device, vgid string, destroy bool) (d *e
 		logger.Info("Data on device %v (host %v) will be destroyed", device, host)
 		commands = append(commands, fmt.Sprintf("wipefs --all %v", device))
 	}
-	commands = append(commands, fmt.Sprintf("pvcreate -qq --metadatasize=128M --dataalignment=%v '%v'", s.config.PVDataAlignment, device))
-	commands = append(commands, fmt.Sprintf("vgcreate -qq --autobackup=%v %v %v", conv.BoolToYN(s.BackupLVM), paths.VgIdToName(vgid), device))
+	commands = append(commands, fmt.Sprintf("pvcreate -qq --metadatasize=128M --dataalignment=%v '%v'", s.PVDataAlignment(), device))
+	commands = append(commands, fmt.Sprintf("vgcreate -qq --physicalextentsize=%v --autobackup=%v %v %v",
+
+		// Physical extent size
+		s.VGPhysicalExtentSize(),
+
+		// Autobackup
+		conv.BoolToYN(s.BackupLVM),
+
+		// Device
+		paths.VgIdToName(vgid), device),
+	)
 
 	// Execute command
 	err := rex.AnyError(s.RemoteExecutor.ExecCommands(host, commands, 5))
--- a/etc/heketi.json
+++ b/etc/heketi.json
@@ -52,6 +52,7 @@
       "user": "sshuser",
       "port": "Optional: ssh port.  Default is 22",
       "fstab": "Optional: Specify fstab file on node.  Default is /etc/fstab",
+      "pv_data_alignment": "Optional: Specify PV data alignment size. Default is 256K",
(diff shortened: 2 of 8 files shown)


executors/cmdexec/device.go
```diff
@@ -37,12 +37,18 @@ func (s *CmdExecutor) DeviceSetup(host, device, vgid string, destroy bool) (d *e
...
 
 	if destroy {
 		logger.Info("Data on device %v (host %v) will be destroyed", device, host)
 		commands = append(commands, fmt.Sprintf("wipefs --all %v", device))
 	}
-	commands = append(commands, fmt.Sprintf("pvcreate -qq --metadatasize=128M --dataalignment=256K '%v'", device))
+
+	if s.config.PVDataAlignment != "" {
+		d_alignment = s.config.PVDataAlignment
+	}
```
phlogistonjohn: This is a great start but I would like to see the defaulting of the value handled like the recent gluster cli timeout is handled, with a helper function that handles returning the configured value or a default. Thanks

phlogistonjohn: This can be a private function in this case though.

rsevilla87: @phlogistonjohn implemented in a similar way, PTAL.

executors/cmdexec/device.go
```diff
@@ -42,7 +42,8 @@ func (s *CmdExecutor) DeviceSetup(host, device, vgid string, destroy bool) (d *e
 		logger.Info("Data on device %v (host %v) will be destroyed", device, host)
 		commands = append(commands, fmt.Sprintf("wipefs --all %v", device))
 	}
-	commands = append(commands, fmt.Sprintf("pvcreate -qq --metadatasize=128M --dataalignment=256K '%v'", device))
+
```
phlogistonjohn: Was this extra line intentional?

phlogistonjohn:
There still appears to be some missing updates to make the interfaces in the tests work. Please update that so that the tests pass under travis. Once we do that I'll kick off a centos-ci functional tests run.

executors/cmdexec/device.go
rsevilla87: nope, solved ;)

phlogistonjohn:
ok to test

rsevilla87:
@phlogistonjohn PTAL Why are CI tests breaking?

phlogistonjohn:
Flaky test. I've restarted it.

phlogistonjohn:
I owe you a proper review, but since you widened the scope of the changes I have not yet gotten the chance to dive deeply into this.
One thing that jumps out at me is the change to the "default" command lines for xfs formatting. I'm a bit concerned about these new params that were not part of the command before. Care to explain why the the new defaults are what we want?

rsevilla87:
Thanks for your comments @phlogistonjohn 

su and sw options in xfs formatting indicate strip unit size and strip width respectively. These units should be configurable on RAID based deployments to ensure filesystem I/O alignment, though taking a look to the docs I'm changing this behavior to only use su and sw in case they were indicated in the config file.

Documentation states that
_For RAID 10 and JBOD, the -d su=<>,sw=<> option can be omitted. By default, XFS will use the thin-p chunk size and other parameters to make layout decisions._

phlogistonjohn:
ok to test

phlogistonjohn:
retest this please

rsevilla87:
retest this please

rsevilla87:
@phlogistonjohn tests still breaking :(

phlogistonjohn:
Hmmm, yes, centos-ci has been rather flaky lately.
In this case the failing test case TestCreateClusterTLSCert seems unrelated to your changes.
When run the TestEnabledTLS suite of tests locally with this PR checked out it passed.

phlogistonjohn:
We'll want at least one more review so @obnoxxx or @raghavendra-talur PTAL when you have the opportunity.

raghavendra-talur:
@rsevilla87 @phlogistonjohn @obnoxxx The code changes look excellent. I have no changes to suggest.

However, I am not totally sure if these options(chunk size, block size, alignment size and xfs options) belong to executor. They are attributes of devices(and device in particular) than of executor. I understand that there is no place in heketi config as of today which can take these attributes and hence this is the current choice.

I would be happy to merge this PR as-is if we could create a Issue highlighting the need that someone would want to add a SSD with no raid and different chunk size than in the config and there must be a way to do it and how such refactoring can be done later.

raghavendra-talur:
Ok, having looked through existing code I see one way to provide different settings for a particular device. The config changes in this PR can serve as default and in cases where the admin wants to add a device with different options then they can provide the same in device add request.

Device add request is part of the API and will have to be extended with these options. As I don't see any option where "0" holds a meaning I don't think it will have any problem with backward compatibility.
example commit: https://github.com/heketi/heketi/commit/1be4c44cfbb86cfbc1a3d5c74be049d187fbb8cd

obnoxxx:
Thanks for your review, @raghavendra-talur. The ability to configure these settings per device instead of globally is indeed an excellent point!

rsevilla87:
> Ok, having looked through existing code I see one way to provide different settings for a particular device. The config changes in this PR can serve as default and in cases where the admin wants to add a device with different options then they can provide the same in device add request.
> 
> Device add request is part of the API and will have to be extended with these options. As I don't see any option where "0" holds a meaning I don't think it will have any problem with backward compatibility.
> example commit: [1be4c44](https://github.com/heketi/heketi/commit/1be4c44cfbb86cfbc1a3d5c74be049d187fbb8cd)

You're right, device entry should be extended in a future to make heketi able to provision bricks with different IO alignment, and also taking into account that Heketi is able to manage several glusterFS clusters with different disk configuration :)

phlogistonjohn:
+1 to eventually supporting different per-device settings. maybe we could consider the device tags for that, but perhaps that would need more validation than tags provide.
Regardless, I think this PR is good as is and we can continue to enhance in the future. We've got 2 reviews and passing tests so I'm merging it now.
Thanks!

PR #1465, opened 2018-12-19
docs: Add a blockvolume design doc

obnoxxx:

The design for the blockvolume feature of heketi was originally contained in the [end-to-end design](https://github.com/gluster/gluster-kubernetes/blob/master/docs/design/gluster-block-provisioning.md)
in [gluster-kubernetes](https://github.com/gluster/gluster-kubernetes). Now this is finally extracted here into a proper heketi specific design doc (after the fact).


humblec:
@obnoxxx not sure the thoughts behind pulling heketi specific parts now (or very very late) in this repo :), but as its an extract from original doc nothing else to review I feel .

PR #1466, opened 2018-12-19
Diagnostics: examine gluster

raghavendra-talur:
This PR introduces diagnostics ability to heketi. 

Primarily, heketi provides a single command to fetch all the necessary information from Gluster to compare its database with the Gluster backend.

Further, it also performs some comparison between the two data sets to report problems and suggest fixes.

### Does this PR fix issues?

Fixes #

### DIFF
--- a/apps/glusterfs/examiner.go
+++ b/apps/glusterfs/examiner.go
@@ -1,0 +1,153 @@
+//
+// Copyright (c) 2018 The heketi Authors
+//
+// This file is licensed to you under your choice of the GNU Lesser
+// General Public License, version 3 or any later version (LGPLv3 or
+// later), or the GNU General Public License, version 2 (GPLv2), in all
+// cases as published by the Free Software Foundation.
+//
+
+package glusterfs
+
+import (
+	"fmt"
+	"reflect"
+
+	"github.com/boltdb/bolt"
+	"github.com/heketi/heketi/executors"
+)
+
+type ExaminerMode string
+
+const (
+	OnDemandExaminer ExaminerMode = "ondemand"
+	OfflineExaminer  ExaminerMode = "offline"
+)
+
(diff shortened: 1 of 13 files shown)


apps/glusterfs/examiner.go
```diff
@@ -0,0 +1,153 @@
...
+//
+
+package glusterfs
+
+import (
+	"fmt"
+	"reflect"
+
+	"github.com/boltdb/bolt"
+	"github.com/heketi/heketi/executors"
```
phlogistonjohn: Minor style nit: I've been trying to put a blank line between "external" imports and "intneral" imports even if there's a small number. If the list grows it prevents things from becoming a jumble.

apps/glusterfs/examiner.go
```diff
@@ -0,0 +1,153 @@
...
+	for _, node := range cdata.NodesData {
+		var volList []string
+		if node.VolumeInfo == nil {
+			errorstrings = append(errorstrings, fmt.Sprintf("heketi volume list could not be compared with volume list of node %v due to missing info", node.NodeHeketiID))
+			continue
+		}
+		for _, volume := range node.VolumeInfo.Volumes.VolumeList {
+			volList = append(volList, volume.VolumeName)
+		}
+		if !reflect.DeepEqual(heketiVolList, volList) {
```
phlogistonjohn: If you are using lists here, do they need to be sorted at some point, or not, or did I miss it?

phlogistonjohn:
General approach seems pretty solid and a good foundation for building more comparisions in the future.

phlogistonjohn:
Travis falied one out of the three version on what appears to be a flaky test. Since I'm 100% sure this change didn't impact that test I am merging anyway.

PR #1472, opened 2018-12-22
Fix db corruption from rollback

obnoxxx:

The volume cleanup code (function expungeVolumeWithOp) used to re-read the volume entry from the DB instead of operating on the in-memory copy from the operation. Hence the in-memory copy of the volume-entry of the operation was still containig the list of brick-ids from the original attempt even after the completion of expungeVolumeWithOp().

In the case the cleanup was called from the rollback of a volume create operation, when the operation was retried, the next Build() process would just continue using the in-memory volume entry and add new bricks to the existing list instead of starting over from an empty list.  When the volume entry was saved to disk at the end of Build(), it would contain the new brick ids and the old brick ids. Those old brick ids did not point to existing brick entries, because those had been cleaned  up correctly in the rollback. Hence the database was rendered inconsistent.

This patch fixes this problem by having expungeVolumeWithOp() act on the volume entry from the operation instead of re-loading it from the database.

### DIFF
--- a/apps/glusterfs/brick_entry.go
+++ b/apps/glusterfs/brick_entry.go
@@ -386,6 +386,7 @@ func (b *BrickEntry) remove(tx *bolt.Tx, v *VolumeEntry) error {
 
 	// Delete brick from volume entry
 	v.BrickDelete(b.Info.Id)
+	err = v.Save(tx)
 	if err != nil {
 		logger.Err(err)
 		return err
--- a/apps/glusterfs/operations_manage.go
+++ b/apps/glusterfs/operations_manage.go
@@ -283,6 +283,7 @@ func retryOperation(o Operation,
 	}
 	if e := o.Rollback(executor); e != nil {
 		logger.LogError("%v Rollback error: %v", label, e)
+		markFailedIfSupported(o)
 	}
 	// if we exceeded our retries, pull the "real" error out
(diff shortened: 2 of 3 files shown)


obnoxxx:
retest this please.

Note, is 5 second waiting enough for cleanup after restart?
```
02:57:38 --- FAIL: TestCleanupAfterRestart (61.36s)
02:57:38     assert.go:44: 
02:57:38         
ASSERT:	func (github.com/heketi/heketi/tests/functional/TestErrorHandling/tests.TestCleanupAfterRestart) 0x1310038
02:57:38         
	File /root/go/src/github.com/heketi/heketi/tests/functional/TestErrorHandling/tests/background_cleanup_test.go:97
02:57:38         
	Info: [expected info.InFlight == 0, got: 1]
```

obnoxxx:
retest this please

obnoxxx:
all tests passed :-)

phlogistonjohn:
Please compare to my counter-proposal in PR #1473 I believe that change retains the good isolation between operation steps but resets the (already) leaky volume entry object explicitly where needed.

phlogistonjohn:
Regardless of which solution we take, we should include a test which is available in PR #1473

obnoxxx:
closing since #1473 has been merged which is the better fix

PR #1483, opened 2019-01-07
logging: grab output of 'lsof /path/to/brick' in case unmounting fails

nixpanic:

There are regular problems reported where Heketi fails to remove the bricks from a Gluster volume. This is most often caused by an open filedescriptor in one of the brick processes. Debugging this needs to be done in a timely fashion, as processes may get restarted and any trace
of the kept open file is removed.

By logging the output of 'lsof /path/to/brick' when unmounting fails, we should be able to find the filenames that the brick processes did not close when the Gluster volume got stopped and deleted.

### DIFF
--- a/docs/admin/server.md
+++ b/docs/admin/server.md
@@ -27,6 +27,7 @@ The config file has information required to run the Heketi server.  The config f
         * fstab: _string_, Fstab file where to store mount points
         * backup_lvm_metadata: _bool_, Create archives of the LVM metadata when running vgcreate/lvcreate
         * sudo: _bool_, set to true when SSHing as a non root user
+	* debug_umount_failures: _bool_, Enable to capture more details in case brick unmounting fails. Can be overridden by the HEKETI_DEBUG_UMOUNT_FAILURES environment variable.
     * kubexec: _map_, Kubernetes configuration
         * host: _string_, Kubernetes API host.  Example `https://myhost:8443`.  Can also be use using environment variable HEKETI_KUBE_APIHOST
         * cert: _string_, Certificate file to for HTTPS connection. Can also be use using environment variable HEKETI_KUBE_CERTFILE
@@ -36,6 +37,7 @@ The config file has information required to run the Heketi server.  The config f
         * namespace: _string_, Kubernetes namespace or OpenShift project where GlusterFS containers/Pods are running. Can also be use using environment variable HEKETI_KUBE_NAMESPACE.
         * fstab: _string_, Fstab file where to store mount points
         * backup_lvm_metadata: _bool_, Create archives of the LVM metadata when running vgcreate/lvcreate
+	* debug_umount_failures: _bool_, Enable to capture more details in case brick unmounting fails. Can be overridden by the HEKETI_DEBUG_UMOUNT_FAILURES environment variable.
 
 ## Advanced Options
 The following configuration options should only be set on advanced configurations under `glusterfs` section:
--- a/etc/heketi.json
+++ b/etc/heketi.json
@@ -58,7 +58,9 @@
       "xfs_sw": "Optional: Specify number of data disks in the underlying RAID device",
       "xfs_su": "Optional: Specifies a stripe unit or RAID chunk size.",
       "backup_lvm_metadata": false,
-      "gluster_cli_timeout": "Optional: Timeout, in seconds, passed to the gluster cli invocations"
+      "gluster_cli_timeout": "Optional: Timeout, in seconds, passed to the gluster cli invocations",
+      "_debug_umount_failures": "Optional: boolean to capture more details in case brick unmounting fails",
+      "debug_umount_failures": true
     },
 
     "_kubeexec_comment": "Kubernetes configuration",
@@ -76,7 +78,9 @@
       "xfs_sw": "Optional: Specify number of data disks in the underlying RAID device.",
       "xfs_su": "Optional: Specifies a stripe unit or RAID chunk size.",
       "backup_lvm_metadata": false,
-      "gluster_cli_timeout": "Optional: Timeout, in seconds, passed to the gluster cli invocations"
+      "gluster_cli_timeout": "Optional: Timeout, in seconds, passed to the gluster cli invocations",
+      "_debug_umount_failures": "Optional: boolean to capture more details in case brick unmounting fails",
+      "debug_umount_failures": true
     },
 
     "_db_comment": "Database file name",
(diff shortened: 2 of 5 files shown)


nixpanic:
On Mon, Jan 07, 2019 at 08:21:26AM -0800, John Mulligan wrote:
> phlogistonjohn commented on this pull request.
> 
> Maybe we should make this optional behind a config varable and/or only trigger on lsof if the output suggests it. IIRC lsof is not a "cheap" command to run.

I thought about that too. But then I decided against it, as this should
only happen in the rare case where unmounting fails. Running lsof can be
expensive indeed, however I think it weighs well against the improved
debugability.

Also, the timeout is set to 5 seconds, so even if there is a delay, it
would not be very noticeable.

phlogistonjohn:
> On Mon, Jan 07, 2019 at 08:21:26AM -0800, John Mulligan wrote: phlogistonjohn commented on this pull request. Maybe we should make this optional behind a config varable and/or only trigger on lsof if the output suggests it. IIRC lsof is not a "cheap" command to run.
> I thought about that too. But then I decided against it, as this should only happen in the rare case where unmounting fails. Running lsof can be expensive indeed, however I think it weighs well against the improved debugability. Also, the timeout is set to 5 seconds, so even if there is a delay, it would not be very noticeable.

That would be true if the timeout worked consistently, but it doesn't do anything on kubeexec today.
I will however buy the argument that unmounting errors ought to be rare. I'm still trying to weigh the tradeoffs.

raghavendra-talur:
> > On Mon, Jan 07, 2019 at 08:21:26AM -0800, John Mulligan wrote: phlogistonjohn commented on this pull request. Maybe we should make this optional behind a config varable and/or only trigger on lsof if the output suggests it. IIRC lsof is not a "cheap" command to run.
> > I thought about that too. But then I decided against it, as this should only happen in the rare case where unmounting fails. Running lsof can be expensive indeed, however I think it weighs well against the improved debugability. Also, the timeout is set to 5 seconds, so even if there is a delay, it would not be very noticeable.
> 
> That would be true if the timeout worked consistently, but it doesn't do anything on kubeexec today.
> I will however buy the argument that unmounting errors ought to be rare. I'm still trying to weigh the tradeoffs.

I agree that we need to have the lsof behind a switch. Also, just getting lsof for the brick mount path might not be sufficient. How do we get the list of files under the mount point that are open?

nixpanic:
On Tue, Jan 08, 2019 at 05:02:42AM -0800, Raghavendra Talur wrote:
> I agree that we need to have the lsof behind a switch. Also, just getting lsof for the brick mount path might not be sufficient. How do we get the list of files under the mount point that are open?

The command ddies it like this. It recursively lists all open files.

nixpanic:
What kind of switch would you like to have? Something in the heketi configuration file, an environment variable, or something else?

obnoxxx:
@nixpanic Any update? I agree to hide it behind a switch. Do you need help to do the switch implementation?

nixpanic:
@phlogistonjohn @obnoxxx how about logging the `lsof` output for the brick when `loglevel=debug`? Just adding an other switch can be done, but I think that is less user-friendly.

nixpanic:
This is an updated version that only logs `lsof` when the loglevel is set to `debug`.

phlogistonjohn:
> This is an updated version that only logs `lsof` when the loglevel is set to `debug`.

It's not a bad idea in concept, but I bet you didn't test this out on a standard deploy of heketi because I think you would have noticed that it would always capture the output because the logger for the executor package is *always* set to debug and thus you'd always be getting the lsof results.
You're welcome to take a swing at cleaning up the logging in the executor package but I honestly think it would be less work to add a separate config parameter.

nixpanic:
@phlogistonjohn I can add a change that passes the loglevel from the app to the executor. Is that what you mean?

phlogistonjohn:
> @phlogistonjohn I can add a change that passes the loglevel from the app to the executor. Is that what you mean?

No I just don't like the idea of having the behavior piggyback off of the log level. Could you please make an explicit config parameter?

nixpanic:
On Fri, Jan 18, 2019 at 03:38:43PM +0000, John Mulligan wrote:
> > @phlogistonjohn I can add a change that passes the loglevel from the app to the executor. Is that what you mean?
> 
> No I just don't like the idea of having the behavior piggyback off of the log level. Could you please make an explicit config parameter?

Done! Enabled it by default in the `etc/heketi.json` for now.

nixpanic:
On Mon, Jan 21, 2019 at 03:37:00AM -0800, Michael Adam wrote:
> obnoxxx requested changes on this pull request.
> 
> If I am not mistaken, this is the json config variable, but we don't have the corresponding env var yet. (`HEKETI_DEBUG_UMOUNT_FAILURES` ...).

Yes, it is in the heketi.json.

> Please add that as well.

sure, I can do that

> If I am not mistaken it is disabled by default and has to be enabled through the config setting. Absence of the setting means false.

Disabling it by default is not very useful... Most users run with the
default configuration and will not change the option. As the failure is
rare, and we *really* need the output to debug the problems, I prefer to
have it enabled by default. It can be disabled in case it causes any
harm in certain environments.

obnoxxx:
> >  If I am not mistaken it is disabled by default and has to be enabled through the config setting. Absence of the setting means false.
> 
> Disabling it by default is not very useful... Most users run with the
default configuration and will not change the option. As the failure is
rare, and we *really* need the output to debug the problems, I prefer to
have it enabled by default. It can be disabled in case it causes any
harm in certain environments.

That was my point: It is *disabled* in the code. You have to enable it with the config setting. But the heketi.json file ist just an example. It does not reflect heketi's default values really. Distributors can and should change this.

If you want to have something that's enabled by default, then you should create a config setting that can *disable* this feature. See for instance the setting `HEKETI_DISABLE_HEALTH_MONITOR`, introduced by commit ce2393e1c983acfa0df3bd5ef993d16054bdb799.

phlogistonjohn:
> LGTM, thanks!
> 
> @phlogistonjohn did you imagine a config setting to turn it _off_ or to turn it _on_? ...

I was assuming an option to turn it on, so that we don't incur the costs of that command on healthy clusters. However, I can see the point of wanting it on by default. Especially if we believe this is happening commonly enough. But truthfully, I'm not so wedded to a particular default behavior as I was to the idea of making it configurable, which it now is.

obnoxxx:
retest this please

PR #1486, opened 2019-01-09
Ensure pending volumes filter works on delete operations

phlogistonjohn:

The functions used to filter out pending volumes and block volumes were not correctly checking for all cases where a pending operation was indicating that a volume or block volume was pending & incomplete.

### Does this PR fix issues?

Fixes [rhbz#1662141](https://bugzilla.redhat.com/show_bug.cgi?id=1662141)

### Notes for the reviewer

It is now possible for a volume to go pending for delete and then be unmarked pending if the delete fails. This implies that it is possible for a volume to vanish from the listing and then reappear some time later. This is a bit odd but I don't think this behavior is worse than failing the listing commands because a pending delete operation is in the db.

### DIFF
--- a/apps/glusterfs/listing.go
+++ b/apps/glusterfs/listing.go
@@ -90,8 +90,12 @@ func UpdateClusterInfoComplete(tx *bolt.Tx, ci *api.ClusterInfoResponse) error {
 // an error if the db cannot be read.
 func MapPendingVolumes(tx *bolt.Tx) (map[string]string, error) {
 	return mapPendingItems(tx, func(op *PendingOperationEntry, a PendingOperationAction) bool {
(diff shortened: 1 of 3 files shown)


phlogistonjohn:
retest this please

phlogistonjohn:
centos-ci failure appears due to flaky test. Will start another run, but please don't wait for that to succeed for reviews thanks.

phlogistonjohn:
retest this please

phlogistonjohn:
retest this please

obnoxxx:
the centos-ci failure seems to be entirely unrelated

obnoxxx:
I think I have spotted a possible way of reducing the flakyness of this test. will check and possibly file a separate PR for improving that.

PR #1487, opened 2019-01-09
Minor correction in error message

vbellur:
Signed-off-by: Vijay Bellur <vbellur@redhat.com>

Fixes a minor error in an error message!


obnoxxx:
ok to test

PR #1491, opened 2019-01-14
Tweak travis config to use newer golang versions, don't do extra work on older versions

phlogistonjohn:

This change bumps the versions of golang we're using in travis, dropping 1.8 and adding 1.11.
It also stops running "make release" on older versions and does that as a separate task.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,13 +17,14 @@ git:
   depth: 9999999
 matrix:
   include:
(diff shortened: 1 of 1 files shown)


obnoxxx:
yay, success!

obnoxxx:
Ignoring centos-ci test results since this will just affect travis.
Merging.

PR #1492, opened 2019-01-15
tests: try to select best python version for use in travis

phlogistonjohn:

Travis runs have been breaking lately and it appears to be due
to a change in the version of python (3.6 -> 3.7). This change
tries to make it simple to enable and use at least one version
of python 3.


phlogistonjohn:
@obnoxxx @raghavendra-talur  Please expedite reviews of this PR so we can unblock other PRs that are failing because of the changes in travis ci.

PR #1500, opened 2019-01-17
extras: introduce delete_unowned_bricks.sh script for brick cleanup

raghavendra-talur:
This script is useful for cleaning of bricks when they are not in use
by any Gluster volume and heketi does not have any reference to them.
This is possible when the heketi database has been scrubbed of brick
entries to make the db consistent.

The script must be run on each node/pod of Gluster as it cleans up only
bricks local to the node. The script is run in two phases. In the first
phase(scan), a list is made of unused LVs on the node and files
containing the list are created per VG. In the second phase(delete),
the script is invoked with those files as input.

ip-prefix argument is required to determine bricks local to the node.

fstab-path argument is required and is usually "/var/lib/heketi/fstab"
in a pod and "/etc/fstab" if Gluster is not containerized.

NOTE: make sure heketi is not running while the script is run.

Before running the script in delete mode, check the files that it generated have valid IDs.

Pair-Programmed-With: John Mulligan <jmulligan@redhat.com>


phlogistonjohn:
FWIW we ought to be able to ignore centos-ci status for this PR. It can have no impact on the ci tests.

extras/tools/delete_unowned_bricks.sh
```diff
@@ -0,0 +1,288 @@
...
+
+    if [[ -z "$fstab_path" ]]; then
+        echo "error: fstab path is required" >&2
+        exit 2
+    fi
+
+    echo ""
+    echo "Checking for glusterd on local and peer nodes"
+    checkLocalNodeHasGlusterdRunning
+    #checkPeerNodeStatus
```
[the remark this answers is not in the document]
raghavendra-talur: This function is not used because it is not guaranteed that we would have ssh access to other gluster nodes. The script acts only locally and has to be run on all the nodes one after other. Will delete references to peer actions everywhere.

extras/tools/delete_unowned_bricks.sh
```diff
@@ -0,0 +1,266 @@
...
+    echo '-h, --help        show this help text'
+    echo ''
+}
+
+
+function parse_args () {
+    args=$(getopt \
+              --options h \
+              --long dryrun,lvlist:,vgid:,prompt,ip-prefix:,fstab-path:,help \
+              -n 'delete_unowned_bricks.sh' \
```
raghavendra-talur: I am not sure if we should have basename $0 here or explicitly name it.

extras/tools/delete_unowned_bricks.sh
```diff
@@ -0,0 +1,288 @@
...
+
+    for eachVG in $allvgs
+    do
+        lvs | grep brick_  | grep "$eachVG" |  awk '{print $1}' | awk  -F '_' '{print $2}' | sort -u > "All_in_${eachVG}.txt"
+    done
+
+    for eachVG in $allvgs
+    do
+        sdiff "gluster_bricks_in_${eachVG}.txt" "All_in_${eachVG}.txt" | grep \> | awk '{print $2}' > "to_delete_${eachVG}.txt"
+        echo "./delete_unowned_bricks.sh delete --ip-prefix=${ip_prefix} --fstab-path=${fstab_path} --vgid=${eachVG} --lvlist=to_delete_${eachVG}.txt --prompt"
```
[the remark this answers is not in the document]
raghavendra-talur: done

extras/tools/delete_unowned_bricks.sh
```diff
@@ -0,0 +1,288 @@
...
+}
+
+function checkPeerNodeStatus() {
+    peers=$(gluster pool list | tail -n +2 | awk '{print $2}')
+    if [ $? -ne 0 ]
+    then
+        echo "Could not fetch pool list"
+        exit 2
+    fi
+    for eachPeer in ${peers}
```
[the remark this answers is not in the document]
raghavendra-talur: done

extras/tools/delete_unowned_bricks.sh
```diff
@@ -0,0 +1,288 @@
...
+        exit 2
+    fi
+    if ! systemctl --quiet is-active glusterd
+    then
+        echo "Glusterd is not active"
+        exit 2
+    fi
+}
+
+function checkPeerNodeStatus() {
```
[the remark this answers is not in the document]
raghavendra-talur: removed.

raghavendra-talur:
> This looks great. (I can't deny that I was slightly puzzled for this not to be a python script... :-)
> 
> A few minor comments/requests inline.
> 
> And: The instructions from the commit message should probably go into a / the readme.
> 

Would it be better to have it as part of the "usage" display?

phlogistonjohn:
> > This looks great. (I can't deny that I was slightly puzzled for this not to be a python script... :-)
> > A few minor comments/requests inline.
> > And: The instructions from the commit message should probably go into a / the readme.
> 
> Would it be better to have it as part of the "usage" display?

+1 to that, perhaps reworded slightly to be more like an example

extras/tools/delete_unowned_bricks.sh
obnoxxx: also $0, afaiac, I just didn't spot this in my first review.

obnoxxx:
> > > And: The instructions from the commit message should probably go into a / the readme.
> > 
> > Would it be better to have it as part of the "usage" display?
> 
> +1 to that, perhaps reworded slightly to be more like an example

That would work for me.

extras/tools/delete_unowned_bricks.sh
obnoxxx: use `$0`

extras/tools/delete_unowned_bricks.sh
```diff
@@ -0,0 +1,266 @@
...
+        read -r -p 'Continue? [y/N]: ' pval
+        if ! echo "$pval" | grep -q -i '^y'; then
+            return 0
+        fi
+    fi
+    "$@"
+}
+
+function checkLocalNodeHasGlusterdRunning() {
+    # Check if local node has gluster binary and if glusterd is running
```
obnoxxx: Just a cosmetic thing: wouldn't it be better to have the comment block *before* the function?

raghavendra-talur: removed the comment. Not useful anymore.

obnoxxx:
Shellcheck error on the script:

```
-- Running: 020-test-shellscripts.sh --
In ./extras/tools/delete_unowned_bricks.sh line 11:
script_name=$(basename $0)
                       ^-- SC2086: Double quote to prevent globbing and word splitting.
In ./extras/tools/delete_unowned_bricks.sh line 59:
              -n ${script_name} \
                 ^------------^ SC2086: Double quote to prevent globbing and word splitting.
For more information:
  https://www.shellcheck.net/wiki/SC2086 -- Double quote to prevent globbing ...
detected issues in ./extras/tools/delete_unowned_bricks.sh
failed ./tests/020-test-shellscripts.sh [1]
```

Issue #1478, opened 2019-01-03
Running device delete returns an error that a device is not in failed state, but deletes the device anyway


phlogistonjohn:
### Kind of issue

Bug

### Observed behavior

Running the "device delete" subcommand on an empty, but non-disabled device first prints an error message indicating that the device is "not in failed state" which implies the device can't be deleted at this time. However, on the back end the command has already run the vgremove command and if the command is run a 2nd time it errors out running the vgremove command again.

Combined server logs and commands below:
```
$ ./heketi-cli device delete 6108d71d8b0bea
8f1f27ae8d53f69d87
[heketi] INFO 2019/01/03 13:48:41 Deleting device 6108d71d8b0bea8f1f27ae8d53f69d87 on
 node 88891699f3de4a84eb8bf696e648c15f
[negroni] 2019-01-03T13:48:41-05:00 | 202 |      242.051µs | localhost:8080 | DELETE 
/devices/6108d71d8b0bea8f1f27ae8d53f69d87   
[asynchttp] INFO 2019/01/03 13:48:41 asynchttp.go:288: Started job fd87e308cee065fa7d
61d0323afcc2a3
[negroni] 2019-01-03T13:48:41-05:00 | 200 |      12.383µs | localhost:8080 | GET /que
ue/fd87e308cee065fa7d61d0323afcc2a3
[cmdexec] DEBUG 2019/01/03 13:48:42 heketi/pkg/remoteexec/ssh/ssh.go:187:ssh.(*SshExec).ExecCommands: Ran command [/bin/bash -c 'sudo vgremove -qq vg_6108d71d8b0bea8f1f27ae8d53f69d87'] on 192.168.10.100:22: Stdout []: Stderr []
[cmdexec] DEBUG 2019/01/03 13:48:42 heketi/pkg/remoteexec/ssh/ssh.go:187:ssh.(*SshExec).ExecCommands: Ran command [/bin/bash -c 'sudo pvremove -qq '/dev/vdd''] on 192.168.10.100:22: Stdout []: Stderr []
[...]

PR #1503, opened 2019-01-21
check device state before deleting the device

Madhu-1:
This PR moves checks in DeviceEntry.Delete to a new function called CheckDelete and calls the  new function from both App.DeviceDelete and DeviceEntry.Delete 
### Does this PR fix issues?

Fixes: #1478

### DIFF
--- a/apps/glusterfs/app_device.go
+++ b/apps/glusterfs/app_device.go
@@ -220,18 +220,21 @@ func (a *App) DeviceDelete(w http.ResponseWriter, r *http.Request) {
 			return logger.Err(err)
 		}
 
-		// Check if we can delete the device
-		if device.HasBricks() {
-			http.Error(w, device.ConflictString(), http.StatusConflict)
-			logger.LogError(device.ConflictString())
-			return ErrConflict
-		}
-
 		// Access node entry
 		node, err = NewNodeEntryFromId(tx, device.NodeId)
 		if err != nil {
 			http.Error(w, err.Error(), http.StatusInternalServerError)
 			return logger.Err(err)
+		}
+
+		// Check if we can delete the device
+		if err := device.CheckDelete(); err != nil {
+			if err == ErrConflict {
+				http.Error(w, device.ConflictString(), http.StatusConflict)
+			} else {
+				http.Error(w, err.Error(), http.StatusBadRequest)
+			}
+			return err
 		}
 
 		return nil
--- a/apps/glusterfs/device_entry.go
+++ b/apps/glusterfs/device_entry.go
@@ -161,20 +161,27 @@ func (d *DeviceEntry) ConflictString() string {
 	return fmt.Sprintf("Unable to delete device [%v] because it contains bricks", d.Info.Id)
 }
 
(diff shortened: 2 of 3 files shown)


apps/glusterfs/app_device.go
```diff
@@ -220,20 +220,23 @@ func (a *App) DeviceDelete(w http.ResponseWriter, r *http.Request) {
...
 		// Access node entry
 		node, err = NewNodeEntryFromId(tx, device.NodeId)
 		if err != nil {
 			http.Error(w, err.Error(), http.StatusInternalServerError)
 			return logger.Err(err)
 		}
 
+		// Check if we can delete the device
+		if err := device.CheckDelete(tx); err != nil {
+			if err.Error() != ErrConflict.Error() {
```
phlogistonjohn: I'm not sure why you are comparing the error strings. Could you elaborate if there's a specific reason for this?

obnoxxx:
ok to test

apps/glusterfs/app_device.go
Madhu-1: @phlogistonjohn am comparing the error to return the HTTP status code.
if the device state is not in failed state am considering it as a bad request

phlogistonjohn: I mean why is the line `if err.Error() != ErrConflict.Error() {` and not `if err != ErrConflict` ? ErrConflict is a error constant after all.

Madhu-1: yeah i could have just compared err instead of comparing string. updated PTAL

apps/glusterfs/app_device.go
```diff
@@ -220,20 +220,23 @@ func (a *App) DeviceDelete(w http.ResponseWriter, r *http.Request) {
...
 		// Access node entry
 		node, err = NewNodeEntryFromId(tx, device.NodeId)
 		if err != nil {
 			http.Error(w, err.Error(), http.StatusInternalServerError)
 			return logger.Err(err)
 		}
 
+		// Check if we can delete the device
+		if err := device.CheckDelete(tx); err != nil {
+			if err != ErrConflict {
```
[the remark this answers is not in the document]
Madhu-1: done

apps/glusterfs/device_entry.go
```diff
@@ -161,14 +161,11 @@ func (d *DeviceEntry) ConflictString() string {
 	return fmt.Sprintf("Unable to delete device [%v] because it contains bricks", d.Info.Id)
 }
 
-func (d *DeviceEntry) Delete(tx *bolt.Tx) error {
-	godbc.Require(tx != nil)
-
+func (d *DeviceEntry) CheckDelete(tx *bolt.Tx) error {
```
phlogistonjohn: If CheckDelete doesn't use tx it should not take it as an argument.

Madhu-1: missed that one :( fixed now, PTAL.

obnoxxx:
for the record: all tests passed.

PR #1508, opened 2019-01-21
handle block volume "does not exist" errors if BHV never was created

phlogistonjohn:

In the case that we request a delete of a block volume but the
    block hosting volume never existed the error response from gluster
    block is a little different than the usual. In this case we
    get "Volume X does not exist" which heketi was not treating as
    a does-not-exist type error when processing gluster-block delete
    response.

This includes a fix and two tests for gluster-block volume rollback behavior (using the error handling suite).

### DIFF
--- a/executors/cmdexec/block_volume.go
+++ b/executors/cmdexec/block_volume.go
@@ -126,8 +126,12 @@ func (s *CmdExecutor) BlockVolumeDestroy(host string, blockHostingVolumeName str
 	}
 
 	if blockVolumeDelete.Result == "FAIL" {
-		if strings.Contains(blockVolumeDelete.ErrMsg, "doesn't exist") &&
-			strings.Contains(blockVolumeDelete.ErrMsg, blockVolumeName) {
+		errHas := func(s string) bool {
+			return strings.Contains(blockVolumeDelete.ErrMsg, s)
+		}
+
+		if (errHas("doesn't exist") && errHas(blockVolumeName)) ||
+			(errHas("does not exist") && errHas(blockHostingVolumeName)) {
 			return &executors.VolumeDoesNotExistErr{Name: blockVolumeName}
 		}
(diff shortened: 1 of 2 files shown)


obnoxxx:
For the record: all tests passed.
It seems like the centos-ci tests runs are a lot more reliable again. :-)

executors/cmdexec/block_volume.go
```diff
@@ -126,8 +126,12 @@ func (s *CmdExecutor) BlockVolumeDestroy(host string, blockHostingVolumeName str
...
 
 	if blockVolumeDelete.Result == "FAIL" {
-		if strings.Contains(blockVolumeDelete.ErrMsg, "doesn't exist") &&
-			strings.Contains(blockVolumeDelete.ErrMsg, blockVolumeName) {
+		errHas := func(s string) bool {
+			return strings.Contains(blockVolumeDelete.ErrMsg, s)
+		}
+
+		if (errHas("doesn't exist") && errHas(blockVolumeName)) ||
+			(errHas("does not exist") && errHas(blockHostingVolumeName)) {
```
obnoxxx: Just a nit: This could have been put like this:

```golang
if (errHas("doesn't exist") || errHas("does not exist")) &&
   errHas(blockHostingVolumeName) {
...
```
which I would find a little more natural. At least shorter. ;-)

phlogistonjohn: Not really. The first pair is the text "doesn't exist" and the block volume name and the 2nd pair is the text "does not exist" and the BHV name. I checked the gluster-block source code just to be sure but it never prints the block volume name in that error msg, only the gluster volume (BHV) name.
You could try to convince me that I'm being overly cautios about checking that the error messages refer to the volumes, but I like that extra layer of defense at the moment.

obnoxxx: Dang... I misread that. Must have been too late. :-)))

PR #1515, opened 2019-02-04
Upgrade negroni version to 1.0.0 and update glide.lock

### LINKED ISSUE #694: No timestamps in logs
It's harder to debug this way. Could you include some kind of timestamps in the logs?

ansiwen:

Negroni is still locked to version 0.2.0 which doesn't include timestamps in logs. This PR upgrades negroni to version 1.0.0.

### Does this PR fix issues?

Fixes #694

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -1,5 +1,5 @@
-hash: 5fc263bfacc703e0a599166ccae968891dd53d7b4e42a3d8356b3194e45833bb
-updated: 2018-06-01T15:20:52.053211272+02:00
+hash: a7e0c0723a91776f34ff8a742912c032e115701284efe18eed15e96da0ed00fd
+updated: 2019-02-07T15:18:42.253317799+01:00
 imports:
 - name: bitbucket.org/ww/goautoneg
   version: 75cd24fc2f2c2a2088577d12123ddee5f54e0675
@@ -137,7 +137,7 @@ imports:
   subpackages:
   - codec
 - name: github.com/urfave/negroni
-  version: fde5e16d32adc7ad637e9cd9ad21d4ebc6192535
+  version: c6a59be0ce122566695fbd5e48a77f8f10c8a63a
 - name: golang.org/x/crypto
   version: d172538b2cfce0c13cee31e647d0367aa8cd2486
   subpackages:
--- a/glide.yaml
+++ b/glide.yaml
@@ -15,7 +15,7 @@ import:
 - package: github.com/spf13/pflag
   version: e57e3eeb33f795204c1ca35f56c44f83227c6e66
 - package: github.com/urfave/negroni
-  version: ^0.2.0
+  version: ^1.0.0
 - package: golang.org/x/crypto
   subpackages:
   - ssh


phlogistonjohn:
So this change is a perfect example of my frustration with glide in that when I look at the changes to glide.lock. The glide.yaml change looks 100% reasonable but I don't see why all these other packages get touched in glide.lock

I know @obnoxxx and I have had conversations about this in the past bu AFAIR we have never come to a complete conclusion what our policy should be around updating the glide.lock. 
I guess it is time to have another one of those conversations. :-)

ansiwen:
@phlogistonjohn I can fully understand. I simply don't know of (and after my research I believe there isn't) another way to update glide.lock only for negroni _and_ having a correct hash. So I simply gave up and take it as it is.

ansiwen:
And by the way: glide die die die!

ansiwen:
ok to test

raghavendra-talur:
@centos-ci add to whitelist

raghavendra-talur:
@centos-ci ok to test

phlogistonjohn:
One of the new tests in the test suite seems to fail all the time in the openshift image. Since we control the image contents (via extras/openshift-ci/build-image/Dockerfile)  we should be able to fix that.... but I haven't had a change to look into it yet. For now let's just ignore openshift-ci

phlogistonjohn:
retest this please

obnoxxx:
again a TLS error in centos-ci:
```
11:31:02 === RUN   TestCreateClusterTLSCert
11:31:02 Heketi v8.0.0-347-gb769069-pr_1515
11:31:02 [heketi] INFO 2019/02/06 11:31:02 Loaded mock executor
11:31:03 --- FAIL: TestCreateClusterTLSCert (0.40s)
11:31:03     assert.go:44: 
11:31:03         
ASSERT:	func (github.com/heketi/heketi/tests/functional/TestEnabledTLS/client_tls_test_test.testClientActions) 0x699feb
11:31:03         
	File /root/go/src/github.com/heketi/heketi/tests/functional/TestEnabledTLS/client_tls_test/client_tls_test.go:95
11:31:03         
	Info: [expected err == nil, got: Post https://localhost:8080/clusters: dial tcp [::1]:8080: connect: connection refused]
```
does not seem related

obnoxxx:
retest this please

glide.lock
```diff
@@ -137,7 +137,7 @@ imports:
   subpackages:
   - codec
 - name: github.com/urfave/negroni
-  version: fde5e16d32adc7ad637e9cd9ad21d4ebc6192535
+  version: c6a59be0ce122566695fbd5e48a77f8f10c8a63a
```
obnoxxx: In principle it should be possible to only change this hunk of `glide.lock`. We've done it before. E.g. in  patch 01f515ac688013097efab91971f3219483580189. The problem is that in this case, when you change the `glide.yaml` file, the hash in the `glide.lock` will be incorrect (the hash is actually that of the `glide.yaml` file. If you do the change of this negroni version without the hash update at the top of the lock file, you will get a warning:
```
$ make                                                                    
echo "Installing vendor directory"                                                                                                                 
Installing vendor directory                                                                                                                        
if [ "" ]; then mkdir -p ""; fi                                                
glide install -v                                                                                                                                  
[WARN]  Lock file may be out of date. Hash check of YAML failed. You may need to run 'update'  
[INFO]  Downloading dependencies. Please wait...
...
```
but it works nicely otherwise.
You can silence the warning by putting the up-to-date hash of the yaml file into the lock file, but that will be potentially incorrect. Then again, the lock file is not invariant for the hash of the yaml. It would change as dependent projects evolve. So I suggest adapting the hash in the file.

I suggest that we:
1. do the change with only this hunk and the hash hunk at the top of this file as the `glide.lock` part of this PR
2. Commit the complete update of the glide.lock file as part of a later PR (or separate patch in this PR)
3. Get rid of glide :-)

glide.lock
```diff
@@ -1,28 +1,28 @@
-hash: 5fc263bfacc703e0a599166ccae968891dd53d7b4e42a3d8356b3194e45833bb
-updated: 2018-06-01T15:20:52.053211272+02:00
+hash: a7e0c0723a91776f34ff8a742912c032e115701284efe18eed15e96da0ed00fd
```
obnoxxx: This is the hash of the `glide.yaml` file.

ansiwen: I wasn't aware that the hash only depends on `glide.yaml`. Then it's easy of course. I pushed a new version, that only changes the hash and the negroni version in `glide.lock`.

PR #1517, opened 2019-02-05
tests: temporarily disable goimports

phlogistonjohn:

It appears that in quite a few of our (totally valid) files the
goimports tool gets confused and removes valid imports.
The problem is similar to that mentioned in
  https://github.com/golang/go/issues/29557
but I am not entirely sure that's the exact issue.
So until all our files are goimports safe or goimports behavior
is improved I'm disabling the check script.

FWIW this script was never run in travis ci as it appears goimports
tool was not added to our travis instances and was not
immediately caught when the script was added.

### Notes for the reviewer

Should stop the openshift-ci tests from breaking all the time.

### DIFF
--- a/tests/030-goimports.sh
+++ b/tests/030-goimports.sh
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+echo "warning: goimports check temporarily disabled" >&2
+exit 0
+
 GOIMPORTSCHECK="$(command -v goimports 2>/dev/null)"
 
 if [[ -z "${GOIMPORTSCHECK}" ]]; then


obnoxxx:
I don't really understand the problem that we are trying to avoid here.
I am running the goimports test successfully on my machine(s).

obnoxxx:
> > I don't really understand the problem that we are trying to avoid here.
> > I am running the goimports test successfully on my machine(s).

> Interesting. It behaves the same was locally for me as it does in the container the tests are run in in openshift (it always fails on some files). I don't know how consistent this behavior is given my previous comment that goimports was not running in travis-ci and so I'm unaware if it would have succeeded or failed there.

how do you see what's going wrong in the openshift-ci? I only see that the 030-goimports.sh fails, nothing else. i'm likely looking at the wrong stuff...

obnoxxx:
it's really weird. i don't know what's different. I do consistently have this test passing on my machine...
```bash
$ COUNT=1; while [[ $COUNT -lt 20 ]] ; do echo -n "$COUNT: "; ./tests/030-goimports.sh || exit 1 ; echo OK; COUNT=$((COUNT+1)) ; done
1: OK
2: OK
3: OK
4: OK
5: OK
6: OK
7: OK
8: OK
9: OK
10: OK
11: OK
12: OK
13: OK
14: OK
15: OK
16: OK
17: OK
18: OK
19: OK
$ 
```
also for higher counts. just ran 50.

The flakiness might depend on the version of goimports installed.
I don't know if hardcoding the skipping in the script itself is a good way to approach it.
Can we solve the skipping differently?

phlogistonjohn:
Sure, but this was meant to be a short term fix to resolve the constantly failing openshift-ci runs. I would not have done it this way if the tests were running in travis but since I am certain that the test is being skipped in travis I went with this quick-and-dirty approach.

phlogistonjohn:
Since openshift-ci is no longer running automatically on all PRs and marking them as failed CI this PR seems less relevant now. I suppose next time goimports gives us trouble we will have to figure out how to fix it the right way.

PR #1518, opened 2019-02-11
implement strict zone checking

obnoxxx:

Zones are currently a best-effort concept. While the expectation is that e.g. a replica-3 brick set will span 3 zones, this is not always the case. This PR introduces strict zone checking. The behavior can be controlled with the new config option `zone_checking` (and the corresponding environment variable `HEKETI_ZONE_CHECKING`) which can be set to `strict`, which only lets volume creates succeed if each brick set is spread across sufficiently many zones,  or to `none` (the default) which lets heketi behave like before.

### Notes for the reviewer

Filed this as a WIP PR to see what the CI thinks and to start the discussion. Will need to squash a few patches before merging if we agree about this approach.

### DIFF
--- a/apps/glusterfs/volume_entry_test.go
+++ b/apps/glusterfs/volume_entry_test.go
@@ -42,7 +42,17 @@ func setupSampleDbWithTopology(app *App,
 	clusters, nodes_per_cluster, devices_per_node int,
 	disksize uint64) error {
 
-	zones_per_cluster := nodes_per_cluster
+	return setupSampleDbWithTopologyWithZones(app,
+		clusters,
+		nodes_per_cluster,
+		nodes_per_cluster,
+		devices_per_node,
+		disksize)
+}
+
+func setupSampleDbWithTopologyWithZones(app *App,
+	clusters, zones_per_cluster, nodes_per_cluster, devices_per_node int,
+	disksize uint64) error {
 
 	err := app.db.Update(func(tx *bolt.Tx) error {
 		for c := 0; c < clusters; c++ {
--- a/apps/glusterfs/volume_entry_test.go
+++ b/apps/glusterfs/volume_entry_test.go
@@ -2772,3 +2772,81 @@ func TestVolumeEntryBlockCapacityLimits(t *testing.T) {
 		tests.Assert(t, err != nil, "expected err != nil, got", err)
 	})
 }
+
+func TestVolumeCreateTooFewZones(t *testing.T) {
+	tmpfile := tests.Tempfile()
+	defer os.Remove(tmpfile)
+
+	// Create the app
+	app := NewTestApp(tmpfile)
+	defer app.Close()
+
+	origZoneChecking := ZoneChecking
+	defer func() {
+		ZoneChecking = origZoneChecking
+	}()
+
+	err := setupSampleDbWithTopologyWithZones(app,
+		1,    // clusters
+		2,    // zones_per_cluster
+		3,    // nodes_per_cluster
+		4,    // devices_per_node,
+		2*TB, // disksize)
+	)
+	tests.Assert(t, err == nil, "expected err == nil, got", err)
+
+	t.Run("ZoneChecking none replica 3", func(t *testing.T) {
+		// Without strict zone checking, a replica-3 volume fits into
+		// two zones.
+		ZoneChecking = ZONE_CHECKING_NONE
+		req := &api.VolumeCreateRequest{}
+		req.Size = 10
+		req.Durability.Type = api.DurabilityReplicate
+		req.Durability.Replicate.Replica = 2
+		v := NewVolumeEntryFromRequest(req)
+
+		err = v.Create(app.db, app.executor)
+		tests.Assert(t, err == nil, "expected err == nil, got:", err)
+	})
+
+	t.Run("ZoneChecking invalid replica 3", func(t *testing.T) {
+		// Specifying an invalid string for ZoneChecking behaves
+		// like "none". A replica-3 volume can fit into two zones.
+		ZoneChecking = "invalid"
+		req := &api.VolumeCreateRequest{}
+		req.Size = 10
+		req.Durability.Type = api.DurabilityReplicate
+		req.Durability.Replicate.Replica = 3
+		v := NewVolumeEntryFromRequest(req)
+
+		err = v.Create(app.db, app.executor)
+		tests.Assert(t, err == nil, "expected err == nil, got:", err)
+	})
+
+	t.Run("ZoneChecking strict replica 3", func(t *testing.T) {
+		// With strict zone checking, a replica-3 volume does not fit
+		// into two zones.
+		ZoneChecking = ZONE_CHECKING_STRICT
+		req := &api.VolumeCreateRequest{}
+		req.Size = 10
+		req.Durability.Type = api.DurabilityReplicate
+		req.Durability.Replicate.Replica = 3
+		v := NewVolumeEntryFromRequest(req)
+
+		err = v.Create(app.db, app.executor)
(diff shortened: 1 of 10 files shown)


obnoxxx:
Updated the PR as discussed yesterday.
* No changes to the node check as of now. (can easily be added later)
* Default of `zone_checkin` is now `none`, i.e. current behavior.

There are 3 tests failing now in `client_test.go` with `unexpected EOF`... Strange! need to investigate.

obnoxxx:
hmmm. python is also hitting `unexpected EOF`. digging deeper...

obnoxxx:
Found the issue and fixed it: The way I was generating the generator function had issues with the scope. I changed the definition of the filter functions to have the invocation within the scope of the db transaction... Tests passed locally for me now. Let's see how it goes here...

obnoxxx:
@phlogistonjohn FYI, I left the change I did to fix the issue as a "SQ: ..." commit so you can see what the fix was. This would be squashed before merging.

phlogistonjohn:
## Commit message suggestions:

> xxx TODO - apps: introduces new zone_filter module

apps: introduce zone_filter.go for checking device zones

Introduce a new file for determining information about what devices
are in what zones and using that information to check that a device
being added to a brick set belongs to a unique zone.


> xxx: merge v. simple 1-use function into its caller

apps: merge a simple one-use func into its caller

Merge a function, largely left over from previous refactoring,
into its one and only caller.

apps/glusterfs/volume_entry_test.go
```diff
@@ -100,6 +100,61 @@ func setupSampleDbWithTopologyWithZones(app *App,
 	return nil
 }
 
+// This creates and unbalanced topology in the sense that the zones
+// do not contain the same number of nodes, but zone #i contains #i
+// nodes. Each node contains zone-number many devices.
+func setupSampleDbWithUnbalancedTology(app *App,
```
phlogistonjohn: Tology -> Topology

phlogistonjohn:
I'm currently investigating the EOF issue you mentioned because the whole thing looks backwards to me. I would not like to see the db in the filter function's scope and only see it used to populate the DeviceZoneMap. So that fact that you changed it around to capture the dbtx in the worries me and I want to better understand what's happening.

obnoxxx:
I'm incorporating your changes and will push an update.

regarding the EOF issue: 
I think the changing from db to dbtx is not the fix for the EOF issue, but just the right thing to do (so that we operate exactly on the state inside the transaction).
The main point was having the invocation in the context where we have the DB. I just did the helper function with the db argument to avoid code duplication between call to PlaceAll and Replace...

obnoxxx:
updated.
Kept "SQ: print a warning when an unknown ZoneChecking is given " separate for easier review for now. will squash later.

obnoxxx:
@phlogistonjohn updated with one patch that proves that using the tx was not the fix.

apps/glusterfs/volume_entry_test.go
obnoxxx: done

phlogistonjohn:
I created the following in my local branch and it passed unit test locally. https://gist.github.com/phlogistonjohn/2888bd646203584365c059530983872e
I am trying to take the advantage of constructing all the filter functions up front and avoiding capturing a db object.

obnoxxx:
(fixed a silly gofmt issue)

obnoxxx:
@phlogistonjohn 

* Tried your patch and it works.
* It is very close to my original patch though... ==> confusing...
* I (temporarily reverted all the SQ-fixup patches and added a very minimal patch "SQ: Fix the stuff" on top of my original one. ==> That fixes it for me, locally..

I still don't fully understand it yet....

Here is the actual patch: https://github.com/heketi/heketi/pull/1518/commits/7f119c7a19dc9058b40fac4063ba27fcd262bb03

The main point is not wrapping the chosen filter in yet another filter function (which I originally only did to be expandable, since I added the node checking here as well...)

obnoxxx:
Now here is the truly minimal fix to my original PR: https://github.com/heketi/heketi/pull/1518/commits/2e821c1e9a745924855f9ee58ec3cb561f33f76b

obnoxxx:
(the dzm was not initialized in the block under the `ZONE_CHECKING_STRICT` case...

phlogistonjohn:
> Now here is the truly minimal fix to my original PR: [2e821c1](https://github.com/heketi/heketi/commit/2e821c1e9a745924855f9ee58ec3cb561f33f76b)

Weird. It's certainly an improvement that I would have suggested but it's very odd that it fixes the error.

obnoxxx:
This is now in a mergable state.
Additional tasks could be done in this PR or in a follow-up PR.

obnoxxx:
Fixed gofmt issue.
(And also re-installed the go plugin for vim that prevents me from making these mistakes on the machine that I was pushing from...)

obnoxxx:
Fixed a few comments (found while looking at PR #1520 ... :-)

keesturam:
First of all, I like the way we have an option for HEKETI_ZONE_CHECKING. This is good.

Few questions:
If HEKETI_ZONE_CHECKING is set to strict, is is mandatory to have 3 zones in heketi? If the answer is yes, then few follow up questions
Would topology load fail if there are no 3 zones?
what happens in case of existing environments where there are only 2 zones? (I don't think there is a way to modify zones using cli)

I may have more questions based on the answers.

obnoxxx:
> First of all, I like the way we have an option for HEKETI_ZONE_CHECKING. This is good.

Thanks for your feed-back!

> Few questions:
> If HEKETI_ZONE_CHECKING is set to strict, is is mandatory to have 3 zones in heketi?

No. The number of (available) zones limits the durability type of a volume, but there are more durability types than replica-3:

For instance other types are: replica-1 (aka durability none), replica-2 exist, disperse with 4+2, etc.

So 1 zone would be sufficient for replica-1/none, 2 zones for replica-2, but 6 zones would be required for disperse 4+2. Of course you are right in that 3 zones would be the minimum required to host the default type of replica-3. (And also for replica-2+arbiter.)

> If the answer is yes, then few follow up questions
> Would topology load fail if there are no 3 zones?

No. First of all because of the various types. But then also because you can later add more nodes and new zones along with them.

> what happens in case of existing environments where there are only 2 zones? (I don't think there is a way to modify zones using cli)

This is correct: We did not expose a cli way to modify the zones yet. This is legitimate RFE, but also potentially problematic since it may change volumes that previously spread across 3 zones to now only cover 2 zones... We might add a zone change command that would refuse the operation if strict zone checking is on, the volume was spread over sufficiently many zones (non-degraded),  and the operation would reduce the zones (degrade it).... This needs more thought..

In principle you can edit the zone of a node by exporting heketis' database, editing the json dump, and re-importing the modified database.


> I may have more questions based on the answers.

Sure!

apps/glusterfs/volume_entry_test.go
```diff
@@ -2905,3 +2905,86 @@ func TestVolumeCreateTooFewZones(t *testing.T) {
...
+	err := setupSampleDbWithUnbalancedTopology(app, 3, 500*GB)
+	tests.Assert(t, err == nil, "expected err == nil, got", err)
+
+	t.Run("ZoneChecking strict", func(t *testing.T) {
+		ZoneChecking = ZONE_CHECKING_STRICT
+		ok := testVolumeCreateUnbalanced(t, app)
+		tests.Assert(t, ok)
+	})
+
+	// Verify that, whithout strict zone checking, this topology produces
```
phlogistonjohn: Trivial: whithout->without

obnoxxx: fixed

PR #1520, opened 2019-02-12
Strict zone checking followup

phlogistonjohn:

PR #1518 added the core support for strict zone checking. This PR adds the ability to request strict zone checking on a per-volume basis by setting the 'user.heketi.zone-checking' volume option to 'strict'.
This PR also adds tests for the volume option, for volume expand, and device replace (set state failure).

### Notes for the reviewer

This PR depends on PR #1518 and should only be merged after that one.

### DIFF
--- a/apps/glusterfs/placer_settings.go
+++ b/apps/glusterfs/placer_settings.go
@@ -9,9 +9,12 @@
 
 package glusterfs
 
+type ZoneCheckingStrategy string
+
 const (
-	ZONE_CHECKING_NONE   = "none"
-	ZONE_CHECKING_STRICT = "strict"
+	ZONE_CHECKING_UNSET  ZoneCheckingStrategy = ""
+	ZONE_CHECKING_NONE   ZoneCheckingStrategy = "none"
+	ZONE_CHECKING_STRICT ZoneCheckingStrategy = "strict"
 )
 
 var (
--- a/apps/glusterfs/volume_entry_allocate.go
+++ b/apps/glusterfs/volume_entry_allocate.go
@@ -262,11 +262,15 @@ func (v *VolumeEntry) prepForBrickReplacement(db wdb.DB,
 	return
 }
 
-func generateDeviceFilter(db wdb.RODB) (DeviceFilter, error) {
+func (v *VolumeEntry) generateDeviceFilter(db wdb.RODB) (DeviceFilter, error) {
 
 	var filter DeviceFilter = nil
+	zoneChecking := v.GetZoneCheckingStrategy()
+	if zoneChecking == ZONE_CHECKING_UNSET {
+		zoneChecking = ZoneChecking
+	}
 
-	switch ZoneChecking {
+	switch zoneChecking {
 	case ZONE_CHECKING_STRICT:
 		dzm, err := NewDeviceZoneMapFromDb(db)
 		if err != nil {
@@ -300,7 +304,7 @@ func (v *VolumeEntry) allocBrickReplacement(db wdb.DB,
 
 		var err error
 		txdb := wdb.WrapTx(tx)
-		defaultFilter, err := generateDeviceFilter(txdb)
+		defaultFilter, err := v.generateDeviceFilter(txdb)
 		if err != nil {
 			return err
 		}
@@ -505,7 +509,7 @@ func (v *VolumeEntry) allocBricks(
 		placer := PlacerForVolume(v)
 
 		txdb := wdb.WrapTx(tx)
-		deviceFilter, err := generateDeviceFilter(txdb)
+		deviceFilter, err := v.generateDeviceFilter(txdb)
 		if err != nil {
 			return err
 		}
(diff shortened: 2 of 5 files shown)


apps/glusterfs/volume_entry_test.go
```diff
@@ -2901,6 +2901,62 @@ func TestVolumeCreateTooFewZones(t *testing.T) {
...
+		req.GlusterVolumeOptions = []string{"user.heketi.zone-checking none"}
+		v := NewVolumeEntryFromRequest(req)
+
+		err = v.Create(app.db, app.executor)
+		tests.Assert(t, err == nil, "expected err == nil, got:", err)
+	})
+
+	t.Run("VolOpt ZoneChecking invalid replica 3", func(t *testing.T) {
+		ZoneChecking = ZONE_CHECKING_NONE
+		// Can't fit a replica-3 volume into 2 zones with strict zone checking
```
obnoxxx: I realize this is an inappropriate comment that was carried over by copy'n'paste from the original PR #1518 ... will fix that PR.

apps/glusterfs/volume_entry_test.go
```diff
@@ -2901,6 +2901,62 @@ func TestVolumeCreateTooFewZones(t *testing.T) {
...
+		req.GlusterVolumeOptions = []string{"user.heketi.zone-checking foobar"}
+		v := NewVolumeEntryFromRequest(req)
+
+		err = v.Create(app.db, app.executor)
+		tests.Assert(t, err == nil, "expected err == nil, got:", err)
+	})
+
+	t.Run("VolOpt ZoneChecking strict replica 3", func(t *testing.T) {
+		ZoneChecking = ZONE_CHECKING_NONE
+		// Can't fit a replica-3 volume into 2 zones with strict zone checking
```
obnoxxx: maybe move the comment down?

obnoxxx:
@phlogistonjohn I had one more thought... If we have the per-volume request specification of the zone checking, shouldn't the volume record that as a desired behavior for future operations in it's db entry? ... E.g. if the volume was desired to be created with strict zone checking, then shouldn't for instance subsequent expansions or brick replacements also do strict zone checking? Etc ...

phlogistonjohn:
> @phlogistonjohn I had one more thought... If we have the per-volume request specification of the zone checking, shouldn't the volume record that as a desired behavior for future operations in it's db entry? ... E.g. if the volume was desired to be created with strict zone checking, then shouldn't for instance subsequent expansions or brick replacements also do strict zone checking? Etc ...

It does automatically, as the volume options are persisted in the volume's entry. This is also demonstrated by the behavior of the volume expand and brick-replacement test cases I added.

apps/glusterfs/volume_entry_test.go
phlogistonjohn: Comments in the test have been redone.

obnoxxx:
> It does automatically, as the volume options are persisted in the volume's entry. This is also demonstrated by the behavior of the volume expand and brick-replacement test cases I added.

Dang, true! Great benefit of using the vol-opts as a backdoor / hack ;-)

apps/glusterfs/zone_filter.go
```diff
@@ -0,0 +1,73 @@
+//
+// Copyright (c) 2018 The heketi Authors
```
raghavendra-talur: You might want to change the year here.

apps/glusterfs/placer_settings.go
```diff
@@ -0,0 +1,22 @@
+//
+// Copyright (c) 2017 The heketi Authors
```
raghavendra-talur: You might want to change the year here.

apps/glusterfs/zone_filter.go
```diff
@@ -0,0 +1,73 @@
...
+	})
+	return dzm, err
+}
+
+func (dzm *DeviceZoneMap) Add(deviceId string, zone int) {
+	dzm.AvailableZones[zone] = true
+	dzm.DeviceZones[deviceId] = zone
+}
+
+func (dzm *DeviceZoneMap) Filter(bs *BrickSet, d *DeviceEntry) bool {
```
raghavendra-talur: IMO, having a small description of the function would help here. It is not obvious what it does.

apps/glusterfs/volume_entry_allocate.go
```diff
@@ -262,6 +262,32 @@ func (v *VolumeEntry) prepForBrickReplacement(db wdb.DB,
...
+		dzm, err := NewDeviceZoneMapFromDb(db)
+		if err != nil {
+			return nil, err
+		}
+
+		filter = dzm.Filter
+	case ZONE_CHECKING_NONE:
+	default:
+		logger.Warning(
+			"ZoneChecking set to unknown value '%v', "+
```
raghavendra-talur: To make the log message generic, might skip the word "unknown" .

obnoxxx: Not sure I understand. This is only the `default:` case. That is exactly for unknown values.

apps/glusterfs/zone_filter.go
obnoxxx: I think the commit is actually from 2018... ;-(

PR #1521, opened 2019-02-15
executors: add a hint of possible issues to device setup error

### LINKED ISSUE #595: Improve error message when encountering initialized devices
heketi expects to be able to fully partition and format any devices it's handed. Currently when we encounter a device that's already been initialized we provide the following error message:

```
Unable to add device: Unable to execute command on <node>:   Can't initialize physical volume "<device>" of volume group "<vg>" without -ff
```

We should provide a clearer error message to let the user know why we're failing and/or what they could do to fix it. Something like:

```
Unable to add device: Device "<device>" is already initialized or contains data.
```

The changes would likely need to go in `executor/.../device.go`.

ansiwen:

Device setup can fail, for example if the partition is not wiped or is already initialized. This PR adds a hint about these possible issues to the error returned by the failed command, which can be cryptic.

### Does this PR fix issues?

Fixes #595

### Notes for the reviewer

After considering a smarter solution with individual parsing of the errors of the commands, which is error prone in itself, because there are many possible errors and the error text can also change by newer versions of the tools, this PR follows a very simplistic approach of just adding a general hint.

### DIFF
--- a/executors/cmdexec/device.go
+++ b/executors/cmdexec/device.go
@@ -59,6 +59,7 @@ func (s *CmdExecutor) DeviceSetup(host, device, vgid string, destroy bool) (d *e
 	// Execute command
 	err := rex.AnyError(s.RemoteExecutor.ExecCommands(host, commands, 5))
 	if err != nil {
+		err = fmt.Errorf("Setup of device %v failed (already initialized or contains data?): %v", device, err)
 		return nil, err
 	}
 


ansiwen:
@nixpanic @phlogistonjohn if you guys approve you also need to merge it, since I don't have rights to do it. I just made a fresh rebase, thanks.

phlogistonjohn:
I was waiting for tests to finish and let others have an opportunity to chime in. Since pleny of time has passed since my LGTM and the tests have passed I'll merge now.

PR #1527, opened 2019-02-22
extra tools: make scrub safer by blocking oob size values

phlogistonjohn:

Prevent the scrub script from creating device storage size values that
are out of bounds (OOB), either below zero or larger than the total size
of the device itself. If this condition is hit when running the scrub
script it means that db is already out of sync and the size values
should be pre-adjusted prior to running scrub.py.


phlogistonjohn:
centos-ci results are irrelevant to this change

Issue #1531, opened 2019-02-25
Failed to create volume with 4 zones with HEKETI_ZONE_CHECKING set to "strict"


Madhu-1:
### Kind of issue

Bug

### Observed behavior
failed to create volume with heketi has nodes on 4 zones

### Expected/desired behavior
volume creation should be successful

### Details on how to reproduce (minimal and precise)
set HEKETI_ZONE_CHECKING  to "strict"
do topology load with 4 zones and create volume

### Information about the environment:

- Heketi version used (e.g. v6.0.0 or master): master
- Operating system used: 
- Heketi compiled from sources, as a package (rpm/deb), or container:  source
- If container, which container image: 
- Using kubernetes, openshift, or direct install: 
- If kubernetes/openshift, is gluster running inside kubernetes/openshift or outside: 
- If kubernetes/openshift, how was it deployed (gk-deploy, openshift-ansible, other, custom): 

### Other useful information

```console
./heketi-cli topology info

Cluster Id: ea26f3116e0a020734819233f1df4d2c

    File:  true
    Block: true

    Volumes:

	Name: vol_aef697f6fc389356d593160b99fd845e
	Size: 1
	Id: aef697f6fc389356d593160b99fd845e
	Cluster Id: ea26f3116e0a020734819233f1df4d2c
	Mount: 192.168.10.103:vol_aef697f6fc389356d593160b99fd845e
	Mount Options: backup-volfile-servers=192.168.10.101,192.168.10.100,192.168.10.102
	Durability Type: replicate
	Replica: 3
	Snapshot: Disabled

		Bricks:
			Id: 5743e524c949dfa9a4b36de1af2cd1d9
			Path: /var/lib/heketi/mounts/vg_2d54c0c53b721137eee4db7db8ea1edf/brick_5743e524c949dfa9a4b36de1af2cd1d9/brick
			Size (GiB): 1
			Node: dd99ae68bd340543458f200eae14a8a9
			Device: 2d54c0c53b721137eee4db7db8ea1edf

			Id: 9682bbe7003fcd484cbfb6081eea2f81
			Path: /var/lib/heketi/mounts/vg_a7abdd5c42032584234484473585639a/brick_9682bbe7003fcd484cbfb6081eea2f81/brick
			Size (GiB): 1
			Node: eef7b512c76fac39cb0b2779d8ba323e
			Device: a7abdd5c42032584234484473585639a

			Id: d78aadcb3d674f6b8c392be45458e207
[...]

PR #1533, opened 2019-02-25
apps: ignore device id with prefix DEVICE

Madhu-1:
device info is stored with prefix DEVICE, we need to ignore these keys when we are listing the devices to add bricks.

Fixes: #1531 
Signed-off-by: Madhu Rajanna <madhupr007@gmail.com>

This PR filter out the device ID which starts with "DEVICE"

### DIFF
--- a/apps/glusterfs/zone_filter.go
+++ b/apps/glusterfs/zone_filter.go
@@ -10,6 +10,8 @@
 package glusterfs
 
 import (
+	"strings"
+
 	"github.com/boltdb/bolt"
 
 	wdb "github.com/heketi/heketi/pkg/db"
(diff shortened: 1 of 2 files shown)


Madhu-1:
@phlogistonjohn @obnoxxx @raghavendra-talur PTAL

Madhu-1:
any idea on how to add a test for this one?

phlogistonjohn:
> any idea on how to add a test for this one?

We'd need a condition where we've ensured that the device's Register function had been called so that these funky "DEVICE+stuff" entries exist in the db prior to calling the filter function. I bet the unit tests generally don't call that function when setting up the mock clusters.
You can add a test that calls that explicitly or we can try adding a more real-world case in the functional tests which ought to get those entries by way of using the proper api.
Whichever of the above seems easier.

phlogistonjohn:
@Madhu-1  see https://github.com/heketi/heketi/blob/master/tests/functional/TestSmokeTest/tests/heketi_test.go#L579  for an example test you can use as an example to base your test(s) on.

Madhu-1:
@phlogistonjohn Thanks for unblocking, added functional test PTAL

obnoxxx:
ok to test

Madhu-1:
I can do that one, is it required to add in this PR?

Madhu-1:
@phlogistonjohn added failure test if we have only 2 zones PTAL

Madhu-1:
@phlogistonjohn done PTAL

PR #1535, opened 2019-02-25
add functions for customizing cluster setup

phlogistonjohn:

Add a function to copy the ClusterEnv so customization can be applied to
a copy of the global cluster env. Add callback functions
CustomizeNodeRequest and CustomizeDeviceRequest so individual tests can
make tweaks to the cluster config without having to do the same as the
setup functions manually.

### Notes for the reviewer

This functionality can serve as a building block for future tests that need to check the behavior of the system and the stock cluster setup functions are too limiting.

### DIFF
--- a/pkg/testutils/cluster.go
+++ b/pkg/testutils/cluster.go
@@ -32,6 +32,10 @@ type ClusterEnv struct {
 	Disks   []string
 
 	HeketiUrl string
+
+	// helper callbacks
+	CustomizeNodeRequest   func(int, *api.NodeAddRequest)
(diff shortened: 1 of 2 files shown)


phlogistonjohn:
@Madhu-1  as previously discussed.

phlogistonjohn:
@raghavendra-talur @obnoxxx PTAL ... I have created this to help unblock Madhu who needs it for functional testing strict zone checking fixes

raghavendra-talur:
In future, we might need more than the index of the node as input to customize functions. We can extend it when needed.

PR #836, opened 2017-08-15
Fix 830

obnoxxx:
Resolves #830

PR #1536, opened 2019-02-26
tests: fix and enable TestHeketiCreateVolumeWithGid

ansiwen:
PR #836 was meant to re-enable the HeketiCreateVolumeWithGid test, but since it didn't add the _Test_ prefix, it wasn't.

This PR fixes the test (almost rewritten) and enables the test for real this time.

### DIFF
--- a/tests/functional/TestSmokeTest/tests/heketi_test.go
+++ b/tests/functional/TestSmokeTest/tests/heketi_test.go
@@ -155,7 +155,7 @@ func TestHeketiSmokeTest(t *testing.T) {
 	tests.Assert(t, err == nil, "expected err == nil, got:", err)
 }
 
-func HeketiCreateVolumeWithGid(t *testing.T) {
+func TestHeketiCreateVolumeWithGid(t *testing.T) {
 	na := testutils.RequireNodeAccess(t)
 	// Setup the VM storage topology
 	teardownCluster(t)
--- a/tests/functional/TestSmokeTest/tests/heketi_test.go
+++ b/tests/functional/TestSmokeTest/tests/heketi_test.go
@@ -18,7 +18,6 @@ import (
 	client "github.com/heketi/heketi/client/api/go-client"
 	"github.com/heketi/heketi/pkg/glusterfs/api"
 	"github.com/heketi/heketi/pkg/logging"
-	"github.com/heketi/heketi/pkg/remoteexec/ssh"
 	"github.com/heketi/heketi/pkg/testutils"
 	"github.com/heketi/heketi/pkg/utils"
 	"github.com/heketi/tests"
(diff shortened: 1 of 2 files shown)


obnoxxx:
ok to test

phlogistonjohn:
The newly enabled test is failing. @ansiwen please investigate, thanks!

ansiwen:
@phlogistonjohn yeah, sorry, this was mainly based on a short conversation with @obnoxxx, and I wrongly assumed it's self-explanatory. It seems like the test never got fixed then at the time? I will have a look into it.

ansiwen:
@raghavendra-talur @phlogistonjohn rewrote the test, PTAL

tests/functional/TestSmokeTest/tests/heketi_test.go
```diff
@@ -171,48 +170,33 @@ func HeketiCreateVolumeWithGid(t *testing.T) {
...
-		"sudo useradd writer1 -G writegroup -p'$6$WBG5yf03$3DvyE41cicXEZDW.HDeJg3S4oEoELqKWoS/n6l28vorNxhIlcBe2SLQFDhqq6.Pq'",
-		"sudo useradd writer2 -G writegroup -p'$6$WBG5yf03$3DvyE41cicXEZDW.HDeJg3S4oEoELqKWoS/n6l28vorNxhIlcBe2SLQFDhqq6.Pq'",
+		"sudo groupadd -f -g 2333 writegroup",
+		"grep -q ^writer1: /etc/passwd || sudo useradd writer1 -G writegroup",
+		"grep -q ^writer2: /etc/passwd || sudo useradd writer2 -G writegroup",
+		"sudo umount /mnt 2>/dev/null || true",
 		fmt.Sprintf("sudo mount -t glusterfs %v /mnt", volInfo.Mount.GlusterFS.MountPoint),
+		"sudo runuser -u writer1 -- touch /mnt/writer1testfile",
+		"sudo runuser -u writer1 -- mkdir /mnt/writer1dir",
+		"sudo runuser -u writer1 -- chmod 770 /mnt/writer1dir",
```
raghavendra-talur: The expectation is that it works without chmod. When a pv is created in kubernetes with a gid all the writes from any user who belong to the group must be able to write.

tests/functional/TestSmokeTest/tests/heketi_test.go
```diff
@@ -171,48 +170,33 @@ func HeketiCreateVolumeWithGid(t *testing.T) {
...
-	// SSH into system and create two writers belonging to writegroup gid
-	vagrantexec := na.Use(logger)
+	// SSH into system, create two writers belonging to writegroup gid and
+	// make sure both can write and the sticky group bit is set
+	exec := na.Use(logger)
 	cmd := []string{
-		"sudo groupadd writegroup",
-		"sudo useradd writer1 -G writegroup -p'$6$WBG5yf03$3DvyE41cicXEZDW.HDeJg3S4oEoELqKWoS/n6l28vorNxhIlcBe2SLQFDhqq6.Pq'",
-		"sudo useradd writer2 -G writegroup -p'$6$WBG5yf03$3DvyE41cicXEZDW.HDeJg3S4oEoELqKWoS/n6l28vorNxhIlcBe2SLQFDhqq6.Pq'",
+		"sudo groupadd -f -g 2333 writegroup",
```
phlogistonjohn: Do you want to make this conditional like the lines below?

ansiwen: no need, because of the `-f`

raghavendra-talur:
Please update commit message to link to this kubernetes document as the reason for sticky bit and chmod. 
https://github.com/kubernetes/community/blob/master/contributors/design-proposals/storage/volume-ownership-management.md

tests/functional/TestSmokeTest/tests/heketi_test.go
ansiwen: according to my tests the sticky bit only sets the group, not the permissions. the chmod was neccessary, or the mode was 2755.

raghavendra-talur: I stand corrected. SGID only sets group inheritance not permission inheritance. So we need to check if SGID is sufficient as per kubernetes or if we need something more.

raghavendra-talur: This PR addresses the problem the original test had. With that, we can merge the PR.

We will investigate on kubernetes requirements and work on it in a different PR, if required.

phlogistonjohn:
Merged with gusto. :-)

PR #1542, opened 2019-03-02
Bump kubernetes dependencies to client-go 1.10/k8s 1.13

munnerz:

This pull-request follows up on #1272 and bumps client-go to 1.10 (i.e. the latest).

From my understanding of the previous pull request, this should help unblock work on getting kubernetes/kubernetes to use go modules eventually 😄 

The changes/concerns expressed in @sigma's comment: https://github.com/heketi/heketi/pull/1272#issuecomment-450219329 are not issues, as the Kubernetes API types as serialised to JSON are still identical, despite the structure names changing.

/cc @sigma @phlogistonjohn @obnoxxx

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,12 +17,13 @@ git:
   depth: 9999999
 matrix:
   include:
-  - go: 1.9.4
-    name: Test using Go 1.9
   - go: "1.10"
     name: Test using Go 1.10
   - go: "1.11"
     name: Test using Go 1.11
+  - go: "1.12"
+    name: Test using Go 1.12
+  # TODO: bump release to use Go 1.12
   - go: "1.11"
     script: make release
     name: make release on Go 1.11
--- a/client/cli/go/cmds/heketi_storage.go
+++ b/client/cli/go/cmds/heketi_storage.go
@@ -21,8 +21,8 @@ import (
 	"github.com/heketi/heketi/pkg/glusterfs/api"
 	"github.com/spf13/cobra"
 
-	kubeapi "k8s.io/kubernetes/pkg/api/v1"
-	batch "k8s.io/kubernetes/pkg/apis/batch/v1"
+	batch "k8s.io/api/batch/v1"
+	kubeapi "k8s.io/api/core/v1"
 )
 
 type KubeList struct {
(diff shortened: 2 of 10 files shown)


munnerz:
This seems to be passing on Travis 😄 

As @sigma described, Kubernetes is using Go features that were only introduced in 1.10 (`strings.Builder`). I've therefore disable travis tests using Go 1.9 as part of this PR, and added Go 1.12 tests.

munnerz:
I am able to build the `pkg/volume/glusterfs` package using this ref of heketi/heketi :tada: https://github.com/kubernetes/kubernetes/compare/master...munnerz:bump-heketi?expand=1

liggitt:
eyes on this would be appreciated. I'm looking to move forward with go modules in k8s 1.15 and getting this ironed out would be helpful

phlogistonjohn:
retest this please

phlogistonjohn:
(centos ci uses a different magic incantation ;-))

phlogistonjohn:
One thing we need to sort out with the PR is the spate of unrelated changes that occurs in glide. This is perpetually an issue with heketi (at least how we use glide). I'm not against updating dependencies but we should not be updating a bunch of stuff unrelated to the core work.
If someone wanted to accelerate this I'd appreciate it someone could look into possibly making a PR that first does a general update and/or locks down more of our depencies. Then, we can base new PRs that need to update dependencies upon that work.

obnoxxx:
I fully agree, the patch looks good. But we need to untangle the `glide.lock` changes.

obnoxxx:
@munnerz I (and I guess others) are more than happy to help with the reworking of the glide.lock patch, if you don't have the time or motivation to do it. :-)

PS: we  really have to get rid of glide...

obnoxxx:
The commit `Bump to client-go 1.10` should read `v10.0.0` instead `1.10` (and so should the PR title, or what am I missing.

Also, we could squash the two patches for moving to v8.0.0 and then to v10.0.0 into one unless there's a good reason to keep the two steps separate. (If you wanted to keep the authorship of the original comment, I could live with that.)

liggitt:
@munnerz any chance of reviving this? looks like this might block getting k8s.io/kubernetes to go1.13 (xref https://github.com/kubernetes/kubernetes/issues/82506)

phlogistonjohn:
PR is obsoleted by  #1649 which was merged. Closing this PR as the goals of this PR are now met. Thank you for your contribution.

PR #1544, opened 2019-03-05
functional tests: simplified teardown

phlogistonjohn:

These changes refactor and simplfy the node/device teardown behavior of the function tests (via the testutils package).

### Notes for the reviewer

This started as part of an effort to improve the speed of the functional tests. However it seems that the changes are good enough on their own to share. :-)


phlogistonjohn:
retest this please

phlogistonjohn:
It turns out you can not run the device delete client api call in parallel. Not sure if this issue lies more in gluster (peer remove) or in heketi itself, but the goal here was to simplify (and speed up) the functional tests. Looking into the reason for the parallelization problems will come later.

PR #1545, opened 2019-03-05
Make request poll delay configurable, and speed up tests

phlogistonjohn:

These changes remove hard coded delays in the api functions and centralizes that logic. It makes the poll delay configurable when the client type is being created and uses this configurable delay to help speed up the unit and functional tests.

### Notes for the reviewer

This has been sitting on my laptop for too long and I wanted to get it out there even in this unfinished state. Note that this PR is based on #1544 so please review and merge that first.

### DIFF
--- a/client/api/go-client/block_volume.go
+++ b/client/api/go-client/block_volume.go
@@ -16,7 +16,6 @@ import (
 	"bytes"
 	"encoding/json"
 	"net/http"
-	"time"
 
 	"github.com/heketi/heketi/pkg/glusterfs/api"
 	"github.com/heketi/heketi/pkg/utils"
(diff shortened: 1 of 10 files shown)


phlogistonjohn:
Based on the recent runs of other PRs the functional tests take 2 hours. This PR completed in 1 hour 30 mins. This is only for 1/1 run and the changes only affect the TestSmokeTest and TestErrorHandling directories, but it seems like it might have been worth the effort.

raghavendra-talur:
@centos-ci retest this please

phlogistonjohn:
Interesting. The tests never ran in centos ci, it seems to have failed due to a patch failing to apply. Investigating....

PR #1548, opened 2019-03-07
apps: refactor app creation and setup

phlogistonjohn:

Previously, the NewApp function was only able to return the App or a
nil. This caused the error handling in functions that called NewApp to
be rather awkward and partly led to lack of good error reporting in
the main.go file.

### DIFF
--- a/apps/glusterfs/app.go
+++ b/apps/glusterfs/app.go
@@ -92,10 +92,23 @@ type App struct {
 	xo *mockexec.MockExecutor
 }
 
-// Use for tests only
-func NewApp(conf *GlusterFSConfig) *App {
-	var err error
+// NewApp constructs a new glusterfs application object and populates
+// the internal structures according to the passed configuration (
+// and environment). If an error occurs the app object will be nil
+// and the error type will be populated.
+func NewApp(conf *GlusterFSConfig) (*App, error) {
 	app := &App{}
+	err := app.setup(conf)
+	if err != nil {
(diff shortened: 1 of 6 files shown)


main.go
```diff
@@ -388,12 +388,12 @@ func setWithEnvVariables(options *config.Config) {
...
 		err := recover()
-		if a == nil {
-			fmt.Fprintln(os.Stderr, "ERROR: Unable to start application")
-			os.Exit(1)
-		} else if err != nil {
+		if err != nil {
 			fmt.Fprintf(os.Stderr, "ERROR: Unable to start application: %s\n", err)
 			os.Exit(1)
+		} else if a == nil {
+			fmt.Fprintln(os.Stderr, "ERROR: Unable to start application")
```
raghavendra-talur: Shouldn't this case be unreachable with the refactoring?

apps/glusterfs/app.go
```diff
@@ -93,9 +93,17 @@ type App struct {
 }
 
 // Use for tests only
```
raghavendra-talur: Can we please get rid of this line? This PR is the closest edit we have got. Pretty please :)

phlogistonjohn: I meant to and it slipped my mind. Thank you for catching that!!!

main.go
phlogistonjohn: yes. Will update

PR #1549, opened 2019-03-08
container: add ability to archive db at container start up

phlogistonjohn:

Add code to the container start-up script that uses the new
HEKETI_DB_ARCHIVE_PATH and HEKETI_DB_ARCHIVE_COUNT environment variables
to keep a limited set of archival copies of the heketi db as it was
before the server process started. If HEKETI_DB_ARCHIVE_PATH is unset
the code block is skipped. Otherwise, a compressed copy of the db will
be made, keeping up to HEKETI_DB_ARCHIVE_COUNT copies of the db.

### Does this PR fix issues?

Fixes [rhbz#1686894](https://bugzilla.redhat.com/show_bug.cgi?id=1686894)

### DIFF
--- a/extras/container/heketi-start.sh
+++ b/extras/container/heketi-start.sh
@@ -7,7 +7,13 @@
 : "${HEKETI_PATH:=/var/lib/heketi}"
 : "${BACKUPDB_PATH:=/backupdb}"
 : "${TMP_PATH:=/tmp}"
+: "${HEKETI_DB_ARCHIVE_PATH:=${HEKETI_PATH}/archive}"
 LOG="${HEKETI_PATH}/container.log"
+
+# allow disabling the archive path
+case "${HEKETI_DB_ARCHIVE_PATH}" in
+    -|/dev/null) HEKETI_DB_ARCHIVE_PATH="" ;;
+esac
 
 info() {
     echo "$*" | tee -a "$LOG"
@@ -20,6 +26,60 @@ error() {
 fail() {
     error "$@"
     exit 1
+}
+
+archive_db() {
+    # create an archive file for the previous copy of the db
+    # before the server starts up
+    mkdir -p "${HEKETI_DB_ARCHIVE_PATH}"
+
+    # only archive the db if the db content is valid
+    if ! /usr/bin/heketi db export --dbfile "${HEKETI_PATH}/heketi.db" --jsonfile - >/dev/null ; then
+        error "Unable to export db. DB contents may not be valid"
+        return 1
+    fi
+
+    mapfile -t afiles < <(find "${HEKETI_DB_ARCHIVE_PATH}" \
+        -maxdepth 1 -name 'heketi.db-archive-*.gz' | sort)
+    newname="heketi.db-archive-$(date +%Y-%m-%d.%s).gz"
+    newpath="${HEKETI_DB_ARCHIVE_PATH}/${newname}"
+    gzip -9 -c "${HEKETI_PATH}/heketi.db" > "${newpath}"
+    if [[ $? -ne 0 ]]; then
+        rm -f "${newpath}"
+        error "Unable to create archive"
+        return 1
+    fi
+    sha256sum "${newpath}" > "${newpath}.sha256"
+    if [[ $? -ne 0 ]]; then
+        rm -f "${newpath}" "${newpath}.sha256"
+        error "Unable to get sha256 sum of archive"
+        return 1
+    fi
+    # if the new file is a dupe, immediately prune it
+    if [ "${#afiles[@]}" -gt 0 ]; then
+        sum="$(awk '{print $1}' "${newpath}.sha256")"
+        prevpath="${afiles[-1]}"
+        if grep -q "$sum" "${prevpath}.sha256" >/dev/null; then
+            # found same hash as previous db, remove new copy
+            rm -f "${newpath}" "${newpath}.sha256"
+        fi
+    fi
+    return 0
+}
+
+prune_archives() {
+    # prune old archive copies
+    HEKETI_DB_ARCHIVE_COUNT="${HEKETI_DB_ARCHIVE_COUNT:-5}"
+    mapfile -t afiles < <(find "${HEKETI_DB_ARCHIVE_PATH}" \
+        -maxdepth 1 -name 'heketi.db-archive-*.gz' | sort)
+    if [[ "${#afiles[@]}" -gt ${HEKETI_DB_ARCHIVE_COUNT} ]]; then
+        curr=0
+        count=$((${#afiles[@]}-HEKETI_DB_ARCHIVE_COUNT))
+        while [[ $curr -lt $count ]]; do
+            rm -f "${afiles[$curr]}"*
+            curr=$((curr+1))
+        done
+    fi
 }
 
 info "Setting up heketi database"
@@ -71,6 +131,13 @@ if [[ $flock_status -ne 0 ]]; then
     fail "Database file is read-only"
 fi
 
+# this creates archival copies of the db if needed for disaster
+# recovery purposes
+if [[ "${HEKETI_DB_ARCHIVE_PATH}" && -f "${HEKETI_PATH}/heketi.db" ]]; then
+    archive_db && prune_archives
+fi
+
+# this is used to restore secret based backups
 if [[ -d "${BACKUPDB_PATH}" ]]; then
     if [[ -f "${BACKUPDB_PATH}/heketi.db.gz" ]] ; then
         gunzip -c "${BACKUPDB_PATH}/heketi.db.gz" > "${TMP_PATH}/heketi.db"
--- a/extras/container/heketi-start.sh
+++ b/extras/container/heketi-start.sh
@@ -8,6 +8,7 @@
 : "${BACKUPDB_PATH:=/backupdb}"
 : "${TMP_PATH:=/tmp}"
 : "${HEKETI_DB_ARCHIVE_PATH:=${HEKETI_PATH}/archive}"
+HEKETI_BIN="/usr/bin/heketi"
 LOG="${HEKETI_PATH}/container.log"
 
 # allow disabling the archive path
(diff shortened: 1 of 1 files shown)


phlogistonjohn:
Does anyone want to enable this by default? Currently the environment (kube template,etc) needs to explicitly enable it but we could choose to make the archive path set by default, such as ${HEKETI_PATH}/archive.

extras/container/heketi-start.sh
```diff
@@ -70,6 +70,32 @@ if [[ $flock_status -ne 0 ]]; then
...
+# recovery purposes
+if [[ "${HEKETI_DB_ARCHIVE_PATH}" ]]; then
+    HEKETI_DB_ARCHIVE_COUNT="${HEKETI_DB_ARCHIVE_COUNT:-3}"
+    # create an archive file for the previous copy of the db
+    # before the server starts up
+    mkdir -p "${HEKETI_DB_ARCHIVE_PATH}"
+    newname="heketi.db-archive-$(date +%Y-%m-%d.%s).gz"
+    gzip -9 -c "${HEKETI_PATH}/heketi.db" > "${newname}"
+    if [[ $? -ne 0 ]]; then
+        error "Unable to create archive"
```
raghavendra-talur: To alarm users slightly less, we should explain in log message that archive failure is expected in first run.

extras/container/heketi-start.sh
```diff
@@ -70,6 +70,32 @@ if [[ $flock_status -ne 0 ]]; then
     fail "Database file is read-only"
 fi
 
+# this creates archival copies of the db if needed for disaster
```
raghavendra-talur: Isn't it better to to this after line 114? That would ensure the archival works with secret based db storage.

extras/container/heketi-start.sh
phlogistonjohn: Good point. I'll actually just skip the whole block if the db didn't previously exist and there's nothing to archive.

raghavendra-talur:
> Does anyone want to enable this by default? Currently the environment (kube template,etc) needs to explicitly enable it but we could choose to make the archive path set by default, such as ${HEKETI_PATH}/archive.

Yes, my preference would be to do it by default and only if set to "disabled" or similar, it should be disabled.

extras/container/heketi-start.sh
phlogistonjohn: I wan't sure if that's actually the behavior we wanted. That next block overwrites whatever was the previous "database file" with the externally backed up copy. If we move this block down we lose the ability to know what was there prior to that restore. Since we have no control over what's in that location I was worried we might be losing interesting state if we copied the backup into the archive rather than the current contents of heketi.db.

raghavendra-talur: Ok, that is a valid point. I am happy with archival being enabled on for users who don't store on secret.

extras/container/heketi-start.sh
```diff
@@ -6,8 +6,14 @@
 
 : "${HEKETI_PATH:=/var/lib/heketi}"
 : "${BACKUPDB_PATH:=/backupdb}"
+: "${HEKETI_DB_ARCHIVE_PATH:=${HEKETI_PATH}/archive}"
 LOG="${HEKETI_PATH}/container.log"
 
+# allow disabling the archive path
+case "${HEKETI_DB_ARCHIVE_PATH}" in
+    -|/dev/null) HEKETI_DB_ARCHIVE_PATH="" ;;
```
raghavendra-talur: Good one!

phlogistonjohn:
> > Does anyone want to enable this by default? Currently the environment (kube template,etc) needs to explicitly enable it but we could choose to make the archive path set by default, such as ${HEKETI_PATH}/archive.
> 
> Yes, my preference would be to do it by default and only if set to "disabled" or similar, it should be disabled.

done

raghavendra-talur:
I will wait till start of next week before merging this PR as it touches startup code.

phlogistonjohn:
I just realized that this approach is flawed as it does not protect against repeated starts of the pod. If the pod is restarted it will recycle the old DB contents out in favor of copies of the same data. In other words, this won't actually protect against the scenerios we care about, unfortunately.

phlogistonjohn:
I've made two major changes to the approach of the archival:
1. The db will not be archived if the db can't be serialized. This is a sanity check against archiving "corrupt" db contents.
2. The compressed db is hashed and if the hash matches the previous archives hash, the new one is discarded to prevent new copies from "rolling over" good old copies if the pod restarts a lot.

Before I remove the "do no merge", @raghavendra-talur could you please give this a quick sanity check?

extras/container/heketi-start.sh
```diff
@@ -21,6 +27,60 @@ fail() {
     exit 1
 }
 
+archive_db() {
+    # create an archive file for the previous copy of the db
+    # before the server starts up
+    mkdir -p "${HEKETI_DB_ARCHIVE_PATH}"
+
+    # only archive the db if the db content is valid
+    if ! $HEKETI_PATH/heketi db export --dbfile "${HEKETI_PATH}/heketi.db" --jsonfile - >/dev/null ; then
```
raghavendra-talur: Heketi path is /var/lib/heketi  and might not have heketi binary.

raghavendra-talur: I am also not sure if jsonfile takes "-". Need to test it out.

raghavendra-talur: Tested the jsonfile part, it creates a file with name "-".

phlogistonjohn: Weird. I thought I added that... Yep:
```
func DbDump(jsonfile string, dbfile string) error {

»»»»var fp *os.File
»»»»if jsonfile == "-" {
»»»»»»»»fp = os.Stdout
»»»»} else {
```
Are you testing with something other than master?

phlogistonjohn: Fixed the binary path. That was debugging stuff I accidentally left in.

extras/container/heketi-start.sh
```diff
@@ -21,6 +28,60 @@ fail() {
...
 }
 
+archive_db() {
+    # create an archive file for the previous copy of the db
+    # before the server starts up
+    mkdir -p "${HEKETI_DB_ARCHIVE_PATH}"
+
+    # only archive the db if the db content is valid
+    if ! "$HEKETI_BIN" db export --dbfile "${HEKETI_PATH}/heketi.db" --jsonfile - >/dev/null ; then
+        error "Unable to export db. DB contents may not be valid"
```
raghavendra-talur: Check: ensure db export returns a negative for all possible corrupt dbs.

extras/container/heketi-start.sh
raghavendra-talur: Thanks! I was using a heketi binary from a old released branch.

extras/container/heketi-start.sh
phlogistonjohn: I tested it locally and if the db is corrupted (bolt can't read it) the command exited with 2. If the db doesn't exist it exits with 1 but we don't expect the function to be run in that case.

extras/container/heketi-start.sh
```diff
@@ -91,7 +159,7 @@ fi
 # heketi service back to the foreground again.
 if [[ "$(stat -c %s ${HEKETI_PATH}/heketi.db)" == 0 && -n "${HEKETI_TOPOLOGY_FILE}" ]]; then
     # start hketi in the background
-    /usr/bin/heketi --config=/etc/heketi/heketi.json &
+    "$HEKETI_BIN" --config=/etc/heketi/heketi.json &
```
nixpanic: This looks weird. Why not just do it like

    ${HEKETI_BIN} --config=/etc/heketi/heketi.json &

Or is there really an expectation that "HEKETI_BIN" may not have any spaces? Maybe it would be useful to be able to pass additional options that way...

extras/container/heketi-start.sh
```diff
@@ -113,5 +181,5 @@ if [[ "$(stat -c %s ${HEKETI_PATH}/heketi.db)" == 0 && -n "${HEKETI_TOPOLOGY_FIL
     fg %1
 else
     # just start in the foreground
-    exec /usr/bin/heketi --config=/etc/heketi/heketi.json
+    exec "$HEKETI_BIN" --config=/etc/heketi/heketi.json
```
nixpanic: Same here, having " around an executable looks weird. Just use this?

    exec ${HEKETI_BIN} --config=/etc/heketi/heketi.json

extras/container/heketi-start.sh
phlogistonjohn: Force of habit mostly.

phlogistonjohn:
spurious centos-ci failure. This script is not executed as part of centosci

raghavendra-talur:
@phlogistonjohn Let us decide tomorrow if we want to take this in release 9.

extras/container/heketi-start.sh
ansiwen: Well, in theory the path to heketi bin could contain spaces, then you need the double quotes.

extras/container/heketi-start.sh
ansiwen: So `:=` applies also to the empty string, not only to undefined variables?

extras/container/heketi-start.sh
```diff
@@ -22,6 +29,60 @@ fail() {
...
+    mkdir -p "${HEKETI_DB_ARCHIVE_PATH}"
+
+    # only archive the db if the db content is valid
+    if ! "$HEKETI_BIN" db export --dbfile "${HEKETI_PATH}/heketi.db" --jsonfile - >/dev/null ; then
+        error "Unable to export db. DB contents may not be valid"
+        return 1
+    fi
+
+    mapfile -t afiles < <(find "${HEKETI_DB_ARCHIVE_PATH}" \
+        -maxdepth 1 -name 'heketi.db-archive*.gz' | sort)
```
ansiwen: Nitpick: could be `heketi.db-archive-*.gz`

extras/container/heketi-start.sh
```diff
@@ -22,6 +29,60 @@ fail() {
...
+        fi
+    fi
+    return 0
+}
+
+prune_archives() {
+    # prune old archive copies
+    HEKETI_DB_ARCHIVE_COUNT="${HEKETI_DB_ARCHIVE_COUNT:-5}"
+    mapfile -t afiles < <(find "${HEKETI_DB_ARCHIVE_PATH}" \
+        -maxdepth 1 -name 'heketi.db-archive*.gz' | sort)
```
ansiwen: Same

extras/container/heketi-start.sh
phlogistonjohn: done

PR #1558, opened 2019-03-15
Limit volumes per Gluster cluster

raghavendra-talur:
With this change, we can limit volumes that Heketi allows to create in a Gluster cluster.

This change is a departure from the previous behavior of heketi. Earlier, heketi did not limit the volumes per Gluster trusted storage pool but we have seen cases where the glusterd startup times out if the volumes in the trusted storage pool cross 1000. Hence, the default limit is set to 1000.

To override the default value admins should set the `max_volumes_per_cluster option` in `glusterfs` section of heketi.json.
    
Negative value means no limit.
Value set to 0 means use the default.
Positive and valid int value set the limit to that integer.

When env var HEKETI_GLUSTER_MAX_VOLUMES_PER_CLUSTER is set, it overrides the default value and the value provided in heketi config.
    
### DIFF
--- a/apps/glusterfs/app.go
+++ b/apps/glusterfs/app.go
@@ -11,6 +11,7 @@ package glusterfs
 
 import (
 	"fmt"
+	"math"
 	"net/http"
 	"os"
 	"strconv"
@@ -411,10 +412,18 @@ func (a *App) setAdvSettings() {
 		logger.Info("Post Request Volume Options: %v", a.conf.PostReqVolumeOptions)
 		PostReqVolumeOptions = a.conf.PostReqVolumeOptions
 	}
-
 	if a.conf.ZoneChecking != "" {
 		logger.Info("Zone checking: '%v'", a.conf.ZoneChecking)
 		ZoneChecking = ZoneCheckingStrategy(a.conf.ZoneChecking)
+	}
+	if a.conf.MaxVolumesPerCluster < 0 {
+		logger.Info("Volumes per cluster limit is removed as it is set to %v", a.conf.MaxVolumesPerCluster)
+		maxVolumesPerCluster = math.MaxInt32
+	} else if a.conf.MaxVolumesPerCluster == 0 {
+		logger.Info("Volumes per cluster limit is set to default value of %v", maxVolumesPerCluster)
+	} else {
+		logger.Info("Volumes per cluster limit is set to %v", a.conf.MaxVolumesPerCluster)
+		maxVolumesPerCluster = a.conf.MaxVolumesPerCluster
 	}
 }
 
--- a/apps/glusterfs/app.go
+++ b/apps/glusterfs/app.go
@@ -377,6 +377,14 @@ func (a *App) setFromEnvironmentalVariable() {
 	if "" != env {
 		a.conf.ZoneChecking = env
 	}
+
+	env = os.Getenv("HEKETI_GLUSTER_MAX_VOLUMES_PER_CLUSTER")
+	if env != "" {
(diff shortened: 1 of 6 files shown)


apps/glusterfs/app.go
```diff
@@ -403,11 +403,19 @@ func (a *App) setAdvSettings() {
...
 		PostReqVolumeOptions = a.conf.PostReqVolumeOptions
 	}
-
 	if a.conf.ZoneChecking != "" {
 		logger.Info("Zone checking: '%v'", a.conf.ZoneChecking)
 		ZoneChecking = ZoneCheckingStrategy(a.conf.ZoneChecking)
 	}
+	if a.conf.MaxVolumesPerCluster < 0 {
+		logger.Info("Volumes per cluster limit is removed as it is set to %v", a.conf.MaxVolumesPerCluster)
+		MaxVolumesPerCluster = a.conf.MaxVolumesPerCluster
```
phlogistonjohn: Does this work the way you intend it to? Seems to me that the volume count is always going to be larger than a negative number. Wouldn't it make more sense to set this to MAX_INT or the golang equivalent?

apps/glusterfs/cluster_entry.go
```diff
@@ -156,6 +156,11 @@ func (c *ClusterEntry) VolumeDelete(id string) {
 	c.Info.Volumes = sortedstrings.Delete(c.Info.Volumes, id)
 }
 
+// VolumeCount returns number of volumes in cluster *including* the pending ones
+func (c *ClusterEntry) VolumeCount() int {
```
phlogistonjohn: Can we start out by making this a private function? I don't think its needed outside of apps/glusterfs right?

raghavendra-talur: yes! Will do.

apps/glusterfs/app.go
raghavendra-talur: No it doesn't. I would prefer to abstract it out by implementing a function cluster.volumeCount() > limitOfVolumesPerCluster()

raghavendra-talur: I changed to MAX_INT32.

apps/glusterfs/app_config.go
```diff
@@ -36,6 +36,7 @@ type GlusterFSConfig struct {
 	PreReqVolumeOptions  string `json:"pre_request_volume_options"`
 	PostReqVolumeOptions string `json:"post_request_volume_options"`
 	ZoneChecking         string `json:"zone_checking"`
+	maxVolumesPerCluster int    `json:"max_volumes_per_cluster"`
```
phlogistonjohn: Doesn't it need to be "MaxVolumesPerCluster" to make the JSON parsing happy? I think it's also needed if we expect to programmatically set the config as we do in some of the newer functional tests.

raghavendra-talur: done

phlogistonjohn:
The test looks OK with one minor nit that occurs in all three t.Run instances. The cleanup code is not in a defer so it will only run in a success case.
This means a failure in an earlier test will probably make the other sub-tests fail too. I'm not sure we care though.

raghavendra-talur:
> The test looks OK with one minor nit that occurs in all three t.Run instances. The cleanup code is not in a defer so it will only run in a success case.
> This means a failure in an earlier test will probably make the other sub-tests fail too. I'm not sure we care though.

Thanks, I had not understood fully how t.Run interacts with t.FailNow(). Updated the test to cleanup volumes in every t.Run().

PR #1559, opened 2019-03-20
Add a default admin mode option

phlogistonjohn:

Add the ability to set a config param or environment variable
(HEKETI_DEFAULT_STATE) that will be applied on server start up.  This is
useful in the case that you want heketi to start up but do not want the
server to immediately start serving new requests.

Note: this does not change the behavior of internally driven changes
like the background operations cleanup. That would need to be disabled
independently.


phlogistonjohn:
retest this please

PR #1560, opened 2019-03-22
go-client: disable KeepAlives

ateleshev:
This PR fixed connection leak in heketi go-client (TLS enabled)

You can see same problem requested to the golang repository:
https://github.com/golang/go/issues/24719#issuecomment-379371501

### Notes for the reviewer
heketi: 8.0.0
golang: 1.12.1

### Details
Our watcher requests to heketi one time per minute to get devices info. After one day works we have been seeing thousands  of established connections. We use TLS between our watcher and heketi server.

### DIFF
--- a/client/api/go-client/client.go
+++ b/client/api/go-client/client.go
@@ -181,7 +181,8 @@ func (c *Client) doBasic(req *http.Request) (*http.Response, error) {
 	httpClient := &http.Client{}
 	if c.tlsClientConfig != nil {
 		httpClient.Transport = &http.Transport{
-			TLSClientConfig: c.tlsClientConfig,
+			TLSClientConfig:   c.tlsClientConfig,
+			DisableKeepAlives: true,
 		}
 	}
 	httpClient.CheckRedirect = c.checkRedirect
--- a/client/api/go-client/client.go
+++ b/client/api/go-client/client.go
@@ -39,6 +39,10 @@ const (
 	MAX_DELAY = 30
 )
 
+type closeIdler interface {
+	CloseIdleConnections()
+}
+
 type ClientTLSOptions struct {
 	// directly borrow the field names from crypto/tls
 	InsecureSkipVerify bool
@@ -160,10 +164,6 @@ func (c *Client) SetTLSOptions(o *ClientTLSOptions) error {
 // isHTTPClientCloseIdler returns true if HTTP client realize closeIdler interface
 // Details: https://github.com/golang/go/blob/release-branch.go1.12/src/net/http/client.go#L843
 func (c *Client) isHTTPClientCloseIdler(i interface{}) bool {
-	type closeIdler interface {
-		CloseIdleConnections()
-	}
-
 	_, ok := i.(closeIdler)
 	return ok
 }
@@ -171,10 +171,6 @@ func (c *Client) isHTTPClientCloseIdler(i interface{}) bool {
 // closeIdleConnections performs closing idle connections if it possible
 // Details: https://github.com/golang/go/blob/release-branch.go1.12/src/net/http/client.go#L843
 func (c *Client) closeIdleConnections(i interface{}) {
-	type closeIdler interface {
-		CloseIdleConnections()
-	}
-
 	if closer, ok := i.(closeIdler); ok {
 		closer.CloseIdleConnections()
 	}
@@ -208,7 +204,8 @@ func (c *Client) createHTTPTransport() *http.Transport {
 		TLSClientConfig:       c.tlsClientConfig,
 	}
 
-	// For golang 1.11 and less we need disable KeepAlives
+	// If HTTP transport does not support to close idle
+	// connections manually, we need disable KeepAlives
 	if !c.isHTTPClientCloseIdler(tr) {
 		tr.DialContext = (&net.Dialer{
 			Timeout:   30 * time.Second,
--- a/client/api/go-client/client.go
+++ b/client/api/go-client/client.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"math/rand"
+	"net"
 	"net/http"
 	"strconv"
 	"time"
@@ -59,6 +60,9 @@ type Client struct {
 	key      string
 	user     string
 	throttle chan bool
+
+	// http client
+	httpClient *http.Client
 
 	// configuration for TLS support
 	tlsClientConfig *tls.Config
@@ -115,6 +119,11 @@ func NewClientNoAuth(host string) *Client {
 	return NewClient(host, "", "")
 }
 
+// Close performs reset HTTP client
+func (c *Client) Close() {
+	c.resetHTTPClient()
+}
+
 // SetTLSOptions configures an existing heketi client for
 // TLS support based on the ClientTLSOptions.
 func (c *Client) SetTLSOptions(o *ClientTLSOptions) error {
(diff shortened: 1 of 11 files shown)


phlogistonjohn:
@raghavendra-talur PTAL. I wonder if this option should be universally applied, given that we've also seen the client consume many sockets (fds) on non-HTTPS configurations.

phlogistonjohn:
ok to test

ateleshev:
HttpClient hasn't reused, that is why it make sense to disable KeepAlive globally.
@phlogistonjohn Can I update it for HTTP connections?

phlogistonjohn:
I agree, and I think it's worth experimenting with. So go ahead and make the change and we'll see what it looks like.

ateleshev:
@phlogistonjohn PTAL

ateleshev:
@phlogistonjohn I have been testing reusing of http client with enabled KeepAlives, and it's works fine. It significantly decreased count of system calls. How do you think, is it make sense to do it in this PR?

phlogistonjohn:
@ateleshev I wonder if it would make sense to have two (mutually exclusive) PRs, we'd have to pick one in the end, but at least we'd be able to see the test suite run with both options.
Also hoping @raghavendra-talur will provide some input because he's looked into this topic in the past.

raghavendra-talur:
@ateleshev Thanks for the contribution!

There are two problems that we would like to solve
1. You are hitting problem 1 where you are see "seeing thousands of established connections" after one day of heketi client operations in a long running process.
2. We have seen many ports in TIME_WAIT state when many heketi client operations are performed rapidly(in different processes).

@phlogistonjohn and I ran some tests to compare the behavior of heketi-client with regards to port consumption/established connections.

The test case is

a. create 5 volumes
b. list the volumes 100 times

for i in {1..5}; do heketi-cli volume create --size=5 ; done
for i in {1..100}; do heketi-cli volume list ; done

The numbers we get are 
On the master branch: approximately 600
On #1298: approximately 100 
This PR: approximately 600

Therefore, we would like to see your version of reusing the client and KeepAlive enabled. If you would send the updated patch and if the numbers come close to 100, we would like to take your PR and make it part of release 9.

If you don't have time for it, please let us know and we will bring #1298 into a more merge-able state. Ideally, we would like to hear from you by 3rd April to make a decision.

ateleshev:
@raghavendra-talur  I agree with you, my current realisation in this PR solve only first problem.
You may know, standard http package has some problems with closing of idle connections:
 - https://github.com/golang/go/issues/24739
I going to update this PR tomorrow. And we will discuss it

ateleshev:
@raghavendra-talur PTAL

phlogistonjohn:
The changes have suddenly got a lot more complicated looking to me. Although the bulk of the lines just appear to be `defer client.close()` type things. First off, I don't understand if the defers are required. How necessary are they?
Next, I am quite confused about the CloseIdler stuff which appears to be new in go1.12, right? If so that's worrysome to me because we a) need to support older versions of go (typically we want users to be able to compile on centos) b) don't even test on go 1.12 yet. Is this a case of older versions don't reap the benefits, but is still compatible? If so, what are those benefits?

Overall I liked the simplicity, if perhaps limited benefits of the early versions of the PR. However, I'm willing to be convinced of the benefits of this approach but I'm quite skittish to take the current form of the series for release 9. I'm not sure if you care about our release @ateleshev but I'm leaning toward just releasing without the client improvements but trying to get them squared away soon so they can be used in master for a while, any thoughts @raghavendra-talur ?

raghavendra-talur:
@ateleshev Thanks for the links which explain the problem with http library in golang.

@phlogistonjohn @ateleshev I did a quick test on the latest PR using the same test as before. 


go version

```
$ go version
go version go1.11.5 linux/amd64
```

test case
```
a. create 5 volumes
b. list the volumes 100 times
```

The number of connections is approximately 100, which is same as in PR #1298 .

So functionally, we seem to be doing good with the latest update. I do agree with @phlogistonjohn that we need to do a few more things before merging the PR:

a. Write test case(s) to ensure we are not breaking anything
b. Understand the difference between MaxIdleConnsPerHost and CloseIdleConnections
c. Think about the implications of go version requirements.

@ateleshev If you are not waiting on release 9 to have this PR, then we can make the release without this PR and merge it after above tasks are done.

ateleshev:
The PR https://github.com/heketi/heketi/pull/1298 does not solve problem with thousands of established connections
```
c.httpClient.Transport = &http.Transport{
    MaxIdleConnsPerHost: 0,
}
```
MaxIdleConnsPerHost = 0 it means - use default value from standard go http library.

https://golang.org/src/net/http/transport.go#L57
```
// DefaultMaxIdleConnsPerHost is the default value of Transport's
// MaxIdleConnsPerHost.
const DefaultMaxIdleConnsPerHost = 2
```

https://golang.org/src/net/http/transport.go#L730
```
func (t *Transport) maxIdleConnsPerHost() int {
	if v := t.MaxIdleConnsPerHost; v != 0 {
		return v
	}
	return DefaultMaxIdleConnsPerHost
}
```
As you can see, idle connections will not create if specified  MaxIdleConnsPerHost < 0 or DisableKeepAlives = false
https://golang.org/src/net/http/transport.go#L743
```
func (t *Transport) tryPutIdleConn(pconn *persistConn) error {
	if t.DisableKeepAlives || t.MaxIdleConnsPerHost < 0 {
		return errKeepAlivesDisabled
	}
```

There is two issues which affect http connections:
- https://github.com/golang/go/issues/24739
- https://github.com/golang/go/issues/24719

And you can read in this comment to issue:
https://github.com/golang/go/issues/24719#issuecomment-379371501
```
It does not happen with DefaultTransport, or calling CloseIdleConnections.
```

@phlogistonjohn @raghavendra-talur  I clearly understand, the release it is your responsibility, and you must decide which PR make sense to merge

ateleshev:
@phlogistonjohn ```defer client.close()``` - it is solution which guaranteed close all open idle connections correctly before process exit (it needs only if connection use TLS), and it will work only for golang 1.12, in the previous versions of go KeepAlive will be disabled.

```CloseIdleConnections``` will be helpful with TLS certificates rotation in demonized applications.

phlogistonjohn:
I'd love to see some improvement in this area but this PR has gone quite stale. If we could make some small incremental improvements in this area it would be much easier to review and less change to unknown risks to the current users.
If you're not interested in perusing this approach I understand but I think we'd need to close this PR. If I don't hear back in a couple of weeks I will close this out.

ateleshev:
Hello @phlogistonjohn. I need some time to recheck this PR. Would you like to remove ```defer heketi.Close()```?

phlogistonjohn:
To be honest I didn't expect you to respond so quickly so I am quite rusty on the technical aspects of the PR as well. 
I do remember that it started off as a smaller set of changes and I'd like to get back to that point (first) before doing large changes. Small things that make concrete incremental improvements will make things more reviewable and more likely to land in the code quickly. Anything that requires that callers change how access them are much more difficult because most of the code using heketi apis are old-ish and unlikely to change quickly if at all.
We'd also need more tests as @raghavendra-talur commented on Apr 4, 2019.

ateleshev:
Okay, let's improve this step by step

phlogistonjohn:
Unfortunately, we do not think we have the resources to properly review and integrate a change of this extent. I truly appreciate your efforts to improve heketi but I do not think it is reasonable to leave this PR open, lingering, any more.
Thanks again.

PR #1564, opened 2019-03-27
Have the server refuse to start if db has unknown dbattributes

### LINKED ISSUE #1361: heketi should not start on encountering db with unknown attribute
### Kind of issue

Bug

### Observed behavior
Once a heketi database is upgraded/updated by a version of heketi which changes schema, it is possible to run heketi of lower version with that database.

### Expected/desired behavior
Heketi should verify in the DBATTRIBUTE bucket and see that it is able to interpret all the flags. If not, it should not start.

install heketi with version 6
update to version 7
run with version 6 again

phlogistonjohn:

Check that all the dbattributes in the db are known and if there is an
unknown attribute, which is probably from a too-new version of heketi,
refuse to start.
* Adds unit tests
* Adds functional test

### Does this PR fix issues?

Fixes #1361

### Notes for the reviewer

I have made this unconditional, but I'm open to making this test conditional. I didn't think we hit too many cases where we'd need to turn the check off though so I left it out. Any opinions?

### DIFF
--- a/apps/glusterfs/app.go
+++ b/apps/glusterfs/app.go
@@ -219,6 +219,14 @@ func (app *App) initDB() error {
 				return logger.LogError("Unable to initialize buckets: %v", err)
 			}
 
+			// Check that this is db we can safely use
+			validAttributes := validDbAttributeKeys(tx, mapDbAtrributeKeys())
+			if !validAttributes {
+				return logger.LogError(
+					"Unable to initialize db, unknown attributes are present" +
+						" (db from a newer version of heketi?)")
+			}
+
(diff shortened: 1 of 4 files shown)


apps/glusterfs/app.go
```diff
@@ -228,7 +236,7 @@ func (app *App) initDB() error {
 			return nil
 		})
 	}
-	return nil
+	return err
```
raghavendra-talur: Woah! I did not know about this bug, thanks for catching it.

raghavendra-talur:
@centos-ci retest this please

PR #1567, opened 2019-03-29
adding prom metrics for operations

rmartinez3:
Adding prom metrics for stale operations,in flight, failed, new and total.
Useful to expose more metrics and to setup alerts when a stale operation occurs.


pkg/metrics/metrics.go
```diff
@@ -133,6 +163,11 @@ func (m *Metrics) Describe(ch chan<- *prometheus.Desc) {
 	ch <- deviceFreeInBytes
 	ch <- deviceUsedInBytes
 	ch <- brickCount
+	ch <- staleCount
```
phlogistonjohn: It would be nice to have a comment here delineating the "sub sections" of items.

pkg/metrics/metrics.go
```diff
@@ -158,6 +193,37 @@ func (m *Metrics) Collect(ch chan<- prometheus.Metric) {
 		prometheus.GaugeValue,
 		float64(len(topinfo.ClusterList)),
 	)
+
+	opinfo, err := m.app.AppOperationsInfo()
+	if err != nil {
+		log.Println("Can't collect Opertions info for metrics: " + err.Error())
```
phlogistonjohn: "Opertions" -> "Operations"

pkg/metrics/metrics.go
rmartinez3: Added comment

pkg/metrics/metrics.go
rmartinez3: Fixed typo

rmartinez3:
Fixed Tests. Let me know if anything else is needed.

phlogistonjohn:
@rmartinez3 changes look good thanks! Could you please squash all your changes into a single commit and then I think I will be able to approve the PR. Thank you!

PR #1568, opened 2019-03-29
robust device handling

### LINKED ISSUE #1371: Identify block devices by a stable UUID, not boot dependent /dev/sdX
### Kind of issue

Feature request

### Observed behavior

Some cloud providers do not provide stable paths to disks. The ordering of the detection of disks may change when a Virtual Machine reboots.

### Expected/desired behavior

Heketi should be able to handle disks that are in a different order after initial topology import.

When disks get added, they are referenced by their device node (`/dev/sdb`). This is practical because users should know which disks are currently connected and dedicated for Heketi's use.

However, the device node gets stored in the database, and this does not need to stay correct. When a VM (or even a physical server) reboots, `/dev/sdb` does not need to be `/dev/sdb` again. The name of the device-node may change, it could be `/dev/sde` depending on the number of disks attached to the system.

This is not a problem for the tools that are used. Normally `/etc/fstab` contains UUID's of the formatted filesystems or LVM devices.  LVM writes its own header to the disks, and detects the layout of physical volumes, volume groups and logical volumes that way.

In order to have Heketi display the correct device-node after a reboot, the database should not contain (or trust) the device-node that is stored. Instead of storing the device-node, it is better to store the UUID that `pvcreate` writes in the LVM-header on the disk. This UUID can also be used through `/dev/disk/by-uuid/...`.

phlogistonjohn:

Add the ability to operate on persistent IDs and track alternative paths for devices. The original name for the device becomes largely informational after this pr, only to serve as a path fallback if the db entry was created in an older version of heketi and mapping a path by vg id fails.

The topology info and device info outputs are change to de-emphasize the "Name" field as that will always track the input device (to retain backwards compatibility in the API). 

The patches include changes that implement the ability to fetch and store metadata that will better identify a device. New metadata is stored in the db and a dbattribute flag indicating this has been added. Code to handle older devices without the metadata, but taking advantage of some of the new code to handle device removal even if device has changed /dev paths.

Unit tests as well as functional tests have been added.

### Does this PR fix issues?

Fixes #1371


mykaul:
Any updates?

nixpanic:
I've checked the execution and parsing of the output of the `pvs` and `udevadm` commands. That all looks good to me. The other changes look fine, but I do not feel qualified enough to approve those.

phlogistonjohn:
Previous centos ci runs succeeded, latest failures have nothing to do with the changes in this PR. Making an executive decision and merging this now.

PR #1570, opened 2019-04-01
apps: add a new multi-host-error type for handling tries over many hosts

phlogistonjohn:

Instead of hiding the errors returned by various hosts in the
try-on-hosts loop, capture them all in a new multi-hosts-error type.
This type works like the multi-cluster-error type. Additionally,
if all hosts return the same error it will "compact" the error message.

### Does this PR fix issues?

Fixes #


apps/glusterfs/multi_host_error.go
```diff
@@ -0,0 +1,94 @@
...
+	return m
+}
+
+// Error returns the error string for the multi host error.
+// If only one error was captured, it returns the text of that
+// error alone. If more than one error was captured, it returns
+// formatted text containing all captured errors.
+func (m *MultiHostError) Error() string {
+	if len(m.errors) == 0 {
+		return "(missing host error)"
```
raghavendra-talur: Shouldn't this be nil to conform with how error type works?

apps/glusterfs/try_on_host_test.go
```diff
@@ -168,3 +169,37 @@ func TestTryOnHosts(t *testing.T) {
...
+	})
+	t.Run("allErrorDifferent", func(t *testing.T) {
+		err := newTryOnHosts(hosts).run(func(h string) error {
+			return fmt.Errorf("%v is sad", h)
+		})
+		tests.Assert(t, err != nil, "expected err != nil, got:", err)
+		tests.Assert(t, strings.Count(err.Error(), "sad") == 3,
+			"expected 3 instances of 'sad' in", err)
+	})
+}
```
raghavendra-talur: Let us also add a test case for no error case.

apps/glusterfs/multi_host_error.go
phlogistonjohn: The issue with returning nil (actually it'd have to empty string here) is that you've already called the Error() function of the MultiHostError type. Because it's not a simple there-or-not-there case you can have an instance of the type created but w/o tracking any inner errors. This case basically exists to tell the programmer something went wrong with the use of the type. See the `Shorten` function for what actually should be used to extract a returnable error from the type.

apps/glusterfs/multi_host_error.go
```diff
@@ -0,0 +1,94 @@
...
+
+type MultiHostError struct {
+	prefix string
+	errors map[string]error
+}
+
+// NewMultiHostError returns a MultiHostError with the given
+// prefix text. Prefix text will be used in the error string if
+// more than one error is captured.
+func NewMultiHostError(p string) *MultiHostError {
```
ansiwen: You could instead require an initial error or a non-empty error list as an argument to avoid an error with an empty list.

phlogistonjohn:
@raghavendra-talur updated. PTAL

raghavendra-talur:
@ansiwen PTAL.

raghavendra-talur:
@centos-ci retest this please

PR #1571, opened 2019-04-02
docs: add a tag-matching design document

phlogistonjohn:

This is a design doc for a proposed enhancement to heketi for a tag-matching mechanism allows user specified filtering of devices to implement tiering, etc.

### DIFF
--- a/docs/design/tag-matching.md
+++ b/docs/design/tag-matching.md
@@ -1,0 +1,111 @@
+
+# Support for "Tag Matching" when placing bricks
+
+
+## Use case
+
+Users frequently request creating different performance tiers or service
+levels using Heketi managed Gluster. Currently, this is not possible
+unless the devices are split across different nodes such that nodes with
+device types X form one Gluster TSP and nodes with device type Y form
+a different Gluster TSP. Then Heketi can be requested to allocate volumes
+from one cluster or another. If one wants to deploy gluster on bare-metal
+with a mix of device types this workaround fails.
+
+## Summary
+
+This design proposes a simple "tag matching" approach to require Heketi
+only use certain devices when placing a volume. To keep the design simple
+no sort of auto-detection will be performed. The user is required to
+specify a key-value pair on a specific set of devices. Then a new
+volume option with a "user.heketi.device-tag-match" key and a simple matching
+rule (see below) must be specified when the volume is created.
+
+Then Heketi will ensure that bricks for this volume will only be
+placed on devices where the tag matching rule applies.
+
+This approach largely re-uses existing mechanisms established during the
+development of the arbiter feature as well as the zone-checking
+features.
+
+## Workflow Example
+
+Assume three nodes each with three existing devices each. One of the
+devices on the node is a smaller high-performance device. The
+remaining two devices are large slower devices.
+
+For each device on one example node:
+```
+   $ heketi-cli device settags abc1230901344aef90 tier:gold
+   $ heketi-cli device settags bbc1230901344aef91 tier:silver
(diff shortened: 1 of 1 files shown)


raghavendra-talur:
@phlogistonjohn This is awesome! To be able to get this feature based on many underlying parts that were done for Arbiter and Zone matching feature is great.

docs/design/tag-matching.md
```diff
@@ -0,0 +1,72 @@
...
+   $ heketi-cli device settags abc1230901344aef90 tier:gold
+   $ heketi-cli device settags bbc1230901344aef91 tier:silver
+   $ heketi-cli device settags cbc1230901344aef92 tier:silver
+```
+
+Now that devices have been tagged a volume can be requested to
+make use of the tagging:
+```
+    $ heketi-cli volume create --size=5 --gluster-volume-options 'user.heketi.tag-match tier=gold'
+    $ heketi-cli volume create --size=100 --gluster-volume-options 'user.heketi.tag-match tier=silver'
```
raghavendra-talur: I am thinking we need to have device in the key for example `user.heketi.device-tag-match` because we might later want to match only some nodes with more CPU etc(we already have for zone).

phlogistonjohn: OK, sounds good to me.

raghavendra-talur:
@phlogistonjohn Also, I would like to think of a way to protect some devices(like ssd) from being used in initial deployment.

For example:
gk-deploy OR openshift-ansible would provide the topology file and tag devices. It would also immediately ask for a volume for heketidbstorage. Admin might not want the default volumes to use "gold" tier. 

We have arbiter only tags for such use case in arbiter feature. I think we need something similar here too.

phlogistonjohn:
> @phlogistonjohn Also, I would like to think of a way to protect some devices(like ssd) from being used in initial deployment.
> 
> For example:
> gk-deploy OR openshift-ansible would provide the topology file and tag devices. It would also immediately ask for a volume for heketidbstorage. Admin might not want the default volumes to use "gold" tier.
> 
> We have arbiter only tags for such use case in arbiter feature. I think we need something similar here too.

Heketi today has the ability to specify the tags in the topology file. Combine that with recent work to support creating heketidbstorage with volume options and I think this is mainly verifying the pieces work together correctly. I really do not want to create a special case for things like 'heketidbstorage' even though they are common because this feature should be more about mechanism than policy IMO. I'd rather work with those projects to make sure they adapt to new heketi features than 2nd guess what they are doing.

phlogistonjohn:
I think instead of specifying a special tag, we just require users who don't want certain volumes used by initial setup processes to esablish a pattern of tags like so:
The software such as gk-deploy runs `./heketi-cli  setup-openshift-heketi-storage --gluster-volume-options 'user.heketi.tag-match builtinvols!=disabled'`
Then the user is requied to tag devices as builtinvols:disabled in the topology file ahead of time so that the set up process will skip over such devices.

raghavendra-talur:
I think I did not explain well. Here is what I intended

# Cases where some devices are to be used only when required
If admin wants to have some special devices like ssd that ought to be used only for certain volumes, it can be achieved by adding the user.heketi.tag-match key with a value that does not match such devices in  `pre_request_volume_options` in heketi.json.

phlogistonjohn:
I've changed the name of the volume option to "user.heketi.device-tag-match"

I've added sections to cover considerations related to but outside the immediate deign of the feature as we discussed.

phlogistonjohn:
OK. It actually helps to push the change one is talking about.

raghavendra-talur:
@phlogistonjohn Thanks for update. The design document looks good to me.

PR #1575, opened 2019-04-10
Use temporary directory for db extraction

### LINKED ISSUE #1574: Heketi can't start in kubernetes environment
### Kind of issue

Bug

### Observed behavior

Heketi can't start in kubernetes environment when used `HEKETI_BACKUP_DB_TO_KUBE_SECRET=1` and heketi db size more than 1Mb (actually, heketi.db+heketi.db.gz > 1Mb). Because the maximum size of a kubernetes secret 1Mb and heketi uses secret as temporary directory at startup to extract db from backup.

### Expected/desired behavior

Heketi should start normally (when packed db is less than 1Mb).

Setup heketi to kubernetes, use `HEKETI_BACKUP_DB_TO_KUBE_SECRET=1` in settings. 
Make heketi.db more than 1Mb when it unpacked.
Restart container.

- Kubernetes v1.9.3
- Heketi image: heketi/heketi:9

### Other useful information

Heketi start log:
```
Setting up heketi database
  File: /var/lib/heketi/heketi.db
  Size: 1052672         Blocks: 2056       IO Block: 4096   regular file
Device: ca02h/51714d    Inode: 1065043     Links: 1
Access: (0644/-rw-r--r--)  Uid: (    0/    root)   Gid: (    0/    root)
Access: 2019-03-29 08:18:37.841258993 +0000
Modify: 2019-04-09 19:28:12.601967081 +0000
Change: 2019-04-09 19:28:48.213952070 +0000
 Birth: -
/usr/bin/heketi-start.sh: line 75: /backupdb/heketi.db: Read-only file system
error: Unable to extract backup database
```

grig-tar:

This PR fix issue with start the heketi in kubernetes environment when uses backup db to kube secret and sum of sizes packed db and unpacked db is more than 1MiB.
[Kubernetes secrets are limited to 1MiB in size.](https://kubernetes.io/docs/concepts/configuration/secret/#restrictions)

### Does this PR fix issues?

Fixes #1574

### Notes for the reviewer

Line 80 was here only for compatibility with versions prior to 5.0 when db compression was introduced during backup.

### DIFF
--- a/extras/container/heketi-start.sh
+++ b/extras/container/heketi-start.sh
@@ -6,6 +6,7 @@
 
 : "${HEKETI_PATH:=/var/lib/heketi}"
 : "${BACKUPDB_PATH:=/backupdb}"
+: "${TMP_PATH:=/tmp}"
 LOG="${HEKETI_PATH}/container.log"
 
 info() {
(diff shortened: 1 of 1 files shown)


grig-tar:
Updated to use variable with default value.

grig-tar:
Squashed and rebased on actual master.

grig-tar:
Looks like I had wrong with size limit of secret.
Heketi will never start if there will be mounted secret with gzipped backup of db and version of kubernetes >=1.9.4.
From v1.9.4, secrets volumes are now mounted in read-only by default.

https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.9.md#v194 :
>Changes secret, configMap, downwardAPI and projected volumes to mount read-only, instead of allowing applications to write data and then reverting it automatically. Until version 1.11, setting the feature gate ReadOnlyAPIDataVolumes=false will preserve the old behavior. (#58720, @joelsmith)

However, using a temporary directory is a good way to fix the issue.

PR #1599, opened 2019-05-30
Default to authenticated access [behavior change]

phlogistonjohn:

The heketi server will fail to start if authentication parameters
are not supplied in the configuration. The only way to disable
authentication is to pass the --disable-auth option as a server
command line argument.

Note: This is a significant behavior change.

This is being done to ensure more secure deployments by default. Instead of having to opt-in to authentication, Heketi now requires one to opt out of authentication. Generally, there is not much need to even opt out of authentication (just use auth with a well-known "secret") except for testing and backwards compatibility for older clients where auth has not been configured.

### Notes for the reviewer

If this PR is accepted I plan on mailing the list as well as making a change to the README so visitors to the github page see a highlighted note.

Last changes still need to be squashed, was keeping them seperate for smoke testing purposes.

### DIFF
--- a/main.go
+++ b/main.go
@@ -44,6 +44,7 @@ var (
 	deleteAllBricksWithEmptyPath bool
 	dryRun                       bool
 	force                        bool
+	disableAuth                  bool
 )
 
 var RootCmd = &cobra.Command{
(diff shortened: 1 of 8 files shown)


phlogistonjohn:
Updated patches and added proper functional test. Please re-review @raghavendra-talur

phlogistonjohn:
Ping @raghavendra-talur patches are now fully ready for review.

phlogistonjohn:
retest this please

PR #1614, opened 2019-06-27
work around broken api use in glide

phlogistonjohn:

No we really shouldn't be manually editing glide.lock but this fixes
the use of a removed bitbucket api by the apparently unmaintained
glide.

### Does this PR fix issues?

Fixes #

### Notes for the reviewer

Its probably time to start thinking about moving away from glide. Seriously that is.


raghavendra-talur:
> If this works then I think we need a change in the glide.yaml file too but without running the update command.

I take this back. We don't need any change in yaml file.

phlogistonjohn:
Just to be clear, this is very much not the "right way" to use glide, but bitbucket had deprecated the api glide is using over a year ago as well as the last commit in glide occurred roughly 9 months ago (at the time of this writing). I want things to break if people start trying to use glide correctly. :-)

PR #1616, opened 2019-06-27
add client installation command for epel

mojighahar:
I don't find anywhere that mention installation of client and fall into problem.

### DIFF
--- a/docs/admin/install-standalone.md
+++ b/docs/admin/install-standalone.md
@@ -14,7 +14,10 @@ To install Heketi on an EPEL/Fedora please type the following:
 * EPEL
 
 ```
+Server:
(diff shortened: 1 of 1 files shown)


phlogistonjohn:
centos ci not needed due to doc only change.

phlogistonjohn:
@raghavendra-talur PTAL

nixpanic:
The heketi(-client) package is not part of EPEL. Instead this can be installed from the CentOS Storage SIG:
```
# yum install centos-release-gluster
# yum install heketi-client
```

The packages are part of Fedora, so that part of the documentation is still correct (but use `dnf` instead of `yum`).

PR #1617, opened 2019-07-03
Use metadata, not device names, to allow or deny new devices

phlogistonjohn:

It was found that the changes made in PR #1568 were incomplete and that while you could remove a device that had moved from /dev/sdb to /dev/sdd if a new device was added and it took the path of an existing device it could not be added (by /dev/sdx path).

It was also determined that simply removing the device "registration" feature that prevented the adding previously used /dev/ paths could lead to a dangerous situation if the topology flie had an error like listing /dev/sdd twice for the same node and "destroy data" was set. In this case a naive approach would allow the 2nd attempt to add /dev/sdd to wipe and re-use a device that this instance of heketi was already using.

Therefore, this change attempts to both remove the old "registration" of devices while preventing the accidental re-use of a device the current heketi instance is managing. We do this by adding error types that are informational to the device initialization process. If simple no-wipe initialization fails and the destroy-data flag is set, the richer error type will be checked that the device was a) successfully checked b) has lvm pv metadata or not c) if that lvm pv is in use by this heketi.

### Does this PR fix issues?

Fixes #


phlogistonjohn:
retest this please

PR #1621, opened 2019-07-05
Add a NodeSelector to heketi-storage-copy-job

arnoSCC:

Hi!

Installing Heketi through gluster-kubernetes gk-deploy on a mixed linux/windows 1.14.3 K8S cluster, the heketi-storage-copy-job failed due to being scheduled on a windows node. 

All other templates from gluster-kubernetes could be manually patched to achieve node selection, but the heketi-storage-copy-job is defined in heketi-cli.

I just added the NodeSelector in the Job.Spec.Template.Spec, compiled (with Go 1.12.6), replaced the heketi-cli binary in the heketi/heketi:latest image and was able to deploy with success after (and the job had the nodeSelector as intendend.

### Does this PR fix issues?

No open issue fixed.

### Notes for the reviewer

I used the beta.kubernetes.io/os label instead of kubernetes.io/os for compatibility with older K8S


arnoSCC:
Forgot to gofmt the file modified, commit is amended.

raghavendra-talur:
@centos-ci ok to test

raghavendra-talur:
@arnoSCC Thanks for the contribution. The change itself looks good. However, I would like inputs from @phlogistonjohn on the general approach for this problem.

In my research I found this [guide](https://kubernetes.io/docs/setup/production-environment/windows/user-guide-windows-containers/#taints-and-tolerations) which states

> However, we understand that in many cases users have a pre-existing large number of deployments for Linux containers, as well as an ecosystem of off-the-shelf configurations, such as community Helm charts, and programmatic Pod generation cases, such as with Operators. In those situations, you may be hesitant to make the configuration change to add nodeSelectors. The alternative is to use Taints. Because the kubelet can set Taints during registration, it could easily be modified to automatically add a taint when running on Windows only.

I think the copy job falls in the same domain as Helm charts and programmatic pod generation. Hence, the first approach of the users should probably be to use the taint mechanism. 

However, I don't see any side effect of limiting heketi copy job to linux and we can take this change after verifying it does not break any existing setups.

arnoSCC:
@raghavendra-talur Thanks for your answer!

nodeSelector are a best practice but yes, taint approach is valid too. If this PR is deemed harmful, i'll make my customer try this instead of maintaining a fork for their use case.

client/cli/go/cmds/heketi_storage.go
```diff
@@ -246,6 +246,9 @@ func createHeketiCopyJob(volume *api.VolumeInfoResponse) *batch.Job {
 	job.Spec.Parallelism = &p
 	job.Spec.Completions = &c
 	job.Spec.Template.ObjectMeta.Name = HeketiStorageJobName
+	job.Spec.Template.Spec.NodeSelector = map[string]string{
+		"beta.kubernetes.io/os": "linux",
```
phlogistonjohn: Instead of hard-coding a particular selector, I think the better approach would be to add a new command line argument that allows specifying one or more selector and then this can be passed into the Spec like it is here.
Then for your use case you can specify linux os but another user could specify some other custom selector they'd find useful.

arnoSCC:
Being more of a python guy than go, I did the modification as cleanly as possible @phlogistonjohn. Hope it fit what was in your mind.

I was able to modify the gk-deploy script to use the newly added --node-selector flag for setup-openshift-heketi-storage with success.

phlogistonjohn:
retest this please

phlogistonjohn:
retest this please

phlogistonjohn:
(sorry! hit the other button accidentally)

phlogistonjohn:
the vagrant commands in centosci are having issues again. I'll poke the ci again later.

arnoSCC:
Can a retest be triggered if the centos-ci job is fixed?

arnoSCC:
And that's a wrap for the tests :D

Hope this will get merged.

phlogistonjohn:
Hope achieved. :-)
Thanks for the contribution!

arnoSCC:
Perfect thanks!

PR #1622, opened 2019-07-05
tools: print glusterblock volume name for blockvolumes

raghavendra-talur:
Earlier, the error log was referring to gluster volume name from path in
all cases.

### DIFF
--- a/extras/tools/comparison.py
+++ b/extras/tools/comparison.py
@@ -217,10 +217,6 @@ def compile_pvdata(summary, pvdata, matchsc):
         if not g and 'glusterBlockShare' not in ma:
             continue
         sc = elem.get('spec', {}).get('storageClassName', '')
-        if matchsc and sc not in matchsc:
-            sys.stderr.write(
-                'ignoring: {} from storage class "{}"\n'.format(g["path"], sc))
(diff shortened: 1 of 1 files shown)


extras/tools/comparison.py
```diff
@@ -212,7 +212,7 @@ def compile_pvdata(summary, pvdata, matchsc):
         sc = elem.get('spec', {}).get('storageClassName', '')
         if matchsc and sc not in matchsc:
             sys.stderr.write(
-                'ignoring: {} from storage class "{}"\n'.format(g["path"], sc))
+                'ignoring: {} from storage class "{}"\n'.format(g["path"] if g else ma['glusterBlockShare'], sc))
```
phlogistonjohn: We probably need to catch KeyError exceptions (or use .get) instead of using an if statement here. And if we are adding fallbacks we should fall back to something really expected like name of the pv.

phlogistonjohn:
@raghavendra-talur ping. Do you need me to clarify my request at all?

raghavendra-talur:
@phlogistonjohn Used a different approach, have a look. Will need your help for for the approach you mentioned.

PR #1626, opened 2019-07-09
provide an endpoint patch for volumes

raghavendra-talur:
In Kubernetes and OpenShift, the gluster volume provisioner parses the volume information provided by heketi during volume creation to create the endpoint that is used for mounting the volume.

Some heketi operations, like node or device replace can change the cluster information such that the IPs listed in the endpoints become outdated. In this patch, we provide a command that outputs the current mount information for the volume and can be used to patch an existing endpoint resource.

### DIFF
--- a/apps/glusterfs/volume_entry.go
+++ b/apps/glusterfs/volume_entry.go
@@ -237,7 +237,9 @@ func (v *VolumeEntry) NewInfoResponse(tx *bolt.Tx) (*api.VolumeInfoResponse, err
 	info := api.NewVolumeInfoResponse()
 	info.Id = v.Info.Id
 	info.Cluster = v.Info.Cluster
-	info.Mount = v.Info.Mount
+	if err := v.updateMountInfo(wdb.WrapTx(tx), &info.VolumeInfo); err != nil {
+		return nil, err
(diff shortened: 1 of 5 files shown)


humblec:
@raghavendra-talur can you update the PR with some more details ?

apps/glusterfs/volume_entry.go
```diff
@@ -237,7 +237,10 @@ func (v *VolumeEntry) NewInfoResponse(tx *bolt.Tx) (*api.VolumeInfoResponse, err
 	info := api.NewVolumeInfoResponse()
 	info.Id = v.Info.Id
 	info.Cluster = v.Info.Cluster
-	info.Mount = v.Info.Mount
+	if err := v.updateMountInfo(wdb.WrapTx(tx), &info.VolumeInfo); err != nil {
+		return nil, err
+	}
+	//info.Mount = v.Info.Mount
```
phlogistonjohn: Stray commented out code. This can be safely removed.

raghavendra-talur:
> I approved because overall looks good but I would like the stray comment removed and that's pretty trivial.

Done.

PR #1639, opened 2019-08-23
all: fix goimports errors

raghavendra-talur:
There are 3 types of errors fixed:
a. one empty line is required before where third-party pkgs imports
b. packages can have dashes in import path but won't have in name[1],
explicitly name them on import.
c. packages that are just version numbers like "v1" need explicit naming
too[2].

[1] https://github.com/golang/go/issues/17546
[2] https://github.com/golang/go/issues/29041

### DIFF
--- a/pkg/kubernetes/backupdb.go
+++ b/pkg/kubernetes/backupdb.go
@@ -20,7 +20,7 @@ import (
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	restclient "k8s.io/client-go/rest"
-	"k8s.io/kubernetes/pkg/api/v1"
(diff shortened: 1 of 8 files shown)


pkg/kubernetes/backupdb.go
```diff
@@ -20,7 +20,7 @@ import (
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	restclient "k8s.io/client-go/rest"
-	"k8s.io/kubernetes/pkg/api/v1"
+	v1 "k8s.io/kubernetes/pkg/api/v1"
```
phlogistonjohn: I'd prefer better naming scheme than this, but that shouldn't block this pr.

PR #1640, opened 2019-08-23
client: turn off perf xlators by default for heketidbstorage

### LINKED ISSUE #1591: Disable performance translators for heketidbstorage
### Kind of issue

Feature request

### Observed behavior

Periodically we get reports of corrupt boltdb files in the wild. This generally shouldn't happen as bolt is a fairly simple db format and has been stable (ossified if you prefer) for a while now. One area of risk that concerns us (speaking for myself and @raghavendra-talur) is that there could be some issue in the common configuration where the boltdb file resides on a glusterfs volume. In order to reduce the risk and still retain the benefits of keeping a db on gluster we are considering disabling glusterfs (performance) translators for the heketidbstorage volume.
This may be optional, design pending.

raghavendra-talur:
### Does this PR fix issues?

Fixes #1591


raghavendra-talur:
Closing. Updated by #1642

PR #1642, opened 2019-08-30
Use "safer" volume options on heketidbstorage

phlogistonjohn:

This PR starts off by incorporating changes from PR #1640 

New instances of heketidbstorage volume will be created with options to turn of so called perf xlators.

A new "offline" subcommand "update-dbvol" is provided to conveniently update existing heketidbstorage volumes to the new set of options. Note that this *intentionally* doesn't update the options in the db. Currently this is the only tool that changes the volume options, normally heketi only sets volume options at create time. Syncing the volume options from gluster to the db is a topic reserved for future work.

A set of "error handling tests" to check the behavior of the new command.

### Does this PR fix issues?

Fixes #

### DIFF
--- a/client/cli/go/cmds/heketi_storage.go
+++ b/client/cli/go/cmds/heketi_storage.go
@@ -47,6 +47,7 @@ var (
 	heketiStorageDurability   string
 	heketiStorageReplicaCount int
 	heketiStorageOptions      string
+	appendStorageOptions      string
 	HeketiStorageJobSelector  string
 )
 
@@ -75,6 +76,12 @@ func init() {
 		"gluster-volume-options",
 		"",
 		"\n\tOptional: Comma separated list of volume options."+
+			"\n\tSee volume create --help for details.")
+	setupHeketiStorageCommand.Flags().StringVar(&appendStorageOptions,
+		"append-volume-options",
+		"",
+		"\n\tOptional: Comma separated list of volume options to be added"+
+			"\n\tto the existing/default set of options."+
 			"\n\tSee volume create --help for details.")
 	setupHeketiStorageCommand.Flags().StringVar(&HeketiStorageJobSelector,
 		"node-selector",
@@ -150,6 +157,12 @@ func createHeketiStorageVolume(c *client.Client, dt api.DurabilityType, replicaC
 		req.GlusterVolumeOptions = strings.Split(heketiStorageOptions, ",")
 		req.GlusterVolumeOptions = append(
 			req.GlusterVolumeOptions, "user.heketi.dbstoragelevel custom")
+	}
+
+	if appendStorageOptions != "" {
+		req.GlusterVolumeOptions = append(
+			req.GlusterVolumeOptions,
+			strings.Split(appendStorageOptions, ",")...)
 	}
 
 	switch dt {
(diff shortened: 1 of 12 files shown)


client/cli/go/cmds/heketi_storage.go
```diff
@@ -152,6 +159,12 @@ func createHeketiStorageVolume(c *client.Client, dt api.DurabilityType, replicaC
 			req.GlusterVolumeOptions, "user.heketi.dbstoragelevel custom")
 	}
 
+	if appendStorageOptions == "" {
+		req.GlusterVolumeOptions = append(
+			req.GlusterVolumeOptions,
+			strings.Split(appendStorageOptions, ",")...)
```
raghavendra-talur: I think we need to set the storage level to custom even in append mode.

phlogistonjohn: What I was trying to do is provide the user with two different semantics, the first and pre existing option is "here are my volume options" this clearly calls for a "custom" level because the user is now overriding the system's default values. The new option I was trying to give the caller a way to say "on top of the defaults do X". Obviously there's a bit of confusion if you supply one of the options that the new defaults include (with a different value). I had hoped to keep the code simple with the idea that even if the user had done that they were accepting that in the future if they didn't override the default options they were opting in to having those "defaults" changed.
A fancier way to do this would be to look thru both lists and if an option in the append list appeared in the defaults list it would warn and or set the options to custom but I'm still not sure I like that semantically and its certainly way more complicated.
I do not think I want to change the semantics of this new option. So I think we either need to drop it from this PR entirely or leave it as is. Thoughts?

phlogistonjohn: If we keep the option as-is I can add a comment to the code explaining the semantics as I did here.

raghavendra-talur: IMO, any user input should be considered as custom storage level. 

Of the three options
a. heketi defaults
b. user options 
c. user options appended to defaults 

I would prefer to drop "b. user options" as (c) can be used to override the defaults if required.
In that approach, I had https://github.com/raghavendra-talur/heketi/commit/0fe666d0edcbd9b45e47938231c135c472d741b1#diff-7c36e16fac4d1fd7b56c539699be42dfR151 where in the append case all default options are taken but the last option which is storage level.

If you think that is confusing, we can drop the append option.

phlogistonjohn: I disagree and I think it might come down how we see the "level" setting. I don't see the "custom" level as saying the user set some customizations. It's "not safe for heketi to automatically change". Obviously, if the user uses the append option they're applying customizations but they are customizations that they're OK with the system changing in the future (if needed). I see stuff like the arbiter option belonging there versus these perf xlator changes that are part of the new defaults.

raghavendra-talur: Ok, that clears the confusion about what "custom" means. I think we can leave the code as it is in this PR then.

phlogistonjohn:
Somehow I totally missed giving the last commit a proper commit message. It is fixed now.

client/cli/go/cmds/heketi_storage.go
```diff
@@ -152,6 +159,12 @@ func createHeketiStorageVolume(c *client.Client, dt api.DurabilityType, replicaC
 			req.GlusterVolumeOptions, "user.heketi.dbstoragelevel custom")
 	}
 
+	if appendStorageOptions == "" {
```
raghavendra-talur: Is this right or should it be "!=" ?

phlogistonjohn: That's correct. Will fix

PR #1650, opened 2019-09-16
execute LVM commands on the host instead of in a Gluster container

nixpanic:

There are a few potential issues when running LVM commands inside a Gluster container. This can be solved by running the commands on the container host, instead of in the container. The `nsenter` command can be used to 'break out' of the container, and execute the command on the host.

### Does this PR fix issues?

* difficulties with udev synchonisation (file locking under `/run` and IPC semaphores)
* differences/incompatibilities between LVM versions on the host and in the container

### Notes for the reviewer

This PR currently is a RFC to get feedback on the approach taken. Tests have been done, and the needed functionality seems to be present.

### DIFF
--- a/docs/admin/server.md
+++ b/docs/admin/server.md
@@ -28,6 +28,7 @@ The config file has information required to run the Heketi server.  The config f
         * backup_lvm_metadata: _bool_, Create archives of the LVM metadata when running vgcreate/lvcreate
         * sudo: _bool_, set to true when SSHing as a non root user
 	* debug_umount_failures: _bool_, Enable to capture more details in case brick unmounting fails. Can be overridden by the HEKETI_DEBUG_UMOUNT_FAILURES environment variable.
+	* lvm_wrapper: _string_, Use a wrapper for calling LVM related operations. Can be overridden by the HEKETI_LVM_WRAPPER environment variable.
     * kubexec: _map_, Kubernetes configuration
         * host: _string_, Kubernetes API host.  Example `https://myhost:8443`.  Can also be use using environment variable HEKETI_KUBE_APIHOST
         * cert: _string_, Certificate file to for HTTPS connection. Can also be use using environment variable HEKETI_KUBE_CERTFILE
@@ -38,6 +39,7 @@ The config file has information required to run the Heketi server.  The config f
         * fstab: _string_, Fstab file where to store mount points
         * backup_lvm_metadata: _bool_, Create archives of the LVM metadata when running vgcreate/lvcreate
 	* debug_umount_failures: _bool_, Enable to capture more details in case brick unmounting fails. Can be overridden by the HEKETI_DEBUG_UMOUNT_FAILURES environment variable.
+	* lvm_wrapper: _string_, Use a wrapper for calling LVM related operations. Can be overridden by the HEKETI_LVM_WRAPPER environment variable.
 
 ## Advanced Options
 The following configuration options should only be set on advanced configurations under `glusterfs` section:
--- a/executors/cmdexec/cmdexec.go
+++ b/executors/cmdexec/cmdexec.go
@@ -65,6 +65,14 @@ func (c *CmdExecutor) lvmCommand() string {
 	return "/usr/sbin/lvm"
 }
 
+func (c *CmdExecutor) udevCommand() string {
+	if wrapper := c.LVMWrapper(); wrapper != "" {
+		return wrapper + " /usr/bin/udevadm"
+	}
+
+	return "/usr/bin/udevadm"
+}
+
 func setWithEnvVariables(config *CmdConfig) {
 	var env string
 
--- a/executors/cmdexec/cmdexec.go
+++ b/executors/cmdexec/cmdexec.go
@@ -51,6 +51,20 @@ func (c *CmdExecutor) glusterCommand() string {
 	return fmt.Sprintf("gluster --mode=script --timeout=%v", c.GlusterCliTimeout())
 }
 
+// When running Gluster inside a container, the LVM commands need to be
+// executed on the host. The Gluster Server containers provide a wrapper script
+// for `nsenter` and its arguments as `exec-on-host`.
+//
+// All LVM commands will get prefixed with `/usr/sbin/lvm` in order to make
+// things a little more secure (hopefully).
+func (c *CmdExecutor) lvmCommand() string {
+	if wrapper := c.LVMWrapper(); wrapper != "" {
+		return wrapper + " /usr/sbin/lvm"
+	}
+
+	return "/usr/sbin/lvm"
+}
+
 func setWithEnvVariables(config *CmdConfig) {
 	var env string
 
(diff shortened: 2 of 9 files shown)


executors/cmdexec/cmdexec.go
```diff
@@ -23,10 +23,14 @@ import (
...
 	logger     = logging.NewLogger("[cmdexec]", logging.LEVEL_DEBUG)
 	preallocRe = regexp.MustCompile("^[a-zA-Z0-9-_]+$")
+
+	// escape priviledges to host
+	HostExec = "exec-on-host"
 )
 
 type RemoteCommandTransport interface {
 	ExecCommands(host string, commands rex.Cmds, timeoutMinutes int) (rex.Results, error)
+	ExecHostCommands(host string, commands rex.Cmds, timeoutMinutes int) (rex.Results, error)
```
phlogistonjohn: I think this is the wrong layer for this. ExecCommands runs a command on a node and it shouldn't care if that command "breaks out" of a container or not. In fact, this seems like the wrong place to decide if the command is to be run on something "containerized" or not.

phlogistonjohn:
@nixpanic thoughts? Do you need me to do a POC patch to demonstrate my preferred approach?

nixpanic:
On Mon, Sep 23, 2019 at 09:55:23AM -0700, John Mulligan wrote:
> @nixpanic thoughts? Do you need me to do a POC patch to demonstrate my preferred approach?

Thanks, I understand the approach you suggested. I hope to have time
this week to take that route. If there is no update from me in the next
few days, you might want to have a go at it yourself (I'll be on
holidays next week).

nixpanic:
These two commits run `exec-on-host` when kubeexec is the executor. The wrapper script is not yet part of the Gluster server container, so some detection should be done (also for older deployments). In case `exec-on-host true` fails, the old way of running lvm/udevadm commands inside the container should be used. This is not yet implemented.

executors/cmdexec/cmdexec.go
```diff
@@ -35,6 +35,7 @@ type RemoteCommandTransport interface {
 	LVChunkSize() string
 	XfsSw() int
 	XfsSu() int
+	IsContainerizedGluster() bool
```
phlogistonjohn: I think we should change this in favor of pulling the exact "wrapper command" string directly from the config. This way heketi does not have to be "smart" about containerized gluster or not, it would be part of the deployment and users could even create their own wrapper scripts if needed.

phlogistonjohn:
Travis is failing for issues with the PR as well.

nixpanic:
Hi @phlogistonjohn, is this more inline with what you were thinking? I'll need to update the commit messages a little more, but this approach should work well (untested as of now).

nixpanic:
And now with the test fixes included (hopefully) 🤞🏼

docs/admin/server.md
```diff
@@ -28,6 +28,7 @@ The config file has information required to run the Heketi server.  The config f
         * backup_lvm_metadata: _bool_, Create archives of the LVM metadata when running vgcreate/lvcreate
         * sudo: _bool_, set to true when SSHing as a non root user
 	* debug_umount_failures: _bool_, Enable to capture more details in case brick unmounting fails. Can be overridden by the HEKETI_DEBUG_UMOUNT_FAILURES environment variable.
+	* lvm_wrapper: _string_, Use a wrapper for calling LVM related operations. Can be overridden by th HEKETI_LVM_WRAPPER environment variable.
```
phlogistonjohn: Minor typo: "th" not "the"

docs/admin/server.md
```diff
@@ -38,6 +39,7 @@ The config file has information required to run the Heketi server.  The config f
         * fstab: _string_, Fstab file where to store mount points
         * backup_lvm_metadata: _bool_, Create archives of the LVM metadata when running vgcreate/lvcreate
 	* debug_umount_failures: _bool_, Enable to capture more details in case brick unmounting fails. Can be overridden by the HEKETI_DEBUG_UMOUNT_FAILURES environment variable.
+	* lvm_wrapper: _string_, Use a wrapper for calling LVM related operations. Can be overridden by th HEKETI_LVM_WRAPPER environment variable.
```
phlogistonjohn: Same minor typo

nixpanic:
On Wed, Oct 09, 2019 at 10:52:47AM -0700, John Mulligan wrote:
> I won't insist on it but would you be open to adding a specific test
> (probably under the so-called TestErrorHandling suite) to test the
> command execution / configuration of the prefixes? I'd happily help
> guide with writing the test too.

Not sure where to add that. Can you point out to an existing test that I
can use as a basis for it? 'ErrorHandling' only gives a few hits that do
not help me much further.

phlogistonjohn:
> Not sure where to add that. Can you point out to an existing test that I can use as a basis for it? 'ErrorHandling' only gives a few hits that do not help me much further.

Sure that was me checking in if you were open to it. Since you sound like you are, here goes:
In tests/functional/TestErrorHandling/tests/options_test.go TestBlockVolumeAllocDefaults might be a good example of how to toggle the new option

If you want to get really fancy you might look at TestErrorInjection in tests/functional/ErrorHandling/tests/error_inject_test.go to see if doing some kind of assertions on the command string using the regex in an inject config command hook to force a particular result if the new option is set vs unset.

Or you may even want to create a "dummy" wrapper script similar to how test/functional/vagrant/roles/gluster/tasks/main.yml installs some hook scripts for testing. Your script might just log and then pass the command thru when it was called so that the test case could assert the log got updated.

nixpanic:
@phlogistonjohn could you give me a pointer to what could cause the CI tests to fail? I only added one test, but others are failing too :confused:

phlogistonjohn:
> @phlogistonjohn could you give me a pointer to what could cause the CI tests to fail? I only added one test, but others are failing too confused

The issue appears to be unrelated to the new test but rather is caused by the new behavior of always prefixing lvm commands with /usr/sbin/lvm. The test TestErrorInjection uses regexes to key off of commands issued to the nodes and assumes the regex `^lvcreate.*` will match a command heketi will run on the node. With the new prefix this is no longer true. Something like `"(^| )lvcreate.*"` should improve the situation. I checked and this test is only one needing these updates.

phlogistonjohn:
While I was at it I also looked over the code again, things are looking pretty good now, including the test code. My last request after fixing the test regression previously mentioned is to fix the commit messages and then I'll OK the PR.

phlogistonjohn:
@ndevos I think I understand the cause of the CI failure. See my comment above. Please take a look when you have a moment.

nixpanic:
Thanks! That makes sense, I think I have been bitten by something like this before...
I just noticed that `sudo` will be removed from the container image soon, so I'll have the test run through `nsenter` instead (without arguments it stays in the same namespace).

phlogistonjohn:
If you mean the tests, the tests don't execute in containers... so I would be happy to leave sudo in for now. Longer term we could install a custom wrapper script but we can do that in a follow up pr.

nixpanic:
On Thu, Dec 05, 2019 at 05:49:46AM -0800, John Mulligan wrote:
> If you mean the tests, the tests don't execute in containers... so I would be happy to leave sudo in for now. Longer term we could install a custom wrapper script but we can do that in a follow up pr.

I replaced sudo with nsenter, which should have the same result. And the
added advantage is that is works in a container too (where sudo might
not be available).

If you really prefer sudo over nsenter, I'll change it back again.

phlogistonjohn:
It's fine as long it is on the vms... I have not confirmed manually, I will just allow the ci to run for now. :-)

phlogistonjohn:
Tests passing. Code LGTM.

PR #1655, opened 2019-09-24
Implement core "Tag Matching" feature for placing bricks on devices

phlogistonjohn:

As outlined in the [tag matching design doc](/heketi/heketi/blob/master/docs/design/tag-matching.md) an administrator can set up tags on nodes and devices and then when volumes are being created a tag matching rule can be specified such that the bricks that make up the volume are placed only on matching devices.

A simple use case would be devices managed by heketi with heterogeneous performance characteristics. An admin might tag some disks with "perf:fast" and others with "perf:slow". Then depending on the needs of the volume (eg. small/fast vs. large/slow) the volume can be requested with volume options like so "user.heketi.device-tag-match perf=fast" or  "user.heketi.device-tag-match perf=slow". Note that inverse matches are also supported, example "user.heketi.device-tag-match perf!=fast".
Note that this feature is generic and really has nothing to do with performance in particular as the tag keys, tag values, and match rules have no internal meaning to Heketi. Thus the matching can be used for any reason. We do suspect performance matching will be the most common.

### Does this PR fix issues?

Fixes #

### Notes for the reviewer

Currently the matching rule's "selector" string is either `=` or `!=`. I wondered if there could be issues here that could either cause confusion (`=` vs. `==`) or parsing issues if the options string has to be encoded in particular languages/formats (`!=` use of exclamation mark).
I was considering supporting alternate spellings of these but kept the initial PR on the simpler side (and matching the original proposal). Something we might want to discuss.

In addition, to keep this pr simple and self-contained the "Other Considerations" section was not considered at this time.

### DIFF
--- a/apps/glusterfs/tag_matching_rule_test.go
+++ b/apps/glusterfs/tag_matching_rule_test.go
@@ -72,3 +72,82 @@ func TestTagMatchingRuleTest(t *testing.T) {
 		tests.Assert(t, tmr.Test("blob"))
 	})
 }
+
+func TestTagMatchingRuleFilter(t *testing.T) {
+	dsrc := NewTestDeviceSource()
+	var (
+		n         *NodeEntry
+		d         *DeviceEntry
+		deviceAdd func(string, string) *DeviceEntry
+	)
+	n, deviceAdd = tmrNodeAdd(dsrc, "foo")
+	n.Info.Tags["charles"] = "brown"
+	d = deviceAdd("foo000d1", "/dev/d1")
+	d.Info.Tags["tier"] = "one"
+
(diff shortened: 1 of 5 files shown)


apps/glusterfs/tag_matching_rule_test.go
```diff
@@ -72,3 +72,82 @@ func TestTagMatchingRuleTest(t *testing.T) {
 		tests.Assert(t, tmr.Test("blob"))
 	})
 }
+
+func TestTagMatchingRuleFilter(t *testing.T) {
```
raghavendra-talur: If you care about keeping the name same s/Filter/GetFilter/g

PR #1659, opened 2019-10-07
Fixed Heketi database restore logic from a secret based backup

Cyreex:
Fixed Heketi database restore logic from a secret based backup during the startup phase

Currently we are always trying restore a backup from a secret, if it exists. But if we don't have a database file on the Heketi database volume, the container stops working with the error: "Database file did not appear, exiting."

Now we are trying to restore the Heketi database from a secret based backup if secret exists and the database volume doesn't have a database file.

### Notes for the reviewer

### DIFF
--- a/extras/container/heketi-start.sh
+++ b/extras/container/heketi-start.sh
@@ -116,19 +116,23 @@ fi
 
 if [[ ! -f "${HEKETI_PATH}/heketi.db" ]]; then
     info "No database file found"
+    info "Database file is expected, waiting..."
+    check=0
+    while [[ ! -f "${HEKETI_PATH}/heketi.db" ]]; do
+        sleep 5
+        if [[ ${check} -eq 5 ]]; then
+            #Try to restore BD from secret
+            restore_backup
+            break
+        fi
+        ((check+=1))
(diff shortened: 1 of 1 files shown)


hetii:
I would be more then happy but it's not clear for me how this "backup_db_to_kube_secret" even works. 
I know that here you describe the way how the db can be restored, but how it is populated in first place?
In my setup I have backup_db_to_kube_secret set to true as well as heketi-db-backup secret mounted as volume into /backupdb but it is empty all the time.
Btw I don't use gk-deploy as I have glusterfs in separate nodes that act as a kubernetes nodes.

Cyreex:
In the current version of the script, we always restore the backup from a secret, if it exists:

```
if [[ -d "${BACKUPDB_PATH}" ]]; then	
    if [[ -f "${BACKUPDB_PATH}/heketi.db.gz" ]] ; then
...
```
I think this behavior isn't correct. We must restore the database only if we need it.

phlogistonjohn:
We appreciate your intent to improve Heketi, but I think we need to decline this contribution. Thank you for your efforts.

PR #1673, opened 2019-11-06
fix inconsistent db after (failing) volume expand operation

phlogistonjohn:

It was found that a volume expand operation that fails due to errors in the exec phase would leave a trivial setup inconsistent with regards to the sizes of devices. This PR fixes that issue as well as some ancillary items found while working on this problem.

### Does this PR fix issues?

Fixes rhbz#1653567

### DIFF
--- a/apps/glusterfs/operations_volume.go
+++ b/apps/glusterfs/operations_volume.go
@@ -356,7 +356,7 @@ func (ve *VolumeExpandOperation) CleanDone() error {
 		}
 		bricks, err := bricksFromOp(txdb, ve.op, v.Info.Gid)
 		for _, brick := range bricks {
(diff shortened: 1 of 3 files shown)


raghavendra-talur:
@phlogistonjohn The default merge method was changed to "squash and merge" instead of fast-forward and I hit the button out of habit before reading carefully. Let me know if you would want to revert this merge and have separate commits.

phlogistonjohn:
Its fine but please make sure you change the default method back in the github config.

phlogistonjohn:
Also, this reminds me why we shouldn't make exceptions to normal processes. ;-)

raghavendra-talur:
Yup, changed the defaults back.

PR #1682, opened 2019-12-09
fix test scripts for default testenv teardown

raghavendra-talur:
The teardown.sh script fails to run down.sh because the variable evaluates to /down.sh. With these changes the teardown.sh sets the global variables properly and the path evaluates correctly.

### DIFF
--- a/tests/functional/lib.sh
+++ b/tests/functional/lib.sh
@@ -118,9 +118,14 @@ teardown_environment() {
     esac
 }
 
-teardown() {
+_teardown() {
     teardown_environment
     rm -f heketi.db > /dev/null 2>&1
+}
+
+teardown() {
+    init_testenv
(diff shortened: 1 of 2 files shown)


tests/functional/lib.sh
```diff
@@ -139,9 +144,12 @@ pause_test() {
     fi
 }
 
-functional_tests() {
+init_testenv() {
     DEFAULT_TESTENV="${SCRIPT_DIR}/../vagrant"
+}
+functional_tests() {
 
+    init_testenv
```
phlogistonjohn: The formatting here and right above looks messed up.

phlogistonjohn:
Stale PR. Closing. If this work is still relevant please update and resubmit a new PR. Thanks.

PR #1689, opened 2020-01-24
Add support for s390x

sangitanalkar:
As Travis CI [officially supports](https://blog.travis-ci.com/2019-11-12-multi-cpu-architecture-ibm-power-ibm-z) s390x builds, adding support for same and also release binaries for s390x.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,13 @@ install:
 - bash .travis-fork-fix
 - go get github.com/mattn/goveralls
 - go get github.com/wadey/gocovmerge
-- curl https://glide.sh/get | sh
+- if [[ $(uname -m) == 's390x' ]]; then
+      wget https://github.com/Masterminds/glide/releases/download/v0.13.1/glide-v0.13.1-linux-s390x.tar.gz;
+      tar -xzf glide-v0.13.1-linux-s390x.tar.gz linux-s390x/glide --strip=1;
+      export PATH=$PWD:$PATH;
+  else
+      curl https://glide.sh/get | sh;
+  fi
 env:
   global:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -9,7 +9,14 @@ install:
 - bash .travis-fork-fix
 - go get github.com/mattn/goveralls
 - go get github.com/wadey/gocovmerge
-- curl https://glide.sh/get | sh
+- if [[ $(uname -m) == 's390x' ]]; then
+      wget https://github.com/Masterminds/glide/releases/download/v0.13.1/glide-v0.13.1-linux-s390x.tar.gz;
```
phlogistonjohn: Please remove the unnecessary trailing semicolons on this line and the subsequent lines.

.travis.yml
```diff
@@ -9,7 +9,14 @@ install:
...
+      tar -xzf glide-v0.13.1-linux-s390x.tar.gz linux-s390x/glide --strip=1;
+      export PATH=$PWD:$PATH;
+      glide --version;
```
phlogistonjohn: What does this call to `glide --version` accomplish?

.travis.yml
sangitanalkar: Hi @phlogistonjohn , removing the trailing semicolons inside `if-else` condition makes the build fail with `syntax error: unexpected end of file`, hence the semicolons are required.

.travis.yml
sangitanalkar: `glide --version` is not required here, will remove it and commit the change.

.travis.yml
phlogistonjohn: That's weird but may be due to embedding the script within yaml. I'm OK with it for now, perhaps we can improve it later.

phlogistonjohn:
Please squash your changes together. Thanks.

sangitanalkar:
@phlogistonjohn I have squashed the commits, let me know if anything else is needed from my side.

phlogistonjohn:
@raghavendra-talur PTAL

PR #1694, opened 2020-02-04
Randomize pick for BlockHostingVolume

briantward:

Current pick for block hosting volume is the first volume entry in the list.  This changes that to be a random selection, so that should any particular block hosting volume error out, a retry may be performed on the next randomly selected block hosting volume.  Otherwise currently should an error for volume[0] occur and not be fixed, heketi just retries the same erroring block hosting volume when there may be plenty available without issues.


apps/glusterfs/operations_block_volume.go
```diff
@@ -86,8 +88,11 @@ func (bvc *BlockVolumeCreateOperation) Build() error {
 		}
 		reducedSize := ReduceRawSize(BlockHostingVolumeSize)
 		if len(volumes) > 0 {
-			bvc.bvol.Info.BlockHostingVolume = volumes[0].Info.Id
-			bvc.bvol.Info.Cluster = volumes[0].Info.Cluster
+			randSource := rand.NewSource(time.Now().UnixNano())
```
phlogistonjohn: Heketi's main.go seeds the global math rng (see https://github.com/heketi/heketi/blob/master/main.go#L480). There's no need to do it here esp. since this doesn't need much more than a simple shuffle.

apps/glusterfs/operations_block_volume.go
```diff
@@ -86,8 +88,11 @@ func (bvc *BlockVolumeCreateOperation) Build() error {
...
+			randGen := rand.New(randSource)
+			randNum := randGen.Intn(len(volumes))
```
phlogistonjohn: Var name should be something like index or randIndex. randNum (random number) is too vague.

apps/glusterfs/operations_block_volume.go
briantward: Thanks for pointing that out.

apps/glusterfs/operations_block_volume.go
briantward: Done.

briantward:
@phlogistonjohn how are we looking?

phlogistonjohn:
@raghavendra-talur  PTAL

phlogistonjohn:
@raghavendra-talur ping.

PR #1649, opened 2019-09-12
Bump k8s libraries to 1.15.3 levels

liggitt:
# What does this PR achieve? Why do we need it?

* Drops go1.9 support (current versions of client-go use strings.Builder, which is go1.10+... only go1.12 and go1.13 are still in support)
* Updates kubernetes client libraries from Kubernetes v1.6 levels (!) to current versions (v1.15.3)
* Removes dependency on k8s.io/kubernetes (greatly reducing vendor size from 197MB on master to 109MB with this PR)
* Needed to unblock Kubernetes moving to go1.13 (https://github.com/kubernetes/kubernetes/issues/82531)
* Added ability to specify known_hosts files for ssh exec with $SSH_KNOWN_HOSTS (previously host verification was just skipped)

### Does this PR fix issues?

Fixes #1648
Fixes #1110

### Notes for the reviewer

Updated version of https://github.com/heketi/heketi/pull/1542

cc @munnerz

PR #1699, opened 2020-02-20
Fix s390x builds in Travis CI

nixpanic:

While setting up the test environment on Travis CI, pip is run as root.
This seems to create a .cache/pip directory in the travis users home
directory. Subsequently running pip as user travis fails:

    WARNING: Building wheel for d2to1 failed: [Errno 13] Permission denied: '/home/travis/.cache/pip/wheels/4d'

### Does this PR fix issues?

Issue was noticed in #1649. All recent builds failed the same way.


phlogistonjohn:
Thanks a ton for looking into this!

PR #1700, opened 2020-03-03
Remove devices by nested brick evict operations

phlogistonjohn:

Replace the old ad-hoc sequence of exec calls and db modifications made
by the device remove operation with iteration of calls into the newer
brick evict operation.

Modulo bugs this should be more reliable than the old approach because
brick evict's clean is smarter and properly handles failures or stale
operations. All the parent has to do it let the child handle the clean
up of the child operation that failed and then clean its own db state
up. Note that this means the operation is cleaned, but the operation is
not _resumed_. Any operation that is failed or incomplete will only
clean the current child and the device will remain in the offline state
and may still have one ore more bricks on it. But at this point it
should be safe to request removal of the device once again.

### Does this PR fix issues?

Probably, but I couldn't find it so far.

### Notes for the reviewer

Changes include prereqs for establishing parent-child relationships between operations, other helper functions, ability to load existing device remove ops from the db (to make them cleanable) and of course both unit-level and functional tests.

This PR is based on PR #1701 - please review/merge that first.

### DIFF
--- a/apps/glusterfs/pendingop_entry.go
+++ b/apps/glusterfs/pendingop_entry.go
@@ -291,6 +291,61 @@ func (p *PendingOperationEntry) RecordRemoveDevice(d *DeviceEntry) {
 	p.Type = OperationRemoveDevice
 }
 
+// RecordChild adds or replaces a child operation for the current
+// pending operation entry. Both child and parent can only have
+// one parent/child relationship. Both are updated.
+func (p *PendingOperationEntry) RecordChild(child *PendingOperationEntry) {
+	// track childs id on parent
+	parentsChild := findChange(p.Actions, OpChildOperation)
+	if parentsChild < 0 {
+		a := PendingOperationAction{Change: OpChildOperation, Id: child.Id}
+		p.Actions = append(p.Actions, a)
+	} else {
+		p.Actions[parentsChild].Id = child.Id
+	}
+	// track parent's id on child
+	childsParent := findChange(child.Actions, OpParentOperation)
+	if childsParent < 0 {
+		a := PendingOperationAction{Change: OpParentOperation, Id: p.Id}
+		child.Actions = append(child.Actions, a)
+	} else {
+		child.Actions[childsParent].Id = p.Id
+	}
+}
+
+// ClearChild removes any child operations from the parent.
+func (p *PendingOperationEntry) ClearChild() {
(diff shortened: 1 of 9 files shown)


tests/functional/TestErrorHandling/tests/device_remove_test.go
```diff
@@ -0,0 +1,441 @@
...
+	"path"
+	"testing"
+	"time"
+
+	"github.com/heketi/tests"
+
+	inj "github.com/heketi/heketi/executors/injectexec"
+	"github.com/heketi/heketi/pkg/glusterfs/api"
+	_ "github.com/heketi/heketi/pkg/remoteexec"
+	_ "github.com/heketi/heketi/pkg/remoteexec/ssh"
```
phlogistonjohn: Forgot to remove these.

apps/glusterfs/pendingop_entry.go
```diff
@@ -291,6 +291,61 @@ func (p *PendingOperationEntry) RecordRemoveDevice(d *DeviceEntry) {
...
+	}
+	// track parent's id on child
+	childsParent := findChange(child.Actions, OpParentOperation)
+	if childsParent < 0 {
+		a := PendingOperationAction{Change: OpParentOperation, Id: p.Id}
+		child.Actions = append(child.Actions, a)
+	} else {
+		child.Actions[childsParent].Id = p.Id
+	}
+}
```
raghavendra-talur: This function does not seem to cover all the cases. In case of a RecordChild where the parent already had a child recorded wouldn't we leave the pointer to the parent intact on the old child?

phlogistonjohn: Discussed this verbally. I don't think that level of defensive programming is needed at this time. If a "stray" child has an invalid back link to the parent that's a bug in the calling code, not this function.

phlogistonjohn:
This PR is never going to be automatically mergable due to some travis-ci bug. Last run: https://travis-ci.org/github/heketi/heketi/builds/667829356  - status successful.
Are you OK to merge this @raghavendra-talur ?

raghavendra-talur:
> This PR is never going to be automatically mergable due to some travis-ci bug. Last run: https://travis-ci.org/github/heketi/heketi/builds/667829356 - status successful.
> Are you OK to merge this @raghavendra-talur ?

Yes, I will merge this now.

PR #1701, opened 2020-03-13
Assorted brick eviction fixes

phlogistonjohn:

Fixes some issues in the brick eviction operation, primary around the operation clean up path.
The three issues are a bad device ID reference when clean+clean-done is run outside of the rollback code; an invalid nested db transaction; and the inability to tell the difference between never checking the status of brick replace from gluster vs. the brick was not replaced. It also adds some debug logging and a test for the first error condition.

### DIFF
--- a/tests/functional/TestErrorHandling/tests/brick_evict_test.go
+++ b/tests/functional/TestErrorHandling/tests/brick_evict_test.go
@@ -19,6 +19,7 @@ import (
 	"path"
 	"sync"
 	"testing"
+	"time"
 
 	"github.com/heketi/tests"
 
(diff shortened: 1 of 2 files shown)


tests/functional/TestErrorHandling/tests/brick_evict_test.go
```diff
@@ -272,6 +273,85 @@ func TestBrickEvict(t *testing.T) {
...
+						Err: "thwack!",
+					},
+				},
+			}
+		})
+		testutils.ServerRestarted(t, heketiServer)
+
+		toEvict := vinfo1.Bricks[0].Id
+		err = heketi.BrickEvict(toEvict, nil)
+		tests.Assert(t, err != nil, "expected err == nil, got:", err)
```
raghavendra-talur: The string parameter needs change, expected err is not nil.

phlogistonjohn: Oops. Copy-n-paste-o.

PR #1703, opened 2020-03-25
Fix(cli): correct example given in add device cmd help

Elgarni:

The example given in `heketi-cli device add -h` is missing a trailing backslash. 
This PR corrects that example, and updates the corresponding part in `maintenance.md` file as well.

### Does this PR fix issues?

Fixes #  N/A


Elgarni:
@phlogistonjohn Sure

PR #1712, opened 2020-04-17
docs: docker hub integration

raghavendra-talur:

Document how the docker hub is integrated with GitHub for container builds.


phlogistonjohn:
Doc looks reasonable to me although I didn't verify it technically. At least we have a starting point and if we find an inaccuracy later we can just fix it then.

PR #1713, opened 2020-04-20
unit-tests: add tests for block volume expand feature

pkalever:

add tests for block volume expand feature

Signed-off-by: Prasanna Kumar Kalever \<prasanna.kalever@redhat.com\>

### DIFF
--- a/apps/glusterfs/block_volume_expand_test.go
+++ b/apps/glusterfs/block_volume_expand_test.go
@@ -1,0 +1,907 @@
+//
+// Copyright (c) 2020 The heketi Authors
+//
+// This file is licensed to you under your choice of the GNU Lesser
+// General Public License, version 3 or any later version (LGPLv3 or
+// later), or the GNU General Public License, version 2 (GPLv2), in all
+// cases as published by the Free Software Foundation.
+//
+
+package glusterfs
+
+import (
+	"fmt"
+	"os"
+	"testing"
+
+	"github.com/heketi/heketi/pkg/glusterfs/api"
+
+	"github.com/boltdb/bolt"
+	"github.com/heketi/heketi/executors"
+	"github.com/heketi/tests"
+)
+
+func TestBlockVolumeExpandOperationFinalize(t *testing.T) {
+	tmpfile := tests.Tempfile()
+	defer os.Remove(tmpfile)
+
+	// Create the app
+	app := NewTestApp(tmpfile)
+	defer app.Close()
+
+	err := setupSampleDbWithTopology(app,
+		1,    // clusters
+		3,    // nodes_per_cluster
+		1,    // devices_per_node,
+		2*TB, // disksize)
+	)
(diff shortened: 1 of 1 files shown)


pkalever:
@phlogistonjohn @raghavendra-talur please help me review this PR.

Also, I request your help in identifying the uncovered scenarios if there are any.

apps/glusterfs/block_volume_expand_test.go
```diff
@@ -0,0 +1,629 @@
...
+		}
+		return nil
+	})
+
+	e = bve.Exec(app.executor)
+	tests.Assert(t, e == nil, "expected e == nil, got", e)
+
+	// pretend that we have called gluster-block cli and got some info
+	// assumption, gluster-block failed way before expanding the backend file
+	app.xo.MockInfoBlockVolume = func(host string, blockhostingvolume string,
```
raghavendra-talur, line 400: Make this function assignment before calling the exec function. The exec function should then fail. You will look for exec err not nil and then proceed to rollback.

apps/glusterfs/block_volume_expand_test.go
```diff
@@ -0,0 +1,629 @@
...
+			tests.Assert(t, e == nil, "expected e == nil, got", e)
+			tests.Assert(t, blockvolume.Info.Size == 100, "expected block volume size == 100, got:", blockvolume.Info.Size)
+		}
+		return nil
+	})
+
+	e = bve.Exec(app.executor)
+	tests.Assert(t, e == nil, "expected e == nil, got", e)
+
+	// pretend that we have called gluster-block cli and got some info
```
raghavendra-talur, line 574: same here.

apps/glusterfs/block_volume_expand_test.go
pkalever: @raghavendra-talur Exec function doesn't use MockInfoBlockVolume, hence it doesn't matter if we assign the function before or after calling Exec()

But I get your point about failing the Exec(), so I will define a fake app.xo.MockBlockVolumeExpand = {} which will return an error, and as a result, Exec() will fail.

pkalever:
> I have made two comments that require change.
> 
> In addition to that, I can suggest one more test case:
> 
> * Have two block volumes, on the same BHV
> * Request for their expansion such that build() for block expansion is run for both of them before either exec runs.
> * It is expected that the second build fails if the first build left less than required space for the second. on the BHV.
> * Then rollback the first build and retry the second build. This time it should succeed.

This make sense to me, I will add this test case and update the PR soon.

Thanks @raghavendra-talur

pkalever:
@raghavendra-talur @phlogistonjohn Updated the PR addressing the review comments from Talur. PTAL.

Thanks!

pkalever:
Thanks @phlogistonjohn, updated the PR by squashing the patches.

PR #1714, opened 2020-04-23
blockvolume: tuning the semantics of expand command

pkalever:
Currently:
------------

##### Filevolume:
$ heketi-cli volume expand --volume=60d46d518074b13a04ce1022c8c7193c --expand-size=10 

##### Blockvolume:
$ heketi-cli blockvolume expand --block-volume=60d46d518074b13a04ce1022c8c7193c --new-size=10

Fix:
----

I see info and delete commands doesn't use any flag or extra option name specific to collecting volume id, instead, they simply read the arg(0) as volumeId.

Just to keep the semantics in alignment with other commands like info and delete, adding support for below syntax too:

##### Filevolume:
$ heketi-cli volume expand 60d46d518074b13a04ce1022c8c7193c --expand-size=10 

##### Blockvolume:
$ heketi-cli blockvolume expand 60d46d518074b13a04ce1022c8c7193c --new-size=10

But we still maintain the old syntax of using `--volume` or `--blockvolume` in expand commands of both `file` and `block` volumes for compatibility.

Signed-off-by: Prasanna Kumar Kalever \<prasanna.kalever@redhat.com\>

### DIFF
--- a/client/cli/go/cmds/block_volume.go
+++ b/client/cli/go/cmds/block_volume.go
@@ -54,7 +54,7 @@ func init() {
 			"\n\tfor this volume only in the clusters specified.")
 	blockVolumeExpandCommand.Flags().IntVar(&bvNewSize, "new-size", 0,
 		"\n\tNet new size of block volume in GiB")
-	blockVolumeExpandCommand.Flags().StringVar(&bvId, "block-volume", "",
+	blockVolumeExpandCommand.Flags().StringVar(&bvId, "blockvolume", "",
 		"\n\tId of block volume to expand")
 	blockVolumeCreateCommand.SilenceUsage = true
 	blockVolumeDeleteCommand.SilenceUsage = true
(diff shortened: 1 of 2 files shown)


pkalever:
> Consistency is good, however if we compare block volume expand more to volume expand we see that it does have a --volume option. Because we're already partly inconsistent I'm not sure which way we want to go.

Ah! Now I get from where I copied `--block-volume` option from.

But honestly having `volume` or `blockvolume` string twice in a simple command look odd for me.
Like:
```
$ heketi-cli blockvoume expand --block-volume=<id> --new-size=<GiB>

or

$ heleti-cli volume expand --volume=<id> --expand-size=<GiB>
```

But I will leave it up to you guys.


Thanks!

pkalever:
@phlogistonjohn @raghavendra-talur made some difference in the approach, please take a look at the description. Thanks!

client/cli/go/cmds/block_volume.go
```diff
@@ -250,17 +250,28 @@ var blockVolumeListCommand = &cobra.Command{
...
+                     [or, you can also use]
+      $ heketi-cli blockvolume expand --blockvolume=60d46d518074b13a04ce1022c8c7193c --new-size=100
+`,
 	RunE: func(cmd *cobra.Command, args []string) error {
+
 		if bvId == "" {
-			return errors.New("Missing --block-volume=$id")
+			s := cmd.Flags().Args()
+			if len(s) < 1 {
+				return errors.New("Missing blockvolume id")
```
phlogistonjohn, line 266: I'd put a space here to make a it read more like a sentence: "Missing block volume id"

pkalever: Sure, I will refresh.

pkalever:
@phlogistonjohn done now.
Similar changes are made in file `client/cli/go/cmds/block_volume.go` at line numbers 256, 266 and 274 .

Thanks!

Issue #1711, opened 2020-04-17
Travis CI s390 builds are rather flaky


phlogistonjohn:
Travis CI jobs were added for the s390 architecture in PR #1689 and things were all good for a while. However in the past week or so the jobs, specifically the Go 1.13 job on s390 started failing almost every run.

For now I've commented out the Go 1.13 job on s390. Because the current maintenance team is not familiar  with s390 and its apparently not a "set it and forget it" thing but need more care and maintenance. I'm inclined to disable s390 in our travis ci.

@sangitanalkar since you added the jobs I'm @-ing you so you are aware. If you care to keep s390 enabled here in heketi could you please investigate what is going on with these jobs. "go get" is crashing before heketi is even built.

If you don't have the time to work on this I will just disable the jobs. I think I will make my decision in roughly a week.

Thanks.

sangitanalkar:
Hi @phlogistonjohn , sure I will investigate and let you know.

sangitanalkar:
Hi @phlogistonjohn, as mentioned in contributing document for heketi ([link](https://github.com/heketi/heketi/blob/master/docs/contributing.md)) , updated travis.yml file to use latest go v1.14 for s390x and raised PR "https://github.com/heketi/heketi/pull/1715"

phlogistonjohn:
Thank you for the assistance. I think this is all set now. Closing issue.

PR #1715, opened 2020-04-24
Add latest go v1.14 in travis for s390x

sangitanalkar:
Add build for latest go v1.14 in travis for s390x and remove for go v1.13 due to panic memory runtime error for s390x observed only on lxd travis.

### Does this PR fix issues?
Raised PR for fixing issue "https://github.com/heketi/heketi/issues/1711"

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,27 +24,29 @@ git:
   depth: 9999999
 matrix:
   include:
-  - go: "1.10"
-    name: Test using Go 1.10
-    arch: amd64
-  - go: "1.11"
-    name: Test using Go 1.11
-    arch: amd64
-  - go: "1.13"
-    name: Test using Go 1.13
-    arch: amd64
-  - go: "1.13"
-    script: make release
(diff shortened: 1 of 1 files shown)


sangitanalkar:
> Sorry, I don't see this as solving the issue as it frequently affected go 1.13 (to the point that nearly every build failed) but also seems to intermittently fail on the other s390 builds. Does travis-ci have a way to mark some jobs as "optional"? If so, we could try something like that as well as using go 1.14 if its more reliable.

Hi @phlogistonjohn , I tried running `go get `command manually on s390x vm with go v1.13, it works fine without any issues like following. Other jobs for remaining go versions are passing.
```
# go version
go version go1.13 linux/s390x
# go get github.com/mattn/goveralls
# uname -i
s390x
#
```

Also Travis-CI has the option of `allow_failures` ([link](https://docs.travis-ci.com/user/customizing-the-build/#jobs-that-are-allowed-to-fail)) , which allows specific jobs to fail without causing entire build to fail.

PR #1725, opened 2020-05-05
executors: use a default error value in the mock executor

phlogistonjohn:

The mock executor treated every executor call as a "silent success" by
    default. This ran contrary to the way the error injection executor
    was making use of the mock executor. Add a default error value to the
    mock executor, defaulting to nil, that is returned by all the default
    executor functions.

Now that the mock executor can return a per-instance default value for
    all errors on non-overridden functions, set the default error to the
    not-supported sentinel error needed by the exec stack. Now, the mock
    functions do not need to be all individually specified in the error
    injection executor.

This all should make adding new functions to the executor system less error prone in the future.

### Notes for the reviewer

This builds upon the work in PR #1724 - please review and merge that first.


pkalever:
@phlogistonjohn what was the recent change?

pkalever:
> @phlogistonjohn what was the recent change?

Ignore, had noticed it myself now.

phlogistonjohn:
@raghavendra-talur this should only affect test code, its out of date but approved, should we merge or drop this?

iamniting:
I think this can be merged as this is approved and does not require any extra work other than rebasing.

iamniting (approved):

phlogistonjohn:
It's too much churn. I realize everyone is approving this, but I do not want to maintain it. Closing.

PR #1735, opened 2020-06-09
Stop the madness of the "mock" executor

### LINKED ISSUE #1734: Stop defaulting to mock executor, and rename it
### Kind of issue
 Bug

### Observed behavior

Far too often issues are being reported that nothing is working correctly with heketi right after install. Frequently, the cause is that heketi is configured to use the "mock" executor. This executor does nothing and just tells the rest of the system that the operation succeeded.

### Expected/desired behavior

First, heketi should always require an executor be explicitly configured. Since this is environmental, we can't have a sensible default here, so we should have no default. I haven't checked the code so I don't exactly remember what we do if no executor is configured, but it should not be to use mock.

Next, we should rename "mock" to something like "testing-only-mock". This will have two effects, first the new name is a reminder that mock is only for testing heketi. Second, this will (quite intentionally) cause heketi to fail to start if "mock" is used. No one who is trying to run heketi in production should be trying to use mock. Folks that have followed poorly constructed blog posts, etc about setting up heketi will get an error (and perhaps we should enhance the logging around this) and force them to read up on configuring an executor.

phlogistonjohn:

Replace the name of the mock executor with "testing-only-mock".
Improperly configured setups that are trying to use "mock" in production will fail to start rather then pretend to work.

Testing configs in the tree are swapped over. Example jsons now say "CONFIGURE ME" instead of "mock".

### Does this PR fix issues?

Fixes #1734

### DIFF
--- a/apps/glusterfs/app.go
+++ b/apps/glusterfs/app.go
@@ -136,7 +136,7 @@ func (app *App) setup(conf *GlusterFSConfig) error {
 
 	// Setup executor
 	switch app.conf.Executor {
-	case "mock":
+	case "testing-only-mock":
 		app.xo, err = mockexec.NewMockExecutor()
 		app.executor = app.xo
 	case "kube", "kubernetes":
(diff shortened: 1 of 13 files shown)


phlogistonjohn:
Not worth the effort.

PR #1739, opened 2020-06-18
apps: Correct the logging of 'DeviceDelete'

iamniting:

This will achieve correct logging. The older logging can lead us to be in false impression that we have deleted a node, which is not true.


phlogistonjohn:
please try and force push one more time. If that doesn't fix travis I may just override the checks as it is a very small patch and looks obviously correct to me.

iamniting:
Hi John, I have pushed changes forcefully. looks like Travis is still stuck.

phlogistonjohn:
Since travis is still stuck, but centos-ci passed I will force this merge. Thanks for your patience.

PR #1758, opened 2020-07-28
Add github actions files

iamniting:

### Does this PR fix issues?

Fixes #


phlogistonjohn:
We spoke briefly on a call earlier about this, I see the 2nd commit in the PR disables installing python dependencies. Since it has been a while since I looked at how the python test works, is this sufficient to disable the test rather than have it fail?
If so, I'd ask that you squash your commits and add a textual comment explaining why the section pertaining to python is disabled - hopefully temporarily.

I realize that the python bits may seem like a "lesser" part of the codebase given how much of heketi is in Go, but that code *is* part of our system and we should try to maintain it too.

iamniting:
> We spoke briefly on a call earlier about this, I see the 2nd commit in the PR disables installing python dependencies. Since it has been a while since I looked at how the python test works, is this sufficient to disable the test rather than have it fail?
> If so, I'd ask that you squash your commits and add a textual comment explaining why the section pertaining to python is disabled - hopefully temporarily.

done
 
> I realize that the python bits may seem like a "lesser" part of the codebase given how much of heketi is in Go, but that code _is_ part of our system and we should try to maintain it too.

Yes, john you are right we should maintain it too. I will try to fix that or will try to find some workaround.

PR #1772, opened 2020-08-26
github Actions: Change branches master to all

iamniting:
With this change github actions will test the changes on all branches
not just master

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,9 +2,9 @@ name: Test the Pull request
 
 on:
   push:
-    branches: [ master ]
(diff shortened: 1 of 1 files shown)


iamniting:
> Seems fine as an experiment. If it doesn't end up being useful we can always change it back.

Thats true @phlogistonjohn I have tried it on my clone repo before creating pull request it seems to be very useful. 
for example https://github.com/iamniting/heketi/actions

Now it can test other branches as well before even creating pull request. Which will help us to test our changes.

raghavendra-talur:
Ignoring the the centos-ci failure as it seems to be gluster-block related.

PR #1795, opened 2020-10-15
Fix brick eviction when a volume is not present in gluster

phlogistonjohn:

There's a panic that can occur in heketi when brick evict is run. If there's  an eviction to be done, including if part of a device remove, and the volume exists in heketi but not gluster heketi will panic.   This happens because gluster command fails to set a non-zero exit code. Heketi treats this as a success and tries to index into the "list of volumes" for the requested volume but there's a zero-length list, and that triggers a go panic.

### Does this PR fix issues?

Fixes rhbz#1888760

### Notes for the reviewer

At first I didn't like the idea of actually removing the volume from gluster in the test and wanted to just use error injection. However, I concluded that doing it with an actual command is more "real world" and could possibly help find other quirks in this workflow in the future rather than the "mocking" approach of error injection which needs to know what commands to hook.

### DIFF
--- a/executors/cmdexec/volume.go
+++ b/executors/cmdexec/volume.go
@@ -285,6 +285,11 @@ func (s *CmdExecutor) VolumeInfo(host string, volume string) (*executors.Volume,
 		return nil, fmt.Errorf("Unable to determine volume info of volume name: %v", volume)
 	}
 	logger.Debug("%+v\n", volumeInfo)
+	if volumeInfo.OpErrStr != "" {
+		// gluster failed but didn't set a non-zero exit code!
+		return nil, fmt.Errorf("Unable to get info for %v: %v",
+			volume, volumeInfo.OpErrStr)
+	}
 	return &volumeInfo.VolInfo.Volumes.VolumeList[0], nil
 }
 
--- a/tests/functional/TestErrorHandling/tests/brick_evict_test.go
+++ b/tests/functional/TestErrorHandling/tests/brick_evict_test.go
@@ -15,6 +15,7 @@
 package tests
 
 import (
+	"fmt"
 	"os"
 	"path"
 	"sync"
@@ -25,6 +26,8 @@ import (
 
 	inj "github.com/heketi/heketi/executors/injectexec"
 	"github.com/heketi/heketi/pkg/glusterfs/api"
+	rex "github.com/heketi/heketi/pkg/remoteexec"
+	"github.com/heketi/heketi/pkg/remoteexec/ssh"
 	"github.com/heketi/heketi/pkg/testutils"
 	"github.com/heketi/heketi/server/config"
 )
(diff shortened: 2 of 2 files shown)


executors/cmdexec/volume.go
```diff
@@ -285,6 +285,11 @@ func (s *CmdExecutor) VolumeInfo(host string, volume string) (*executors.Volume,
 		return nil, fmt.Errorf("Unable to determine volume info of volume name: %v", volume)
 	}
 	logger.Debug("%+v\n", volumeInfo)
+	if volumeInfo.OpErrStr != "" {
```
raghavendra-talur, line 288: Have to ensure that gluster volume info output always have empty string for success cases. Otherwise, there is no way to determine whether the op failed or succeeded. 

I will do a quick check and update here.

raghavendra-talur (approved):
The changes look good to me.

executors/cmdexec/volume.go
iamniting: I have tried the general scenario. It looks fine to me. I am mainly interested in faulty one how that can be reproduced? So we can check the exact behavior there.

[root@dhcp47-119 ~]# gluster v info test1 >/dev/null
Volume test1 does not exist

[root@dhcp47-119 ~]# gluster v info test1 2>/dev/null

[root@dhcp47-119 ~]# gluster v info test >/dev/null

[root@dhcp47-119 ~]# gluster v info test 2>/dev/null 
Volume Name: test
Type: Replicate
Volume ID: 8e7f0e6d-7a6b-4aa1-b60f-d74efa54d93a
Status: Started
Snapshot Count: 0
Number of Bricks: 1 x 4 = 4
Transport-type: tcp
Bricks:
Brick1: dhcp47-139.lab.eng.blr.redhat.com:/bricks/brick1/b1
Brick2: dhcp47-69.lab.eng.blr.redhat.com:/bricks/brick1/b1
Brick3: dhcp46-149.lab.eng.blr.redhat.com:/bricks/brick1/b1
Brick4: dhcp47-119.lab.eng.blr.redhat.com:/bricks/brick1/b1
Options Reconfigured:
ganesha.enable: on
features.cache-invalidation: on
storage.fips-mode-rchecksum: on
transport.address-family: inet
nfs.disable: on
performance.client-io-threads: off
nfs-ganesha: enable
cluster.enable-shared-storage: enable

iamniting:
@phlogistonjohn According to the summary it looks like, In this scenario, we will have a stale entry in the heketi. Now brick eviction will fail again and again and if someone is removing the device they won't be able to remove the device at all until they delete the volume from heketi which will remove the bricks from the backend and volume from heketi db. It will be good if we can mention something like that in the logs. So they can delete the volume and proceed.

As you know better I want to know your opinion.

executors/cmdexec/volume.go
raghavendra-talur: @iamniting Thanks for the output for all the cases.

raghavendra-talur (approved):

phlogistonjohn:
@iamniting could you clarify / restate your message. I'm not really clear on what you mean.

iamniting (approved):

iamniting:
> @iamniting could you clarify / restate your message. I'm not really clear on what you mean.

All doubts are cleared over chat.

PR #1796, opened 2020-10-21
functional tests: add a way to debug g-block test failures

phlogistonjohn:

Lately, the tests are failing due to mysterious error within
gluster-block in the CI environment. This is my 2nd, slightly cleaner,
attempt at diagnosing gluster block errors when the tests fail in the
CI.

### Does this PR fix issues?

Fixes #

### DIFF
--- a/tests/functional/vagrant/dump.sh
+++ b/tests/functional/vagrant/dump.sh
@@ -1,0 +1,17 @@
+#!/bin/sh
+
+set -e
+
+SCRIPT_DIR="$(cd "$(dirname "${0}")" && pwd)"
+cd "${SCRIPT_DIR}"
+
+set +e
+set -x
+for hn in storage0 storage1 storage2 storage3; do
+    echo '+++' $hn
+    vagrant ssh $hn -- rpm -qa | grep gluster
+    vagrant ssh $hn -- systemctl status gluster-blockd
+    vagrant ssh $hn -- sudo grep . /var/log/gluster-block/gluster-block-cli.log /var/log/gluster-block/gluster-block-configshell.log /var/log/gluster-block/gluster-block-gfapi.log /var/log/gluster-block/gluster-blockd.log
+done
+
+exit 0
--- a/tests/functional/lib.sh
+++ b/tests/functional/lib.sh
@@ -118,6 +118,12 @@ teardown_environment() {
     esac
 }
 
+dump_logs() {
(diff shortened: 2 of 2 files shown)


raghavendra-talur (approved):

tests/functional/vagrant/dump.sh
```diff
@@ -0,0 +1,17 @@
...
+
+SCRIPT_DIR="$(cd "$(dirname "${0}")" && pwd)"
+cd "${SCRIPT_DIR}"
+
+set +e
+set -x
+for hn in storage0 storage1 storage2 storage3; do
+    echo '+++' $hn
+    vagrant ssh $hn -- rpm -qa | grep gluster
+    vagrant ssh $hn -- systemctl status gluster-blockd
```
iamniting, line 13: It will be good if we can add tcmu-runner and targetcli as well.

tests/functional/vagrant/dump.sh
```diff
@@ -0,0 +1,17 @@
...
+    vagrant ssh $hn -- sudo grep . /var/log/gluster-block/gluster-block-cli.log /var/log/gluster-block/gluster-block-configshell.log /var/log/gluster-block/gluster-block-gfapi.log /var/log/gluster-block/gluster-blockd.log
```
iamniting, line 14: Is there a specific reason we are doing like it instead of cat

phlogistonjohn: it will prefix the output with the names of the files

tests/functional/vagrant/dump.sh
phlogistonjohn: I don't see a strong need to do that as of yet. Let's prove that the current approach adds value first.

tests/functional/vagrant/dump.sh
iamniting: Perfect

tests/functional/vagrant/dump.sh
iamniting: cool no problem

iamniting (approved):

phlogistonjohn:
I actually wanted this to fail on the random gluster block errors we've been seeing. However, I'm apparently not that lucky. So without knowing if this is really going to help debug g-block failures (but at least it doesn't break successes) I'm merging anyway.

PR #1702, opened 2020-03-24
gluster-block: implement expand feature interface

pkalever:
### What does this PR achieve? Why do we need it?
Implement block volume expand interface in heketi

### Notes for the reviewer

Pushing it live with the intention that I will have my code side review comments ready by that time I finish on the TODO list.

The RFE just works fine now but also check for TODO list at the end of this comment which I got from John and Talur in a demo meeting.

```
Before Expand:
-------------
[root@server1 ~]# heketi-cli blockvolume create --size 1
Name: blockvol_365f428f2bf4b5a37f644c239f148058
Size: 1
Volume Id: 365f428f2bf4b5a37f644c239f148058
Cluster Id: 0b5740b0e3c34e935c1cd73e8bbf3c71
Hosts: [192.168.121.42 192.168.121.68 192.168.121.155]
IQN: iqn.2016-12.org.gluster-block:afcc41a3-1c46-4102-812b-5a0767cfb16a
LUN: 0
Hacount: 3
Username:
Password:
Block Hosting Volume: 7b3e7fceae17a57e0fc9ddedcce75912

[root@server1 ~]# heketi-cli volume info 7b3e7fceae17a57e0fc9ddedcce75912 | grep "Free Size"
Free Size: 8

[root@server1 ~]# targetcli ls | grep vol_7b3e7fceae17a57e0fc9ddedcce75912
 o- blockvol_365f428f2bf4b5a37f644c239f148058 [vol_7b3e7fceae17a57e0fc9ddedcce75912@... (1.0GiB) activated]

After Expand:
------------
[root@server1 ~]# heketi-cli blockvolume expand --block-volume 365f428f2bf4b5a37f644c239f148058 --new-size 2
Name: blockvol_365f428f2bf4b5a37f644c239f148058
Size: 2
Volume Id: 365f428f2bf4b5a37f644c239f148058
Cluster Id: 0b5740b0e3c34e935c1cd73e8bbf3c71
Hosts: [192.168.121.42 192.168.121.68 192.168.121.155]
IQN: iqn.2016-12.org.gluster-block:afcc41a3-1c46-4102-812b-5a0767cfb16a
LUN: 0
Hacount: 3
Username:
Password:
Block Hosting Volume: 7b3e7fceae17a57e0fc9ddedcce75912

[root@server1 ~]# heketi-cli volume info 7b3e7fceae17a57e0fc9ddedcce75912 | grep "Free Size"
Free Size: 7

[...]

PR #1815, opened 2020-12-04
apps: revert the renumbering of enum in pendingop

raghavendra-talur:
In PR #1702 we added a new OperationType and ChangeType. However, it
renumbered the consts. In cases where heketi is upgraded when the
db has pending ops, this can lead to undefined behavior.

### DIFF
--- a/apps/glusterfs/pendingop.go
+++ b/apps/glusterfs/pendingop.go
@@ -28,16 +28,22 @@ import (
 type PendingOperationType int
 
 const (
+	// These values are persisted and we should never change the mapping. Even
+	// if an operation type is deprecated, use a new one and retire the number
+	// associated with it.
+
 	OperationUnknown PendingOperationType = iota
 	OperationCreateVolume
 	OperationDeleteVolume
 	OperationExpandVolume
 	OperationCreateBlockVolume
 	OperationDeleteBlockVolume
-	OperationExpandBlockVolume
 	OperationRemoveDevice
 	OperationCloneVolume
 	OperationBrickEvict
+	OperationExpandBlockVolume
+	// If you have any edit except directly above this line it is probably a
+	// mistake. Read the top comment about renumbering the enums.
 )
 
 // PendingChangeType identifies what kind of lower-level new item or change
@@ -45,6 +51,10 @@ const (
 type PendingChangeType int
 
 const (
+	// These values are persisted and we should never change the mapping. Even
+	// if a change type is deprecated, use a new one and retire the number
+	// associated with it.
+
 	OpUnknown PendingChangeType = iota
 	OpAddBrick
 	OpAddVolume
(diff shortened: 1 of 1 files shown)


phlogistonjohn:
I'm rather upset to see that this slipped by us in code review. This is pretty bad.

Out of the operations that got renumbered I don't think there's a lot of use of OperationCloneVolume but I don't know how many users are using OperationRemoveDevice or OperationBrickEvict. The latter is quite new and I doubt there are many uses of it in the wild, but its a really rough guess.

This issue with just "renumbering back" is that any recent users with stale ops in the db could be "broken" by this change. It is difficult to say but there's probably still more 9x users than 10x out there so renumbering could then be best of a bad situation.

I would suggest that we add very loud, obnoxious comments around these enums to shout to the heavens these are persisted enums and thus the existing ordering must not change. Later we may even consider forcing the use of explict number values instead of letting go iota do it!

raghavendra-talur:
Added comments. Let me know if you would like to phrase it differently.

iamniting (approved):
I have never used iota before, It is good to know that things can be done this way.

phlogistonjohn:
> I have never used iota before, It is good to know that things can be done this way.

Well... I'm feeling pretty burned by this feature right now, so I would suggest NOT using it in the future if the values are meant to be persisted in any way.

apps/glusterfs/pendingop.go
```diff
@@ -28,23 +28,31 @@ import (
 type PendingOperationType int
 
 const (
+	// These are enum and we should never change the mapping. Even if an
```
phlogistonjohn, line 31: Let's change "These are enum" to "These values are persisted". Both here and in the other location.

phlogistonjohn (changes requested):
One wording tweak and then I'll approve. Everything else looks good +1 to a comment at the top and bottom.

apps/glusterfs/pendingop.go
raghavendra-talur: done

apps/glusterfs/pendingop.go
```diff
@@ -28,23 +28,33 @@ import (
 type PendingOperationType int
 
 const (
+	// These values are persisted and we should never change the mapping. Even
+	// if an operation type is deprecated, use a new one and retire the number
+	// associated with it.
+
```
raghavendra-talur, line 34: Needed an empty line here because the comment was being taken as a doc for the exported const OperationUnknown.

apps/glusterfs/pendingop.go
```diff
@@ -28,23 +28,33 @@ import (
...
 
 // PendingChangeType identifies what kind of lower-level new item or change
 // is being made to the system as part of a higher-level pending operation.
 type PendingChangeType int
 
 const (
+	// These values are persisted and we should never change the mapping. Even
+	// if a change type is deprecated, use a new one and retire the number
+	// associated with it.
+
```
raghavendra-talur, line 57: Needed an empty line here because the comment was being taken as a doc for the exported const OpUnknown.

phlogistonjohn (approved):
LGTM thanks

PR #1822, opened 2020-12-21
Switch to form3-tech/jwt-go.

dlorenc:

The old library is abaondoned and has some severe CVEs in it. The form3-tech
fork has fixed those CVEs and is the most popular fork. CVE is here:
https://snyk.io/vuln/SNYK-GOLANG-GITHUBCOMDGRIJALVAJWTGO-596515

### Does this PR fix issues?
Fixes #

### Notes for the reviewer

I understand this is going toward maintenance mode. Hoping to get this update in to help unblock other libraries that import heketi from updating.

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -1,10 +1,11 @@
 package: github.com/heketi/heketi
 import:
 - package: github.com/auth0/go-jwt-middleware
+  version: 36081240882bbf356af6efb152969e4b0bcf4456
 - package: github.com/boltdb/bolt
   version: ^1.3.0
-- package: github.com/dgrijalva/jwt-go
-  version: ^3.0.0
+- package: github.com/form3tech-oss/jwt-go
+  version: ^3.2.2
 - package: github.com/gorilla/context
 - package: github.com/gorilla/mux
 - package: github.com/heketi/rest
--- a/glide.yaml
+++ b/glide.yaml
@@ -1,7 +1,7 @@
 package: github.com/heketi/heketi
 import:
 - package: github.com/auth0/go-jwt-middleware
-  version: 36081240882bbf356af6efb152969e4b0bcf4456
+  version: v1.0.0
 - package: github.com/boltdb/bolt
   version: ^1.3.0
 - package: github.com/form3tech-oss/jwt-go
--- a/middleware/jwt.go
+++ b/middleware/jwt.go
@@ -22,7 +22,7 @@ import (
 	"time"
 
 	jwtmiddleware "github.com/auth0/go-jwt-middleware"
-	jwt "github.com/dgrijalva/jwt-go"
+	jwt "github.com/form3tech-oss/jwt-go"
 	"github.com/heketi/heketi/pkg/logging"
 )
 
@@ -42,7 +42,7 @@ func init() {
 	}
 }
 
-// From https://github.com/dgrijalva/jwt-go/pull/139 it is understood
+// From https://github.com/form3tech-oss/jwt-go/pull/139 it is understood
 // that if the machine where jwt token is generated and/or the machine
 // where jwt token is verified have any clock skew then there is a
 // possibility of getting a "Token used before issued" error.
--- a/middleware/jwt.go
+++ b/middleware/jwt.go
@@ -42,7 +42,7 @@ func init() {
 	}
 }
 
-// From https://github.com/form3tech-oss/jwt-go/pull/139 it is understood
+// From https://github.com/dgrijalva/jwt-go/pull/139 it is understood
 // that if the machine where jwt token is generated and/or the machine
 // where jwt token is verified have any clock skew then there is a
 // possibility of getting a "Token used before issued" error.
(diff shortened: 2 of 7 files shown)


raghavendra-talur:
@centos-ci  ok-to-test

dlorenc:
The test was failing with something that looked unrelated. I added a commit that I think addresses this.

dlorenc:
The build error looks like some kind of timeout. Is there a way to rerun it?

dlorenc:
This time it looked like a vagrant issue:

```
14:30:12 + vagrant plugin install vagrant-libvirt
14:30:25 Installing the 'vagrant-libvirt' plugin. This can take a few minutes...
14:30:41 Building native extensions.  This could take a while...
14:30:43 Vagrant failed to properly resolve required dependencies. These
14:30:43 errors can commonly be caused by misconfigured plugin installations
14:30:43 or transient network issues. The reported error is:
14:30:43 
14:30:43 nokogiri requires Ruby version < 3.1.dev, >= 2.5.

```

iamniting:
Python tests are also getting failed in Github actions. It looks like this got introduced now in the new python version.

```
2021-01-06T14:08:20.5728022Z >       headers['Authorization'] = b'bearer ' + token
2021-01-06T14:08:20.5728729Z E       TypeError: can't concat str to bytes
```

```
>>> 'hello' + 'hello'
'hellohello'
>>> b'hello' + 'hello'
Traceback (most recent call last):
  File "<input>", line 1, in <module>
    b'hello' + 'hello'
TypeError: can't concat str to bytes
```

phlogistonjohn:
The changes generally look pretty reasonable. The changes to glide look minimal, which is very important, so thanks for keeping them small.

Once some of the CI issues are sorted out we can do proper reviews.

One small request would be to fold your 2 changesets together as having the test fix in with the import changes is more bisection friendly.

dlorenc:
> One small request would be to fold your 2 changesets together as having the test fix in with the import changes is more bisection friendly.

All set! I'll keep poking at the CI stuff, not sure if I'll be able to fix it though.

phlogistonjohn:
Looks like a legitimate build failure due to one of our other dependencies, [github.com/auth0/go-jwt-middleware](https://github.com/auth0/go-jwt-middleware) still relying on the previous jwt lib. But looking at that project's readme on github its also been switched to use this fork. Perhaps the glide entry for that lib just needs to be tweaked too, @dlorenc ?  Were you able to build with the current glide config locally?

iamniting (approved):

phlogistonjohn:
Great, thanks! The centos ci seems to be having an issue with glide timing out (possibly a network issue?). I want to take a little more time to see if we can sort that out, but another week passes w/o centos ci passing one of may just need to run some tests manually and then we can get on with preparing to merge

dlorenc:
> Great, thanks! The centos ci seems to be having an issue with glide timing out (possibly a network issue?). I want to take a little more time to see if we can sort that out, but another week passes w/o centos ci passing one of may just need to run some tests manually and then we can get on with preparing to merge

Is there a way to rerun it? The glide steps take several minutes on my laptop. I think that happened earlier to me too, but a rerun fixed it.

iamniting:
The above comment will run the CI again. But I tried many times on the other PR it is not helping out.

phlogistonjohn:
@iamniting I think @dlorenc is suggesting to run the glide command more than once in the test, not to start the CI job again.
It's worth a try I think.

phlogistonjohn:
I have an experiment at #1833 to try a Fonzie-type approach to getting the CI will do anything.

iamniting:
Cool this one has also passed, Now centos CI is on track

phlogistonjohn (approved):
Tests are now passing and everything seems to check out. Thanks!

phlogistonjohn:
@raghavendra-talur PTAL, thanks.

middleware/jwt.go
```diff
@@ -42,7 +42,7 @@ func init() {
 	}
 }
 
-// From https://github.com/dgrijalva/jwt-go/pull/139 it is understood
+// From https://github.com/form3tech-oss/jwt-go/pull/139 it is understood
```
raghavendra-talur, line 45: This change does not make sense. The PR is only available in the dgrijalva repo. We can keep the comment as-is or modify the comment to not refer to any PR altogether.

glide.yaml
```diff
@@ -1,10 +1,11 @@
 package: github.com/heketi/heketi
 import:
 - package: github.com/auth0/go-jwt-middleware
+  version: 36081240882bbf356af6efb152969e4b0bcf4456
```
raghavendra-talur, line 4: This commit is from 2019 and as @phlogistonjohn mentioned, it will bring in dgrijalva/jwt-go as its dependency. I suggest that the version be bumped up to v1.0.0. v1.0.0 uses the fork "github.com/form3tech-oss/jwt-go".

raghavendra-talur, line 4:
```suggestion
  version: v1.0.0
```

raghavendra-talur (changes requested):

phlogistonjohn:
@dlorenc have you seen @raghavendra-talur 's feedback? Do you plan on making changes in reponse?

raghavendra-talur (approved):

PR #1830, opened 2021-01-07
client: lock PyJWT to "pyjwt >= 1.4.0, < 2.0.0"

iamniting:
Fix "TypeError: can't concat str to bytes"

The PR will fix the test failures in the Github actions which are required to test incoming PR's


iamniting:
Failure examples from the other run https://github.com/heketi/heketi/pull/1822/checks?check_run_id=1656756010#step:7:2057

client/api/python/heketi/heketi.py
```diff
@@ -55,7 +55,7 @@ def _set_token_in_header(self, method, uri, headers={}):
         claims['qsh'] = hashlib.sha256(val).hexdigest()
 
         token = jwt.encode(claims, self.key, algorithm='HS256')
-        headers['Authorization'] = b'bearer ' + token
+        headers['Authorization'] = 'bearer ' + token
```
phlogistonjohn, line 58: Why is changing the type from bytes to (unicode) str the right approach? Why should token be encoded to bytes instead? Also if this recently changed perhaps it should handling token being either str or bytes?

iamniting: Actually, I was thinking that we are getting str from jwt.encode() and did not want to touch it. Moreover, we are creating bytes ourselves that's the main reason to change it this way.

That's also possible that jwt.encode() might be sending bytes before instead of string.

iamniting:
Hi @phlogistonjohn as discussed over chat I have made changes to the requirements.txt and locked pyJWT

phlogistonjohn (approved):
This looks good. If we really need to move to 2.x at some point in the future we can do so at our own pace.

raghavendra-talur (approved):

PR #1833, opened 2021-01-20
makefile: set a timeout on glide but allow a 2nd attempt

phlogistonjohn:
Set a timeout of 20minutes on running glide but if glide fails
run it again. This is just an experiment to see if it makes the
CI behave better.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -88,7 +88,7 @@ endif
 glide.lock: glide.yaml
 	echo "Glide.yaml has changed, updating glide.lock"
 	if [ "$(GLIDE_HOME)" ]; then mkdir -p "$(GLIDE_HOME)"; fi
-	$(GLIDE) update -v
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -88,7 +88,7 @@ endif
 glide.lock: glide.yaml
 	echo "Glide.yaml has changed, updating glide.lock"
 	if [ "$(GLIDE_HOME)" ]; then mkdir -p "$(GLIDE_HOME)"; fi
-	$(GLIDE) update -v
+	timeout -v 20m $(GLIDE) update -v || timeout -v 20m $(GLIDE) update -v
```
iamniting, line 91: John, changes are required in the install, Install is the one where CentOS ci is failing. I raised a test PR to see if that works. If it works then we can make the same changes here and get this merged.

iamniting (changes requested):

phlogistonjohn:
glitchy button

PR #1834, opened 2021-01-21
Makefile glide experiment

iamniting:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,7 +83,7 @@ ifndef HGPATH
 endif
 	echo "Installing vendor directory"
 	if [ "$(GLIDE_HOME)" ]; then mkdir -p "$(GLIDE_HOME)"; fi
-	$(GLIDE) install -v
+	timeout 20m $(GLIDE) install -v || timeout 20m $(GLIDE) install -v
 
 glide.lock: glide.yaml
 	echo "Glide.yaml has changed, updating glide.lock"
@@ -155,7 +155,7 @@ linux_arm_dist:
 
 linux_arm64_dist:
 	GOOS=linux GOARCH=arm64 $(MAKE) dist
-	
+
 linux_s390x_dist:
 	GOOS=linux GOARCH=s390x $(MAKE) dist
 


iamniting:
The good part is that it worked, But it got timeout after 3 hours. I think Centos CI has set a limit to run any CI for up to 3 hrs. They abort it after 3 hours. Can we request them to increase this time?

Makefile
```diff
@@ -83,7 +83,7 @@ ifndef HGPATH
 endif
 	echo "Installing vendor directory"
 	if [ "$(GLIDE_HOME)" ]; then mkdir -p "$(GLIDE_HOME)"; fi
-	$(GLIDE) install -v
+	timeout 20m $(GLIDE) install -v || $(GLIDE) install -v
```
phlogistonjohn, line 86: Use timeout -v and put the timeout on both glide invocations please.

phlogistonjohn (changes requested):
Looks like the general gist of it worked. It just needs a few small tweaks and we can merge it.

Makefile
iamniting: I removed -v intentionally as I see that in centos timeout does not have that option

Screenshot from 2021-01-21 19-57-08

phlogistonjohn: OK. We just want the 2nd invocation to also time out in case it gets stuck

phlogistonjohn (approved):

iamniting:
> The good part is that it worked, But it got timeout after 3 hours. I think Centos CI has set a limit to run any CI for up to 3 hrs. They abort it after 3 hours. Can we request them to increase this time?

The PR has been raised to increase the timeout to 5 hrs https://github.com/gluster/centosci/pull/130

iamniting:
The PR has passed now and it took less than 3 hrs after increasing the timeout.

Screenshot from 2021-01-22 16-44-00

phlogistonjohn:
Fantastic. I will give @raghavendra-talur to the end of the day to look & review.

PR #1863, opened 2021-06-27
Fix topology never load on first run

pulcov:

On first run, flock function create file heketi.db if not exists, and topology isn't load after.

### Does this PR fix issues?

Function flock is call only if file heketi.db is present.


phlogistonjohn:
Hi. Sorry for the extremely late response, however as heketi is in deep maintenance mode we don't have much bandwidth for maintaining heketi beyond critical fixes and security items. As such, we don't have the capability to really judge this change now. Because the existing script apparently works for most users I have to decline your contribution and just suggest that if you need these changes on your setup, you maintain a fork of heketi for now.

I hope you understand.

PR #1867, opened 2021-07-19
Build with go modules

phlogistonjohn:
Build heketi using go modules. As one of the commit messages notes this is also a full dependencies refresh as a side effect.

The motivation to do this is largely due to go planning on [dropping support for non-modules builds in Go 1.17](https://blog.golang.org/go116-module-changes#TOC_2.) (the next release at the time of this post).  And we're concerned that we'll have to keep heketi on life support for some time after that release is out. If we need to make real changes or deal with security fixes in Go or dependencies after Go 1.17 is the only supported Go version we'd be having problems.

So it is.


iamniting (commented):
these changes look good to me.

Moreover, we need some changes in our Docker file and GitHub action as well
https://github.com/heketi/heketi/blob/master/extras/docker/fromsource/Dockerfile
https://github.com/heketi/heketi/blob/master/.github/workflows/ci.yml

phlogistonjohn:
I've marked this do not merge for now, despite taking it out of draft. It's basically ready for review but before we think about merging it it needs to be tested using the k8s executor, which our automated tests can't do.

So once I or someone (but let's face it, I probably need to do it myself) tests out these changes under k8s and things work OK we can then remove the label and think about merging it.

iamniting (approved):
Changes look good to me. 

There are some changes in Travis CI WRT the glide, But I think we can ignore them as of now as we are no longer considering Travis CI.

phlogistonjohn:
Some basic testing on a k8s cluster has been performed. no obvious issues detected.

extras/release/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+# install dependencies, build and cleanup
+RUN mkdir $BUILD_HOME $GOPATH && \
+    yum -y install git make mercurial && \
+    mkdir -p $GOPATH/src/github.com/heketi && \
+    curl -o /tmp/$GOTAR https://storage.googleapis.com/golang/$GOTAR && \
+    tar xzf /tmp/$GOTAR -C /usr/local && \
+    cd $GOPATH/src/github.com/heketi && \
+    git clone -b $HEKETI_BRANCH $HEKETI_REPO && \
+    cd $GOPATH/src/github.com/heketi/heketi && \
+    true
```
iamniting, line 23: Do not we want to run make commands here in this Dockerfile

iamniting (changes requested):

iamniting:
> Some basic testing on a k8s cluster has been performed. no obvious issues detected.

That's awesome

extras/release/Dockerfile
phlogistonjohn: Not for this. The idea here is the container merely acts as a "clean" environment for us to perform the builds. It doesn't do the work for us, unlike the fromsource container.

iamniting (approved):

raghavendra-talur (approved):

go.mod
```diff
@@ -0,0 +1,20 @@
+module github.com/heketi/heketi
```
pohly, line 1: This has to include the version:
```
module github.com/heketi/heketi/v10
```

Then all imports in your code also need to use that instead of plain `github.com/heketi/heketi`.

It works as-is when working inside the repo, but any consumers referencing the code run into https://github.com/heketi/heketi/issues/1897

PR #1874, opened 2021-08-04
workflows: add go versions 1.15 & 1.16 to actions runs

phlogistonjohn:
We're only testing "old" go versions currently. We should test old versions as needed, but also test versions actually supported by Go.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


phlogistonjohn:
Note, I don't plan on dropping the older go versions in this PR. I first want to merge this, then change the list of checks that are required to pass in github ui, then drop 2 of the old ones - retaining one old version to ensure no-one accidentally adds something that prevents us from building on older versions.

iamniting (approved):

raghavendra-talur (approved):

PR #1875, opened 2021-08-10
extras: add dockerfile for building with old versions of go

phlogistonjohn:
This isn't meant for production, just dev & test when needing to ensure
that heketi can be built using older go versions.

You can use the GO_VERSION build arg to specify other go versions w/o having to edit the dockerfile too.

### DIFF
--- a/extras/docker/oldgo-fromsource/Dockerfile
+++ b/extras/docker/oldgo-fromsource/Dockerfile
@@ -1,0 +1,47 @@
+# set author and base
+FROM fedora
+MAINTAINER Heketi Developers <heketi-devel@gluster.org>
+ARG HEKETI_REPO="https://github.com/heketi/heketi.git"
+ARG HEKETI_BRANCH="master"
+ARG GO_VERSION=1.13.15
+
+LABEL version="1.3.1"
+LABEL description="Development build"
+
+# let's setup all the necessary environment variables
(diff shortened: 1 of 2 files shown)


extras/docker/oldgo-fromsource/Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+LABEL description="Development build"
+
+# let's setup all the necessary environment variables
+ENV BUILD_HOME=/build
+ENV GOPATH=$BUILD_HOME/golang
+ENV PATH=$GOPATH/bin:$PATH
+# where to clone from
+ENV HEKETI_REPO=$HEKETI_REPO
+ENV HEKETI_BRANCH=$HEKETI_BRANCH
+ENV GO111MODULE=off
```
iamniting, line 18: I think this should be on now as we are using go modules

iamniting (changes requested):

extras/docker/oldgo-fromsource/Dockerfile
phlogistonjohn: I don't want it enabled for the old stuff because I want to ensure the old way of building continues to work. We can consider something like this for the 'fromsource' container though.

iamniting (approved):

raghavendra-talur (approved):

PR #1890, opened 2021-11-28
Updates to put emphasis on heketi's "deep" maintenance status

phlogistonjohn:
Updates to the issue template and readme folllow.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,12 +1,14 @@
 <!--
-This issue template is meant mainly for bug reports and feature requests.
+This issue template is meant mainly for bug reports.
 
-If you are looking for help, you can reach out to the IRC channel
-#heketi on freenode or use the mailing list heketi-devel@gluster.org.
+Please note that the heketi project is in maintenance only. Feature requests are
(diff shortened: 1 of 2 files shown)


iamniting (approved):

amarts:
While you are at it, should we provide possible next steps for people visiting this page / project? 

Like adding a small section on 'Other opensource projects which would help you with storage solutions' or something like it? It can mention rook.io, kadalu.io, openebs etc, and also mention people to check with gluster community from gluster.org if there are any new projects if they prefer to use GlusterFS as their backend?

It may help those who doesn't want to run another search on search engine.

phlogistonjohn:
Hi @amarts - I'd prefer not to endorse any particular "successor" project. While I understand that it might mean a bit of extra work for people to research solutions but I prefer not to nudge people in any one direction. Thanks!

phlogistonjohn:
@amarts as an alternative to having the project promote any particular set of projects, we opted to create a wiki page that anyone should be able to contribute to without our direct involvement. https://github.com/heketi/heketi/wiki/Heketi-Alternatives

Feel free to add any projects there that you want. I added rook there as an example.

PR #1903, opened 2022-06-11
Update k8s.io dependencies to v0.24.1

dims:
Signed-off-by: Davanum Srinivas <davanum@gmail.com>

### Does this PR fix issues?

Fixes #

PR #1898, opened 2022-04-20
Fix the v10.4 release module name

slaskawi:
### What does this PR achieve? Why do we need it?

This Pull Request adds the `v10` prefix to the module names, so that we can pull them into the kubernetes repo.

### Does this PR fix issues?

Fixes #1897

### Notes for the reviewer

We will need another release containing this commit.

PR #1904, opened 2022-06-14
update kuberenetes dependencies to 1.24.1

humblec:
There are few things covered in this PR.

- [x] Kube dependencies are updated to latest version 1.24.1
- [x] go version in CI  validation has been switched to 1.17 and 1.18
- [x] The secret creation and update has been modified to have
latest signature. Also the listOptions on pod has new signature.

Signed-off-by: Humble Chirammal <hchiramm@redhat.com>
Signed-off-by: Davanum Srinivas [davanum@gmail.com](mailto:davanum@gmail.com)

Reference #https://github.com/heketi/heketi/pull/1903


phlogistonjohn:
Could you please explain the rationale behind the change?   We've had k8s dependencies that greatly lagged the current kubernetes version in the past and it never seemed to bother anyone.  
I'm really at the point where I'd like to archive heketi but we can't do that quite yet. That said I am trying to keep the changes to the barest minimum. 

If we chose not to merge these changes, what problems would you see?

dims:
@phlogistonjohn of course i can explain. There is a vendored dependency in kubernetes called `github.com/form3tech-oss/jwt-go` which is marked `This repository has been archived by the owner. It is now read-only.` To remove this dependency from kubernetes, @Haimantika filed [2] in this repository. So now, we will need a `heketi` release with the commit and we should be done ... right?

Unfortunately that does not work as there is circular dependency between kubernetes and heketi. and the new version of heketi (with older k8s.io/* dependency) will fail when we try to integrate into kubernetes. So we need to update heketi first and then pull in a new release tag (or master tip) of heketi into kubernetes.

Hope this helps.

[1] https://github.com/kubernetes/kubernetes/blob/master/go.mod#L167
[2] https://github.com/heketi/heketi/commit/f0ea1e7addf6b3eac635fc0eaae7c41cb57bba72

dims:
of course the other option is to entirely remove heketi and glusterfs if you all think that there is no need for it in kubernetes/kubernetes repository. that would definitely break the circular dependency.

dims:
`I'd be unhappy` ... we'll need to battle for it as i will be MORE unhappy than you! :)

https://github.com/heketi/heketi/pull/1898 i have no idea if this will make it easier or harder, will defer to you.

dims:
Looks like this is still open - https://github.com/kubernetes/kubernetes/issues/100897

"glusterfs CSI migration support or deprecation needed"

raghavendra-talur:
> and the new version of heketi (with older k8s.io/* dependency) will fail when we try to integrate into kubernetes. So we need to update heketi first and then pull in a new release tag (or master tip) of heketi into kubernetes.

@humblec @dims 
Is the failure you refer to here a compilation error when you try to compile Kubernetes with updated Heketi version? Can I try it with just the latest k8s and by changing the version of Heketi using go mod?

.github/workflows/ci.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        go: ["1.15", "1.16"]
+        go: ["1.17", "1.18"]
```
phlogistonjohn, line 16: Feel free to add the newer versions of Go, but let's keep the old versions too. We need to ensure that heketi can be compiled on older OS platforms.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/heketi/heketi
 
-go 1.14
+go 1.18
```
phlogistonjohn, line 3: Let's not change this.

.github/workflows/ci.yml
humblec: I doubt this is going to compile with `1.15` or `1.16` versions, but let me add those and check as well.

humblec:
``` go build  -ldflags "-X main.HEKETI_VERSION=v10.0.0-87-gabe8b550-HEAD -extldflags '-z relro -z now'" -o heketi
Error: vendor/k8s.io/client-go/plugin/pkg/client/auth/exec/metrics.go:21:2: cannot find package "." in:
	/home/runner/go/src/github.com/heketi/heketi/vendor/io/fs```

File: go.mod [hunk 3]
```diff
@@ -10,11 +10,34 @@ require (
...
+	golang.org/x/crypto v0.0.0-20220525230936-793ad666bf5e
+	k8s.io/api v0.24.1
+	k8s.io/apimachinery v0.24.1
+	k8s.io/client-go v0.24.1
+)
+
+require (
+	github.com/asaskevich/govalidator v0.0.0-20210307081110-f21760c49a8d // indirect
+	github.com/emicklei/go-restful/v3 v3.8.0 // indirect
+	github.com/form3tech-oss/jwt-go v3.2.5+incompatible // indirect
```
slaskawi, line 25: Could you please make sure we don't include this one? I'm currently working on replace it with `golang-jwt/jwt` here: https://github.com/kubernetes/kubernetes/issues/104007

slaskawi (changes requested):

go.mod
humblec: @slaskawi rebased to master which picked that change.

humblec:
@phlogistonjohn enabling 1.15 and 1.16 test fail  https://github.com/heketi/heketi/pull/1904#discussion_r899017925 with the changes bringing here in this PR . Considering we have the last heketi version ( `v10.4`) with these  go versions 1.15 and 1.16 already , also accounting this project is in maintenance mode, can we go ahead and enable 1.17 and 1.18 support  along with all updated dependencies in master? We can also have a release to this point ( so that we cover 1 year gap  since last release happened on Aug 2021)  with older go versions and dependencies then pick this PR in master. wdyt?

go.mod
slaskawi: Awesome! Thank you!

slaskawi (approved):
LGTM

humblec:
@phlogistonjohn @raghavendra-talur ptal.. thanks.

phlogistonjohn:
Hi again:

@raghavendra-talur and I were looking into seeing if the changes in PR #1898 were sufficient or not. We looked at @dims commands above. Unfortunately, we were stopped at the command `./hack/update-vendor.sh` which failed with some very vague and unhelpful error message.
We were also seeing some issues with tags, so we experimented with using "pretend" tags from my heketi fork like so: `./hack/pin-dependency.sh github.com/heketi/heketi/v10=github.com/phlogistonjohn/heketi/v10 v10.4.1`

I'd be happy to share the output of the failed command with you if requested.

So, once again I'm somewhat in the same position as before. I don't want to force newer versions of Go because I need to know that I can build heketi on older distros at any time. I would be willing to look at a minimum set of changes needed to get k8s building - but I can't tell from this PR what the minimum set of changes is really needed.

One idea I have is to simply tag heketi with v10.4.1 as it is today (without this PR). And then let you k8s folks give that a try. If additional changes on top of that are needed we can then figure it out and do a v10.4.2. I would definitely prefer to take things step by step.  Other suggestions are welcome.

If you think that this is helpful, we can do it next week and the "release" will simply be for "library consumption" not GA.

dims:
@phlogistonjohn i am happy to iterate ... `One idea I have is to simply tag heketi with v10.4.1 as it is today (without this PR). And then let you k8s folks give that a try. If additional changes on top of that are needed we can then figure it out and do a v10.4.2. I would definitely prefer to take things step by step. `

let's do this! thanks

raghavendra-talur:
Basically what we trying to do here is understand the reason why bumping up the k8s version in heketi from `k8s.io/api v0.15.12` to `k8s.io/api v0.24.1` fixes the issue you are seeing.

I expect both the versions to be "compatible" as per go module norms.

phlogistonjohn:
@dims happy to see you are open to the idea. Let's not make a sudden change on a Friday. :-)
I'll double check in on this discussion next Mon and we'll probably create the tag on Tuesday (6/28).

dims:
@phlogistonjohn of course

@raghavendra-talur if that were true if you just switch the k8s dependencies in heketi's go.mod from older k8s to newer k8s, it would compile out of the box (without the code changes that @humblec had in https://github.com/heketi/heketi/pull/1904/commits/49e440f300a7b0e44b52cb1f547626ef97efd5c4 commit). That's the crux of the problem.

dims:
Applied the following diff in https://github.com/kubernetes/kubernetes/

```
diff --git pkg/volume/glusterfs/glusterfs.go pkg/volume/glusterfs/glusterfs.go
index e915db930cfe..0361a8798229 100644
--- pkg/volume/glusterfs/glusterfs.go
+++ pkg/volume/glusterfs/glusterfs.go
@@ -29,8 +29,8 @@ import (
        dstrings "strings"
        "sync"

-       gcli "github.com/heketi/heketi/client/api/go-client"
-       gapi "github.com/heketi/heketi/pkg/glusterfs/api"
+       gcli "github.com/heketi/heketi/v10/client/api/go-client"
+       gapi "github.com/heketi/heketi/v10/pkg/glusterfs/api"
        "k8s.io/klog/v2"
        "k8s.io/mount-utils"
        netutils "k8s.io/utils/net"
diff --git pkg/volume/glusterfs/glusterfs_test.go pkg/volume/glusterfs/glusterfs_test.go
index 43ca2b140028..64b9985cd77b 100644
--- pkg/volume/glusterfs/glusterfs_test.go
+++ pkg/volume/glusterfs/glusterfs_test.go
@@ -22,7 +22,7 @@ import (
        "reflect"
        "testing"

-       gapi "github.com/heketi/heketi/pkg/glusterfs/api"
+       gapi "github.com/heketi/heketi/v10/pkg/glusterfs/api"
        "k8s.io/mount-utils"

        v1 "k8s.io/api/core/v1"
```


then ran the following commands:
```
[dims@dims-m1-7728 15:09] ~/go/src/k8s.io/kubernetes ⟩ hack/pin-dependency.sh github.com/heketi/heketi/v10 v10.4.1
Running: go mod download github.com/heketi/heketi/v10@v10.4.1
Resolved to github.com/heketi/heketi/v10@v10.4.1
Running: go mod edit -require github.com/heketi/heketi/v10@v10.4.1
Running: go mod edit -replace github.com/heketi/heketi/v10=github.com/heketi/heketi/v10@v10.4.1

Run hack/update-vendor.sh to rebuild the vendor directory
[dims@dims-m1-7728 15:09] ~/go/src/k8s.io/kubernetes ⟩ hack/update-vendor.sh
+++ [0628 15:11:48] logfile at /tmp/update-vendor.9CiZ/update-vendor.log
+++ [0628 15:11:48] go.mod: update staging references
+++ [0628 15:12:06] go.mod: propagate to staging modules
+++ [0628 15:12:07] go.mod: sorting staging modules
+++ [0628 15:12:27] go.mod: tidying
tsort: cycle in data
tsort: k8s.io/api
tsort: k8s.io/apimachinery
tsort: k8s.io/client-go
tsort: cycle in data
tsort: k8s.io/api
tsort: k8s.io/apimachinery
An error has occurred. Please see more details in /tmp/update-vendor.9CiZ/update-vendor.log
```

Will show the following error:
```
[dims@dims-m1-7728 15:13] ~/go/src/k8s.io/kubernetes ⟩ tail /tmp/update-vendor.9CiZ/update-vendor.log
	k8s.io/kubernetes/test/integration/client.test imports
	k8s.io/client-go/applyconfigurations/meta/v1: module k8s.io/client-go@latest found (v1.5.2, replaced by k8s.io/client-go@v0.15.12), but does not contain package k8s.io/client-go/applyconfigurations/meta/v1
k8s.io/kubernetes/cmd/kube-apiserver/app imports
	k8s.io/apiextensions-apiserver/pkg/apiserver imports
	k8s.io/apiserver/pkg/endpoints/handlers imports
	k8s.io/apimachinery/pkg/apis/meta/internalversion/validation: module k8s.io/apimachinery@latest found (v0.24.2, replaced by k8s.io/apimachinery@v0.15.12), but does not contain package k8s.io/apimachinery/pkg/apis/meta/internalversion/validation
k8s.io/kubernetes/pkg/kubeapiserver/options imports
	k8s.io/apiserver/pkg/admission/plugin/webhook/mutating imports
	k8s.io/apiserver/pkg/admission/configuration imports
	k8s.io/client-go/listers/admissionregistration/v1: module k8s.io/client-go@latest found (v1.5.2, replaced by k8s.io/client-go@v0.15.12), but does not contain package k8s.io/client-go/listers/admissionregistration/v1
```

This is because the version of k8s imports needed by heketi ( v0.15.12 - https://github.com/heketi/heketi/blob/master/go.mod#L17-L19 ) is not compatible with latest master of kubernetes. 

The only fix is to update heketi/heketi with latest released version of kubernetes and make a release of heketi will work because latest master of kubernetes is compatible with k8s v0.24.1 needed by heketi

NOTE that to be able to update heketi to k8s v0.24.1 will require changes ( see https://github.com/heketi/heketi/pull/1904 PR for patches )

dims:
@phlogistonjohn @raghavendra-talur see above for steps to recreate the problem and the error against latest v10.4.1

humblec:
@raghavendra-talur @phlogistonjohn ping..

dims:
@raghavendra-talur @phlogistonjohn I mentioned to humblec that this situation is not something we can keep dragging on. If heketi based glusterfs is going to be part of kubernetes/kubernetes repository, this is needed. If you can't land this, please say so explicitly so we can just rip the plugin out from kubernetes/kubernetes with an announcement and a short time period. This is a waste of everyone's time on all sides.

humblec:
@raghavendra-talur @phlogistonjohn considering there are good number of users/community which still use glusterfs from kuberenetes,  can you prioritize this?

humblec:
@raghavendra-talur @phlogistonjohn based on the possibilities of this PR merge, we have to make a decision in upstream kube about glusterfs deprecation release. The code freeze is on 2nd Aug.,  appreciated if you can revert on this PR at the earliest.

More or less, I dont get the concern of updating the master branch with latest or supported go versions and dependencies which is handled or proposed in this PR!!

phlogistonjohn:
Hi there, sorry for the slow response, but it is the middle of summer after 
all.

Let me reiterate my priority: I need to keep heketi working for a small number 
of legacy users. Our big concerns are security issues or major bugs within 
heketi server. We need to be able to patch heketi and build it on old distros 
and need to ensure it builds on older versions of Go.

If I had my way our product would be fully EOLed and the Heketi repo would 
already be archived. As it stands, I need to make sure we can handle 
"emergencies" on the heketi master branch.

I'm willing to consider options that don't require us to stop testing against 
older versions of Go. I've been speculating that if building heketi with a 
newer version of Go is the only way to keep heketi as a valid import for k8s 
then the only remaining options are to either see if a special branch can be 
used (a branch that supports only newer Go) for use as an import target but 
not building the server or a separate git repo (call it "heketi-lib" for example). 
Unfortunately, I do not have the resources to test either approach myself. 
I encourage those who are enthusiastic about continued support for heketi 
directly in k8s to do the experiments.

If this is not acceptable or no one is available to do the experiments and 
this leads to to k8s having to drop heketi from new k8s versions - then we 
should do so. Heketi has been effectively deprecated and in minimal maintenance only 
for quite a long time now and we've not been shy about saying so, but we don't 
have a large platform to shout it from. In addition, the k8s issue on this topic been open 
over a year now [1]. Perhaps, I should have participated in that thread and 
been more enthusiastic about saying that heketi support should be deprecated 
and then later removed. I don't exactly remember my thoughts at the time but 
perhaps I was unwilling to sound negative about the suggestions about a 
migration path off of gluster. For me, the only thing I would find unfortunate 
about the suggested plan to remove heketi is the short time frame. It would 
have bee nicer to let people know about this much more ahead of time.


[1] - https://github.com/kubernetes/kubernetes/issues/100897#issuecomment-1177274787

dims:
FYI, https://groups.google.com/a/kubernetes.io/g/dev/c/xEND8HRgfHU/m/26JHfHI7BAAJ 

We are fast-tracking the removal of `heketi/heketi` from `kubernetes/kubernetes` repository

dims:
@humblec we should close this PR out i think.

phlogistonjohn:
> FYI, https://groups.google.com/a/kubernetes.io/g/dev/c/xEND8HRgfHU/m/26JHfHI7BAAJ
> 
> We are fast-tracking the removal of `heketi/heketi` from `kubernetes/kubernetes` repository

Great, thanks for the heads up.

phlogistonjohn:
heketi project to be archived. closing PR.

